PR #24, opened 2025-06-12
Graphmixer

### LINKED ISSUE #17: GraphMixer Implementation

Jacob-Chmura:
Close #17

### DIFF
--- a/examples/linkproppred/graphmixer.py
+++ b/examples/linkproppred/graphmixer.py
@@ -80,8 +80,8 @@ class GraphMixer(nn.Module):
                     channel_dim_expansion=channel_dim_expansion,
                     dropout=dropout,
                 )
+                for _ in range(num_layers)
             ]
-            for _ in range(num_layers)
         )
         self.output_layer = nn.Linear(
             in_features=edge_dim + node_dim, out_features=embed_dim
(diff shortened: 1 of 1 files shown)


examples/linkproppred/graphmixer.py
```diff
@@ -0,0 +1,364 @@
...
+)
+parser.add_argument(
+    '--channel-dim-expansion',
+    type=float,
+    default=0.5,
+    help='channel dimension expansion factor in MLP sub-blocks',
+)
+
+
+class GraphMixer(nn.Module):
```
shenyangHuang, line 51: probably we can just setup as a layer in `tgm`

shenyangHuang: good for now, we should maybe start thinking about layer structures after alpha release

examples/linkproppred/graphmixer.py
```diff
@@ -0,0 +1,364 @@
...
+        out = z + x
+
+        # mix channels
+        z = self.channel_norm(out)
+        z = self.channel_ff(z)
+        out = z + out
+        return out
+
+
+class LinkPredictor(nn.Module):
```
shenyangHuang, line 184: this should be a layer for convenience, all link prediction task uses it, and no need to copy this again and again

shenyangHuang: what do you think

examples/linkproppred/graphmixer.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def _init_hooks(dg: DGraph, time_gap: int) -> List[DGHook]:
+    # Graphmixer always uses 1-hop recent neighbors
+    nbr_hook = RecencyNeighborHook(num_nbrs=[args.n_nbrs], num_nodes=dg.num_nodes)
+
+    # Always produce negative edge prior to neighbor sampling for link prediction
+    neg_hook = NegativeEdgeSamplerHook(low=0, high=dg.num_nodes)
+
+    class GraphMixerHook:
```
shenyangHuang, line 260: this can be a general hook as well?? not sure

shenyangHuang: but this is a great example for users to see how to create their own custom hook without relying on our built-in ones

examples/linkproppred/graphmixer.py
Jacob-Chmura: Yes I agree. Just copying it for now, and we can consolidate our nn module once more examples are in.

examples/linkproppred/graphmixer.py
Jacob-Chmura: > this can be a general hook as well?? not sure 

Its basically a clone of recency sampler, but implemented it here for simplicity, and to serve as a working example for users (and for us) on how custom hooks might look

PR #25, opened 2025-06-12
Recency Multi-Hop

### LINKED ISSUE #11: Multi-hop Recency Sampling

Jacob-Chmura:
Close #11


test/test_hooks/test_recency_nbr_hook.py
```diff
@@ -30,20 +31,392 @@ def test_bad_neighbor_sampler_init():
     with pytest.raises(ValueError):
         RecencyNeighborHook(num_nbrs=[-1], num_nodes=2)
     with pytest.raises(ValueError):
-        RecencyNeighborHook(num_nbrs=[1, 2], num_nodes=2)
+        RecencyNeighborHook(num_nbrs=[], num_nodes=2)
 
 
-def test_neighbor_sampler_hook_init(data):
+def test_neighbour_sampler_hook(data):
```
shenyangHuang: Looks good, a complex example. 
One case we can test is that what if the graph only contains self-edges i.e.
(src,src,time) edges, what does the neighbor sampler return?

Also, what happens with cycles, if you have a-b, b-c, c-a
, can you infinitely sample hops?

tgm/hooks.py
```diff
@@ -282,73 +282,84 @@ def __init__(self, num_nodes: int, num_nbrs: List[int]) -> None:
...
+
+        prev_nbr_nids, prev_nbr_times = None, None
+        for hop, num_nbrs in enumerate(self.num_nbrs):
+            if hop == 0:
+                if hasattr(batch, 'neg'):
+                    batch.neg = batch.neg.to(dg.device)
+                    seed_nodes = torch.cat([batch.src, batch.dst, batch.neg])
+                else:
+                    seed_nodes = torch.cat([batch.src, batch.dst])
+                seed_times = batch.time.repeat(3)
```
Jacob-Chmura: This shape won't align with TGB negatives

examples/linkproppred/tgat.py
```diff
@@ -75,27 +76,43 @@ def __init__(
...
-            (*batch.nbr_nids[hop].shape, self.embed_dim), device=device
-        )
-        time_feat = self.time_encoder(torch.zeros(len(batch.nids[hop]), device=device))
-        nbr_time_feat = self.time_encoder(
-            batch.nbr_times[hop] - batch.time.unsqueeze(dim=1).repeat(3, 1)
-        )
+        z = torch.zeros(len(batch.unique_nids), self.embed_dim, device=device)
+
+        for hop in reversed(range(self.num_layers)):
+            if batch.nids[hop].numel() == 0:
```
Jacob-Chmura, line 89: This occurs, for instance, on the first batch, when neighbourhoods don't exist yet (because we update queues _after_ trying to compute nbrs) to avoid breaking temporal causality. In this case we can just skip the hop.

examples/linkproppred/tgat.py
```diff
@@ -75,27 +76,43 @@ def __init__(
...
+            node_feat = torch.zeros(
+                (*batch.nids[hop].shape, self.embed_dim), device=device
+            )
+            node_time_feat = self.time_encoder(torch.zeros_like(batch.nids[hop]))
+
+            # If next next hops embeddings exist, use them instead of raw features
+            if hop < self.num_layers - 1:
+                nbr_feat = z[batch.nbr_nids_idx[hop]]
+            else:
+                nbr_feat = torch.zeros(
```
Jacob-Chmura, line 102: These should also read from static node features tensor if it exists

examples/linkproppred/tgat.py
```diff
@@ -163,9 +180,11 @@ def eval(loader: DGDataLoader, model: nn.Module, metrics: Metric) -> dict:
 
 def _init_hooks(dg: DGraph, sampling_type: str) -> List[DGHook]:
     if sampling_type == 'uniform':
-        nbr_hook = NeighborSamplerHook(num_nbrs=[args.n_nbrs])
+        nbr_hook = NeighborSamplerHook(num_nbrs=[args.n_nbrs, args.n_nbrs])
     elif sampling_type == 'recency':
-        nbr_hook = RecencyNeighborHook(num_nbrs=[args.n_nbrs], num_nodes=dg.num_nodes)
+        nbr_hook = RecencyNeighborHook(
+            num_nbrs=[args.n_nbrs, args.n_nbrs], num_nodes=dg.num_nodes
+        )
```
Jacob-Chmura: Temporary hard coding to 2 hops for debugging

examples/linkproppred/tgat.py
shenyangHuang: Yes, makes sense

examples/linkproppred/tgat.py
shenyangHuang: yes it should, we should have a separate PR on static node features tensor integration into all examples.

examples/linkproppred/tgat.py
```diff
@@ -75,27 +76,43 @@ def __init__(
...
+
+            out = self.attn[hop](
+                node_feat=node_feat,
+                time_feat=node_time_feat,
+                edge_feat=batch.nbr_feats[hop],
+                nbr_node_feat=nbr_feat,
+                nbr_time_feat=nbr_time_feat,
+                nbr_mask=batch.nbr_mask[hop],
+            )
+            z[batch.nid_to_idx[batch.nids[hop]]] = out
```
shenyangHuang, line 117: what does this do?

examples/linkproppred/tgat.py
shenyangHuang: yeah its fine, in practice, rarely goes over 2 hop.

tgm/hooks.py
```diff
@@ -282,73 +282,61 @@ def __init__(self, num_nodes: int, num_nbrs: List[int]) -> None:
...
+                seed_nodes = torch.cat(seed)
+                seed_times = batch.time.repeat(3)  # TODO: Won't work for tgb
+            else:
+                mask = batch.nbr_mask[hop - 1].bool()
+                seed_nodes = batch.nbr_nids[hop - 1][mask].flatten()
+                seed_times = batch.nbr_times[hop - 1][mask].flatten()
+
+            B = len(seed_nodes)
+            nbr_nids = torch.empty(B, num_nbrs, dtype=torch.long, device=device)
+            nbr_times = torch.empty(B, num_nbrs, dtype=torch.long, device=device)
```
shenyangHuang, line 320: why empty for some, but zero for others?

examples/linkproppred/tgat.py
Jacob-Chmura: We keep a batch-global storage `z` which holds embeddings for all unique nodes in the batch. `batch.nids[hop]` is the seed nodes on the current batch. Then `batch.nid_to_idx` is the inverse map from global to local coordinates. So `batch.nid_to_idx[batch.nids[hop]]` tells us where to _write_ the computed embeddings `out` from this hop.

test/test_hooks/test_recency_nbr_hook.py
Jacob-Chmura: Yeah, I can try to write these examples out. Currently nothing is done to treat self-edges or cycles any different from normal edges. 

In the first case, I expect the node to nbr map to be something like `x: [x, x, ..], y: [y, y, ...]` for a graph like `x-x   y-y`. 

In the second case, yes, loops are not treated any different. Since we use tensors to store outputs, nothing special should happen if a node appears more than once in a nbr list, e.g. `a: [b, c, b, c]` I suppose the self-edge is a special case of loop anyway.

tgm/hooks.py
Jacob-Chmura: Well `nbr_mask` needs to be zeroed out since it's a binary mask that defaults to 0. But `nbr_feats` could be `torch.empty`, you are right

PR #27, opened 2025-06-13
Deprecate Node Slicing

### LINKED ISSUE #26: Deprecate Node Slicing

Jacob-Chmura:
Up for discussion. May currently only have minimal perf improvement (we can afford to vectorized a bunch of methods in the backend). But much simpler to reason about and more likely to lead to correct implementation. 

Close #26


tgm/_storage/backends/array_backend.py
```diff
@@ -30,59 +30,52 @@ def __init__(self, data: DGData) -> None:
...
-        timestamps = set()
-        for i in range(*self._binary_search(slice)):
-            nodes = self._nodes_in_event(i)
-            if slice.node_slice is None or any(x in slice.node_slice for x in nodes):
-                timestamps.add(self._data.timestamps[i].item())
-        return len(timestamps)
+        lb_idx, ub_idx = self._binary_search(slice)
+        return len(self._data.timestamps[lb_idx:ub_idx].unique())
 
     def get_num_events(self, slice: DGSliceTracker) -> int:
```
shenyangHuang, line 76: optional, maybe we should add comment to clearly specify that we are counting both edge and node events here. For the tutorial on the library, we should also outline the design philosophy as well

shenyangHuang (approved):

PR #28, opened 2025-06-14
Adding TGB negative sampling hook 

shenyangHuang:
- [ ] unified all split-mode for validation set to be called `val` instead of `valid` for unification
- [ ] added `TGBNegativeEdgeSamplerHook` in `hooks.py`
- [ ] added TGB example for edgebank, getting the same performance, dataloader and negative sampling hook works
- [ ] added test for `TGBNegativeEdgeSamplerHook`

To do 
- [ ] add TGAT and TGN example with TGB hook, maybe that goes in a different PR

### DIFF
--- a/examples/linkproppred/TGB/edgebank.py
+++ b/examples/linkproppred/TGB/edgebank.py
@@ -1,0 +1,102 @@
+import argparse
+
+import numpy as np
+import torch
+from tgb.linkproppred.dataset_pyg import PyGLinkPropPredDataset
+from tgb.linkproppred.evaluate import Evaluator
+from tqdm import tqdm
+
+from tgm.graph import DGraph
+from tgm.hooks import TGBNegativeEdgeSamplerHook
+from tgm.loader import DGDataLoader
+from tgm.nn import EdgeBankPredictor
+from tgm.timedelta import TimeDeltaDG
+from tgm.util.seed import seed_everything
+
+parser = argparse.ArgumentParser(
+    description='EdgeBank TGB Example',
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
+parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
+parser.add_argument('--dataset', type=str, default='tgbl-wiki', help='Dataset name')
+parser.add_argument('--bsize', type=int, default=200, help='batch size')
+parser.add_argument('--window_ratio', type=float, default=0.15, help='Window ratio')
+parser.add_argument('--pos_prob', type=float, default=1.0, help='Positive edge prob')
+parser.add_argument(
+    '--memory_mode',
+    type=str,
+    default='unlimited',
+    choices=['unlimited', 'fixed'],
+    help='Memory mode',
+)
+
+
+def eval(
+    loader: DGDataLoader,
+    model: EdgeBankPredictor,
+    eval_metric: str,
+    evaluator: Evaluator,
+) -> dict:
+    perf_list = []
+    for batch in tqdm(loader):
+        neg_batch_list = batch.neg_batch_list
+        for idx, neg_batch in enumerate(neg_batch_list):
+            query_src = torch.tensor(
+                [batch.src[idx] for _ in range(len(neg_batch) + 1)]
+            )
+            query_dst = torch.cat([torch.tensor([batch.dst[idx]]), neg_batch])
+
+            y_pred = model(query_src, query_dst)
+            # compute MRR
+            input_dict = {
+                'y_pred_pos': np.array([y_pred[0]]),
+                'y_pred_neg': np.array(y_pred[1:]),
+                'eval_metric': [eval_metric],
+            }
+            perf_list.append(evaluator.eval(input_dict)[eval_metric])
+        model.update(batch.src, batch.dst, batch.time)
+    metric_dict = {}
+    metric_dict[eval_metric] = float(np.mean(perf_list))
+    return metric_dict
+
+
+args = parser.parse_args()
+seed_everything(args.seed)
+
+# * setting up tgb neg sampler
+dataset = PyGLinkPropPredDataset(name=args.dataset, root='datasets')
+eval_metric = dataset.eval_metric
+neg_sampler = dataset.negative_sampler
+evaluator = Evaluator(name=args.dataset)
+dataset.load_val_ns()
+dataset.load_test_ns()
+
+train_dg = DGraph(args.dataset, time_delta=TimeDeltaDG('r'), split='train')
+val_dg = DGraph(args.dataset, time_delta=TimeDeltaDG('r'), split='val')
+test_dg = DGraph(args.dataset, time_delta=TimeDeltaDG('r'), split='test')
+
+train_data = train_dg.materialize(materialize_features=False)
+val_loader = DGDataLoader(
+    val_dg,
+    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')],
+    batch_size=args.bsize,
+)
+test_loader = DGDataLoader(
+    test_dg,
+    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')],
+    batch_size=args.bsize,
+)
+
+model = EdgeBankPredictor(
+    train_data.src,
+    train_data.dst,
+    train_data.time,
+    memory_mode=args.memory_mode,
+    window_ratio=args.window_ratio,
+    pos_prob=args.pos_prob,
+)
+
+val_results = eval(val_loader, model, eval_metric, evaluator)
+print(' '.join(f'{k}={v:.4f}' for k, v in val_results.items()))
+test_results = eval(test_loader, model, eval_metric, evaluator)
+print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
--- a/tgm/hooks.py
+++ b/tgm/hooks.py
@@ -3,8 +3,9 @@ from __future__ import annotations
 from collections import deque
 from typing import Any, Deque, Dict, List, Protocol, Set, runtime_checkable
 
-import torch
 import numpy as np
+import torch
+
 from tgm._storage import DGSliceTracker
 from tgm.graph import DGBatch, DGraph
 
@@ -125,21 +126,23 @@ class NegativeEdgeSamplerHook:
 
 class TGBNegativeEdgeSamplerHook:
     requires: Set[str] = set()
-    produces = {'neg'}
+    produces = {'neg', 'neg_batch_list'}
     r"""Load data from DGraph using pre-generated TGB negative samples.
     Make sure to perform `dataset.load_val_ns()` or `dataset.load_test_ns()` before using this hook.
     Args:
         neg_sampler (object): The negative sampler object to use for sampling.
+        split_mode (str): The split mode to use for sampling, either 'val' or 'test'.
 
     Raises:
         ValueError: If neg_sampler is not provided.
     """
+
     def __init__(self, neg_sampler: object, split_mode: str) -> None:
         if neg_sampler is None:
             raise ValueError('neg_sampler must be provided')
         if split_mode not in ['val', 'test']:
             raise ValueError('split_mode must be one of val, test')
-        if neg_sampler.eval_set[split_mode] is None:
+        if neg_sampler.eval_set[split_mode] is None:  # type: ignore
             raise ValueError(
                 f'please run load_{split_mode}_ns() before using this hook'
             )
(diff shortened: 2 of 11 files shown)


examples/linkproppred/TGB/edgebank.py
Jacob-Chmura, line 1: I'm wondering whether we should have seperate examples for tgb/non-tgb. Or just have everything as tgb (if possible). What are your thoughts? My view is that almost everyone would be doing tgb eval anyway, and it's a bit simpler to have a single example for each method.

examples/linkproppred/TGB/edgebank.py
```diff
@@ -0,0 +1,102 @@
+import argparse
+
+import numpy as np
+import torch
+from tgb.linkproppred.dataset_pyg import PyGLinkPropPredDataset
```
Jacob-Chmura, line 5: TGB is now a dependency for the examples, so it should be added to the examples group: `uv add --group examples tgb`

examples/linkproppred/TGB/edgebank.py
```diff
@@ -0,0 +1,102 @@
...
+            query_dst = torch.cat([torch.tensor([batch.dst[idx]]), neg_batch])
+
+            y_pred = model(query_src, query_dst)
+            # compute MRR
+            input_dict = {
+                'y_pred_pos': np.array([y_pred[0]]),
+                'y_pred_neg': np.array(y_pred[1:]),
+                'eval_metric': [eval_metric],
+            }
+            perf_list.append(evaluator.eval(input_dict)[eval_metric])
```
Jacob-Chmura, line 56: Is this worth vectorizing?

examples/linkproppred/TGB/edgebank.py
```diff
@@ -0,0 +1,102 @@
...
+        model.update(batch.src, batch.dst, batch.time)
+    metric_dict = {}
+    metric_dict[eval_metric] = float(np.mean(perf_list))
+    return metric_dict
+
+
+args = parser.parse_args()
+seed_everything(args.seed)
+
+# * setting up tgb neg sampler
```
Jacob-Chmura:
```suggestion
```

examples/linkproppred/TGB/edgebank.py
```diff
@@ -0,0 +1,102 @@
...
+dataset = PyGLinkPropPredDataset(name=args.dataset, root='datasets')
+eval_metric = dataset.eval_metric
+neg_sampler = dataset.negative_sampler
+evaluator = Evaluator(name=args.dataset)
+dataset.load_val_ns()
+dataset.load_test_ns()
```
Jacob-Chmura, line 71: This is not a huge deal, but it seems not ideal that we (i.e. the user) needs to load in a completely separate copy of the data to initialize the negative sampler. Is there another API in tgb to construct the negative sampler directly?

If not, what do you think about hiding this behind `TBNegativeSamplerHook`: specifically, just let the negative sampler hook take a dataset name (`args.dataset`)?

Jacob-Chmura: It's nice that the evaluator just takes the dataset name. Presumably `dataset.load_ns()` must be called by the user directly? I'm asking because we don't actually use the dataset explicitly. The only thing that actually keeps reference to it is the negative sampler, but we do `load_ns` _after_ that.

Basically, do we need the `load_ns` and if so, how does it interact with the `tgm` notion of time granularity. Can it be unified?

tgm/hooks.py
```diff
@@ -123,6 +124,48 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
         return batch
 
 
+class TGBNegativeEdgeSamplerHook:
+    requires: Set[str] = set()
+    produces = {'neg', 'neg_batch_list'}
+    r"""Load data from DGraph using pre-generated TGB negative samples.
+    Make sure to perform `dataset.load_val_ns()` or `dataset.load_test_ns()` before using this hook.
```
Jacob-Chmura, line 131: Thoughts on taking dataset name as string argument, and managing the ns load ourselves without the user having to remember to do it?

tgm/hooks.py
```diff
@@ -123,6 +124,48 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+    Args:
```
Jacob-Chmura, line 133:
```suggestion
    
    Args:
```

tgm/hooks.py
```diff
@@ -123,6 +124,48 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+        if neg_sampler.eval_set[split_mode] is None:  # type: ignore
+            raise ValueError(
+                f'please run load_{split_mode}_ns() before using this hook'
+            )
+
+        self.neg_sampler = neg_sampler
+        self.split_mode = split_mode
+
+    def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
+        # this might complain if the edge is not found in the negative sampler, which could happen if the user is not using the correct version of dataset
```
Jacob-Chmura, line 155: By complain, you mean raise an error? Or warning? Can we recover from the exception if it occurs?

tgm/hooks.py
```diff
@@ -123,6 +124,48 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+        neg_batch_list = self.neg_sampler.query_batch(  # type: ignore
+            batch.src, batch.dst, batch.time, split_mode=self.split_mode
+        )
+        queries = []
+        tensor_batch_list = []
+        for _, neg_batch in enumerate(neg_batch_list):
```
Jacob-Chmura:
```suggestion
        for neg_batch in neg_batch_list:
```

tgm/hooks.py
```diff
@@ -123,6 +124,48 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+            queries.append(neg_batch)
+            tensor_batch_list.append(torch.tensor(neg_batch, dtype=torch.long))
+        unique_neg = np.unique(np.concatenate(queries))
+        batch.neg = torch.tensor(unique_neg, dtype=torch.long)  # type: ignore
```
Jacob-Chmura, line 165: This isn't used in the current example. Do we need this here?

test/test_hooks/test_tgb_negative_sampling_hook.py
Jacob-Chmura, line 1: Nice

examples/linkproppred/TGB/edgebank.py
shenyangHuang: Thanks Jacob, I think going forward, we can prioritize having TGB evaluation. However, it is also possible that people might want to do their own evaluation and the basic example work like that. As this is a software library, it is always good to have more examples. I think we restructure the example (also easily) when we go with alpha release and when integrating tutorials, not an issue for now I think. 

TLDR: currently and for clarity, let's keep the TGB folder (also it requires TGB dependency). We can restructure for the alpha release

examples/linkproppred/TGB/edgebank.py
shenyangHuang: `uv add --group examples py-tgb`

shenyangHuang: changed in `uv.lock` to 

```
examples = [
    { name = "torchmetrics" },
    { name = "tqdm" },
    { name = "py-tgb" },
]
```

shenyangHuang: ran the above and added to `uv.lock`

shenyangHuang: also added `uv sync --group examples` to README

examples/linkproppred/TGB/edgebank.py
shenyangHuang: what you mean? you mean `np.array`, you have to do that for standardized input to TGB evaluator. Note that the `evaluator` now is from TGB so it handles all the evaluation logic. This is required

examples/linkproppred/TGB/edgebank.py
shenyangHuang: yes and no, by default the dataset in TGB doesn't `load_ns` because some of the larger dataset has very large `ns` size (ten times larger than dataset) so if the user don't do link prediction, they don't need ns.

shenyangHuang: Here is the TLDR trade-off:
- solution one: each TGB hook loads its own dataset, calls load_ns and get its own neg sampler
problem: actually we just need one neg_sampler, this approach will create two copies of dataset, each for val and test, benefit: more straightforward, more abstracted away

- solution two: current approach, each TGB hook gets an input neg_sampler that allows it to get the test and val samples. one copy of dataset in example

shenyangHuang: I think solution one is nicer and more abstracted away and the user think less of what it is doing under the hood. However, I implemented solution 2 because:
- it avoid loading one copy of dataset for each TGB hook
- it makes the negative sampler hook very general actually, as long as you are passing a neg_sampler (similarly setup as TGB one), you can reuse this in your own custom dataset. You would still call `TGBNegativeEdgeSamplerHook` but just make your own neg_sampler, this is quite helpful especially when people want to try their custom negative sample etc. 
- in both cases, the evaluation loop is the same, so I don't think it makes a fundamental difference, user will just copy what we did in TGB example for the most part


TLDR: do we want `TGBNegativeEdgeSamplerHook` to be able to take in custom negative sampler? or do we want to fix it only for TGB? My personal take is that TGB evaluation is simply a novel / more robust way of evaluation and the user can also use their own negative sampler on novel datasets

shenyangHuang: I suggest to keep it as is but happy to discuss more.

tgm/hooks.py
shenyangHuang: it should throw an error, if you can't find the negative sample of an edge, it could be:
- you are running the wrong version of the dataset
- you didn't load your test or val ns (and TGB will complain correspondingly)

both errors will be throw by the neg_sampler from TGB

tgm/hooks.py
shenyangHuang: yes we need this for neighbor sampler :)

README.md
```diff
@@ -49,6 +49,12 @@ uv pip install git+https://github.com/tgm-team/tgm.git
 python -c 'import tgm; print(tgm.__version__)'
 ```
 
+running examples
```
Jacob-Chmura, line 52: Sure, we can just keep this here for temporary purposes, but this will be refined prior to release.

examples/linkproppred/TGB/edgebank.py
Jacob-Chmura: Sounds good.

tgm/hooks.py
Jacob-Chmura: Okay, so safe for use to ignore I suppose (unless we handle the `load_ns` ourselves)

examples/linkproppred/TGB/edgebank.py
Jacob-Chmura: What I mean is, do a single batched inference for the entire batch as opposed to looping through the neg_batch_list one at a time, doing all the inferences, and calling eval separately on each negative sub batch. 

I understand that TGB evaluator expects ndarray, I'm wondering if we can just do a single inference, single `to_numpy()` , then single eval on the batched np array?

examples/linkproppred/TGB/edgebank.py
Jacob-Chmura: Okay. Generally I think the `TGBHook` should be specialized for TGB. And users can create there own hook that's more generic. If I were to use the TGB hook using my own negative sampler, how do I know what API this hook expects? How do I know the exceptions it could raise etc? 

The avoiding of loading multiple copies across test/val/train hooks is valid. Overall, I'm happy to leave this as is, and we can revisit it later. I antipicate that at some point we will make it such that the hook manager can avoid these duplicate loads. 

I also think it's a design from TGB side that I need to understand more.

Jacob-Chmura (approved):

examples/linkproppred/TGB/edgebank.py
shenyangHuang: Hi Jacob, I see what you mean. Looping through each positive edge and its negative edges are how it was setup in TGB and OGB, conceptually simple. Yes, you can just get a stack of tensors for evaluation as well, requires changing the TGB code to work with batch. Definitely might have improvements (at least eliminating some of the loops) but out of scope for this PR. Also if there are many edges in a batch, stacking them all into a tensor might cause OOM (say batch size = 1,000,000), though it is unrealistic in the current evaluation

shenyangHuang: good point to maybe leave it as optimization in the future. Requires TGB code change

shenyangHuang: https://github.com/shenyangHuang/TGB/blob/main/tgb/linkproppred/evaluate.py Here is pointer to TGB evaluator

examples/linkproppred/TGB/edgebank.py
shenyangHuang: Yeah I think avoiding duplicate loads in 
- hook manager 
- dataloader 
is an issue we should address

examples/linkproppred/TGB/edgebank.py
shenyangHuang: See issue #31

examples/linkproppred/TGB/edgebank.py
shenyangHuang: I created an issue to keep track, see #30

README.md
shenyangHuang: yeah, we can keep it for development build and most people will want to run the examples anyways

PR #29, opened 2025-06-16
Fix end time bounds

### LINKED ISSUE #9: Clean up **off-by-one** edge case when iterating by time

Jacob-Chmura:
Close #9

### DIFF
--- a/test/test_storage_impl.py
+++ b/test/test_storage_impl.py
@@ -75,7 +75,7 @@ def test_get_end_time_edge_data(DGStorageImpl, data, request):
     data = request.getfixturevalue(data)
     storage = DGStorageImpl(data)
 
-    assert storage.get_end_time(DGSliceTracker()) == data.timestamps[-1]
+    assert storage.get_end_time(DGSliceTracker()) == data.timestamps[-1] + 1
 
 
 @pytest.mark.parametrize('data', ['edge_only_data', 'edge_only_data_with_features'])
--- a/test/test_storage_impl.py
+++ b/test/test_storage_impl.py
@@ -75,7 +75,7 @@ def test_get_end_time_edge_data(DGStorageImpl, data, request):
     data = request.getfixturevalue(data)
     storage = DGStorageImpl(data)
 
-    assert storage.get_end_time(DGSliceTracker()) == data.timestamps[-1] + 1
+    assert storage.get_end_time(DGSliceTracker()) == data.timestamps[-1]
 
 
 @pytest.mark.parametrize('data', ['edge_only_data', 'edge_only_data_with_features'])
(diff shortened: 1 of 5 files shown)


tgm/_storage/backends/array_backend.py
```diff
@@ -15,16 +15,6 @@ class DGStorageArrayBackend(DGStorageBase):
...
-        # Pre-compute edge event indices
-        self._edge_idx_map = {
-            idx.item(): i for i, idx in enumerate(self._data.edge_event_idx)
-        }
-        self._node_idx_map = (
-            {idx.item(): i for i, idx in enumerate(self._data.node_event_idx)}
-            if self._data.node_event_idx is not None
-            else None
-        )
-
```
Jacob-Chmura, line 27: These are no longer needed but forget to remove them when deprecating node events in previous PR

tgm/_storage/backends/array_backend.py
```diff
@@ -39,7 +29,7 @@ def get_end_time(self, slice: DGSliceTracker) -> Optional[int]:
         lb_idx, ub_idx = self._binary_search(slice)
         if lb_idx >= ub_idx:
             return None
-        return int(self._data.timestamps[ub_idx - 1].item())
+        return int(self._data.timestamps[ub_idx - 1].item()) + 1
```
Jacob-Chmura: The `end_time` is now exclusive. So if the last timestamp on the graph (or on the temporal subgraph) is `time=5`, then `dg.end_time == 6`. That way if the user slices using the end time `dg.slice(0, dg.end_time)` they recover the full graph as expected

tgm/loader.py
Jacob-Chmura, line 1: This was the main upside of the changes in this pr. But also, it makes more sense from an API design perspective to have _both_ event based and time based indexing end-range exclusive.

test/test_storage_impl.py
shenyangHuang, line 1: why add +1? because bound is open?

tgm/_storage/backends/array_backend.py
shenyangHuang: but it is strange that `dg.end_time == 6` even though the last edge is time=5

shenyangHuang: also `dg.start_time == 0` (inclusive) and end time is now exclusive.

shenyangHuang: think about it from a time granularity point of view:
the data is monthly, say from January to June 
`dg.end_time == July`??? not intuitive to me

tgm/loader.py
shenyangHuang: this point I agree. But should we change end time to be +1?

Jacob-Chmura: The intention was to have the semantics such that the user is not forced to do: `dg.slice(start_time, dg.end_time + 1)` themselves (or worse yet, have to figure out what number to add to `dg.end_time`)

Sort of meant to mimic `x[0:len(x)]` in the case of arrays or tensors.

test/test_storage_impl.py
Jacob-Chmura: yes

shenyangHuang (approved):

PR #34, opened 2025-06-16
Deduplication Batch Index Hook

### LINKED ISSUE #32: Inverse Index Lookup Hook

Jacob-Chmura:
Close #32 

High-level usage:
```python

# Just do inference on unique ids
z = model(batch.unique_nids) 

# Use *_idx fields to lookup the relevant embeddings
src_z = z[batch.src_idx]
dst_z = z[batch.dst_idx]

# These are optional fields that are added if using negative sampler or nbr sampler, respectively
neg_z = z[batch.neg_idx] if hasattr(batch, 'neg') else None
nbr_z = [z[idx] for idx in batch.nbr_nids_idx] if hasattr(batch, 'nbr_nids') else None

# If you got custom hooks and want to just get the embeddings for some arbitrary node tensors (e.g. tgb)
query_ids: torch.Tensor = ...
z_query_ids = z[batch.nid_to_idx[query_ids]]
```

### DIFF
--- a/examples/nodeproppred/gcn.py
+++ b/examples/nodeproppred/gcn.py
@@ -59,7 +59,7 @@ class GCN(nn.Module):
     ) -> Tuple[torch.Tensor, torch.Tensor]:
         edge_index = torch.stack([batch.src, batch.dst], dim=0).to(node_feat.device)
         z = self.encoder(node_feat, edge_index)
-        z_node = z[batch.node_ids]
+        z_node = z[batch.nid_to_idx[batch.node_ids]]  # type: ignore
         pred = self.decoder(z_node)
         return pred
 
--- a/tgm/hooks.py
+++ b/tgm/hooks.py
@@ -121,8 +121,9 @@ class DeduplicationHook:
         if hasattr(batch, 'neg'):
             nids.append(batch.neg)
         if hasattr(batch, 'nbr_nids'):
-            for hop_nids in batch.nbr_nids:
-                nids.append(hop_nids.flatten())
+            for hop in range(len(batch.nbr_nids)):
+                hop_nids, hop_mask = batch.nbr_nids[hop], batch.nbr_mask[hop].bool()  # type: ignore
+                nids.append(hop_nids[hop_mask])
 
         all_nids = torch.cat(nids, dim=0)
         unique_nids = torch.unique(all_nids)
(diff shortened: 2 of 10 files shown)


tgm/hooks.py
```diff
@@ -95,6 +98,50 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
 
+class DeduplicationHook:
+    requires: Set[str] = set()
+    produces = {
+        'unique_nids',
+        'nid_to_idx',
+        'src_idx',
+        'dst_idx',
+        'neg_idx',
+        'nbr_nids_idx',
```
shenyangHuang, line 109: why does it produce 'nbr_nids_idx'? We should do Dedup before sampling neighbors right?

tgm/hooks.py
```diff
@@ -95,6 +98,56 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+        if hasattr(batch, 'neg'):
+            nids.append(batch.neg)
+        if hasattr(batch, 'nbr_nids'):
+            for hop in range(len(batch.nbr_nids)):
+                hop_nids, hop_mask = batch.nbr_nids[hop], batch.nbr_mask[hop].bool()  # type: ignore
+                nids.append(hop_nids[hop_mask])
+
+        all_nids = torch.cat(nids, dim=0)
+        unique_nids = torch.unique(all_nids)
+        max_nid = int(unique_nids.max().item())
```
Jacob-Chmura, line 130: The dense tensor representation is going to be slightly better than the native python dict, but we should beware that this becomes very memory hungry when the `max_nid` is very large... And we allocate these tensors on each batch...

tgm/hooks.py
```diff
@@ -95,6 +98,56 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
 
+class DeduplicationHook:
+    requires: Set[str] = set()
+    produces = {
+        'unique_nids',
+        'nid_to_idx',
+        'src_idx',
+        'dst_idx',
+        'neg_idx',
+        'nbr_nids_idx',
```
Jacob-Chmura, line 109: Arguably, from a modularity perspective we could skip these all together and just give the user the `nid_to_idx` and let them lookup whichever indices they want (e.g. how we discussed it would work for tgb negatives). Since the user has `batch.src, batch.dst, batch.neg` they can do `z[batch.nid_to_idx[batch.src]]` themselves.

It gets slightly more complex with neighbours (since you need to mask out invalid neighborhoods, and need to be on the _right hop_. I think this is worth discussing though.

examples/nodeproppred/gcn.py
shenyangHuang, line 1: Looks great

tgm/hooks.py
shenyangHuang: that is true, however, it is still a 1d tensor so should be reasonably good, definitely can optimize later on.

PR #37, opened 2025-06-18
Adding MRR for existing models

shenyangHuang:
- [ ] adding MRR in CTDG for GCN on `tgbl-wiki`, requires iterating though both snapshots and batches of edges, currently the timestamps are not working correctly.  also for other granularities than `r`, maybe we can also support iterating by number of edges.  
- [ ] adding TGB example for TGAT, run into issue #38

### DIFF
--- a/examples/linkproppred/TGB/gcn.py
+++ b/examples/linkproppred/TGB/gcn.py
@@ -1,0 +1,312 @@
+import argparse
+import time
+from typing import Tuple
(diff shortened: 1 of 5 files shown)


Jacob-Chmura (commented):
Presumably this can be closed since I added TGAT MRR and the discrete time methods are being stacked in #87 ?

PR #45, opened 2025-06-20
Integration Test Scaffolding

### LINKED ISSUE #22: Integration Testing

Jacob-Chmura:
# Purpose
The purpose of this PR is to set up the scaffolding and infrastructure for running automated integration tests on the Mila cluster. To achieve this, I created a [self-hosted GitHub runner](https://docs.github.com/en/actions/hosting-your-own-runners) that starts a Daemon process and listens for events. 

The integration test suite wraps each of our examples, following the recipe:
1. Get the GitHub event and do basic setup (clone repo, create environment, load modules)
2. Create shared files across the cluster to keep track of Slurm outputs
3. Each test wraps an individual example (with slightly simpler parameters e.g. 1-epoch)
4. Create an `sbatch job` with the appropriate resources and schedule each example.
5. Read the output artifacts (currently just raw logs) and publish them on the github ui

## Key Changes
- moved all unit tests to `/test/unit` and added integration tests to `/test/integration`
- setup [codecov](https://about.codecov.io/) bot on our unit test suite
- new integration test workflow runs periodically (every Monday at 6:30 AM UTC), but also via manual trigger of push to stable branch
- added basic edgebank example and verified the pipeline

### Security
Above all, it would be beneficial to double-check that everything here complies with cluster policies. There are some safeguards I put in place
- GitHub token published as a secret with only write access to the GitHub CI
- prevent workflow from running on PR's (since this opens us up to vulnerability where someone forks our repo and issues arbitrary code). This means a job will only run when an organization owner approves (@Jacob-Chmura or @shenyangHuang)
- extra safety check to ensure only an org owner is running the job, and set timeouts on everything
- The GitHub CI runner will be running in my user process, so it should be limited in the actions it can take on the cluster anyway

## Relevant Prs
Close #22 

## Out of scope
- Adding integration tests for all our other examples
- Parsing output log files to isolate the key metrics (performance and efficiency). I have this code working in one of my private repos
- Optimizations (caching on the environment and data, run each integration test in parallel)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -49,6 +49,7 @@ jobs:
         run: uv run pytest test -m "slow" -vvv
 
       - name: Get Latest CI Run Log Path
+        if: always()
         id: get-log-dir
         run: |
           # Parse path + ID written by the test runner (2 lines in ~/tgm_ci/latest_path.txt)


.github/workflows/integration.yml
```diff
@@ -0,0 +1,57 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  integration-tests-slurm:
+    name: Integration testing on a SLURM cluster
+    runs-on: self-hosted
+    timeout-minutes: 60
+    if: github.actor == 'jacob-chmura' || github.actor == 'shenyanghuang'
```
Jacob-Chmura, line 20: Extra safety to protect too many jobs on the cluster

test/integration/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+
+            result = subprocess.run(
+                ['sacct', '-j', job_number, '--format=State', '--noheader'],
+                capture_output=True,
+                text=True,
+            )
+            print(result.stdout.strip())
+            state = result.stdout.strip().split()[0]
+            if state in ['COMPLETED', 'FAILED', 'CANCELLED']:
+                break
```
Jacob-Chmura: Need to re-consider poll intervals or explicitly signal output. Current logs should be parsed to extract performance and latency numbers as well.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,74 @@
...
+    env:
+      TGM_CI_LOG_BASE: $HOME/tgm_ci
```
Jacob-Chmura: It would make more sense to run the integration tests in `~scratch` or `SLURM_TMPDIR` to take advantage of faster IO ops. Arguably it is slightly this would become more hardcoded to the mila cluster (whereas otherwise the only assumptions we make is `slurm` exists).

.github/workflows/integration.yml
```diff
@@ -0,0 +1,74 @@
+name: Integration Tests
+
+on: # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - master
```
Jacob-Chmura, line 6: Arguably this may not be necessary but I like having full observability after every commit

test/integration/test_edgebank.py
```diff
@@ -0,0 +1,22 @@
...
+        '--partition=main',
+        '--cpus-per-task=2',
+        '--mem=4G',
+        '--time=0:10:00',
+    ]
+)
+def test_edgebank(slurm_job_runner, dataset):
+    cmd = f"""
+echo "Downloading dataset: {dataset}"
+echo "y" | python -c "from tgb.linkproppred.dataset import LinkPropPredDataset; LinkPropPredDataset('{dataset}')"
```
Jacob-Chmura, line 17: This is a bit unfortunate, and it should probably be downloaded once across all tests and cached.

.github/workflows/integration.yml
shenyangHuang: agree, let's keep for now

.github/workflows/integration.yml
shenyangHuang: do we know how many jobs will be launched automatically at a time? Also, as a  heads-up, the cluster is usually very busy during deadline period. The test slurm jobs will likely pend a long time during these period, i.e. close to ICML, NeurIPS, ICLR etc.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Run Integration Tests
+        run: uv run pytest test -m "slow" -vvv
+
+      - name: Get Latest CI Run Log Path
+        if: always()
+        id: get-log-dir
+        run: |
+          # Parse path + ID written by the test runner (2 lines in ~/tgm_ci/latest_path.txt)
+          mapfile -t lines < "$TGM_CI_LOG_BASE/latest_path.txt"
+          LOG_PATH="${lines[0]}"
```
shenyangHuang, line 57: if you are storing the logs, I suggest directly store it to your scratch directory on mila cluster, to avoid the logs taking over too much of your personal limited disc space on the cluster

shenyangHuang: `#SBATCH -o /network/scratch/<u>/<username>/slurm-%j.out  # Write the log on scratch`

.github/workflows/testing.yml
```diff
@@ -41,7 +41,16 @@ jobs:
         run: uv sync --all-extras --dev --frozen
 
       - name: Test non-gpu tests with pytest
-        run: uv run pytest test -m "not gpu" --cov=tgm
+        run: uv run pytest test -m "not gpu and not slow" --cov=tgm --cov-report=xml --cov-append -vvv
```
shenyangHuang, line 44: mila cluster prioritizes gpu jobs, I would just launch all jobs as GPU jobs and some of them you can choose not to use gpu

test/integration/test_edgebank.py
shenyangHuang: Yes, for now it is fine, but let's come back to it and find a fix to optimize it. If you are pointing to the same environment, then the dataset would be just loaded

.github/workflows/integration.yml
Jacob-Chmura: Each example / integration test runs in a seperate job to isolate resources and keep them independent. So on the order of 10. But they are fairly short lived and only request a single gpu.

You raise a good point though, and it is possible that we'll get failures due to timeout for no allocation. I chose 6am on monday to try to mitigate this.

.github/workflows/integration.yml
Jacob-Chmura: I am actually using `SLURM_TMP_DIR` which should be the fastest. It also get's automatically cleaned up after every job. No retention of logs should be fine as we upload to github right away anyway.

.github/workflows/testing.yml
Jacob-Chmura: To be clear, this is the github action for our unit test suite. So it runs on the github runners (not mila). I don't do gpu tests because the github instances you get for free are cpu-only.

PR #54, opened 2025-06-24
Run Integration Tests in Parallel

### LINKED ISSUE #50: Parallel Dispatch Integration Tests

Jacob-Chmura:
# Purpose
The purpose of this PR is to dispatch all the integration tests at once on the cluster. In practice, we'll only get a _2x_ speedup since the scheduler limits each user to 2 GPUs at once.

## Key Changes
- add `pytest-xdist` as dev dependency
- Rename `slow` pytest marker to `integration` for clarity
- submit 2 concurrent integration tests at a time from the workflow file (no real reason to do more due to slurm constraints)
- fix typo in TGN example script and TGAT test setup

## Relevant Prs
Close #50

## Out of scope
- Parsing output log files to isolate the key metrics (performance and efficiency). I have this code working in one of my private repos


.github/workflows/integration.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         run: uv sync --group examples
 
       - name: Run Integration Tests
-        run: uv run pytest test -m "slow" -vvv
+        run: uv run pytest -m "integration" -n 2 -vvv
```
Jacob-Chmura, line 49: Run's two jobs in parallel.

examples/linkproppred/tgn.py
Jacob-Chmura, line 1: Typo caught while running the tests

shenyangHuang: the tests are working then :)

PR #56, opened 2025-06-24
Fix TGAT Attention Bug

### LINKED ISSUE #52: Device Transfer on Deduplication without Nbrs

Jacob-Chmura:
# Purpose
When running multi-hop sampling, there will often be times when the nbr mask is (partially) full. The temporal multi-head attention module will take this mask into account when computing the attention tensor: https://github.com/tgm-team/tgm/blob/5b4949e5cb06b5fe94e8f292b9412c908d7f0795/tgm/nn/attention.py#L64-L76

But the key value heads propagate the features as is: https://github.com/tgm-team/tgm/blob/5b4949e5cb06b5fe94e8f292b9412c908d7f0795/tgm/nn/attention.py#L54-L57

Since nbr tensors are initialized with `torch.empty()` these garbage values mess with network gradients and lead to `nans`.

## Solution
Safely initialize all features with `torch.zeros`.

## Relevant Prs
Close #52

### DIFF
--- a/examples/linkproppred/tgat.py
+++ b/examples/linkproppred/tgat.py
@@ -116,9 +116,6 @@ class TGAT(nn.Module):
             )
             z[batch.nid_to_idx[batch.nids[hop]]] = out
 
-            # TODO: Merge layers to combine attention results and node original features
-            # z = self.merge_layers[hop](out, node_feat)
-
         z_src, z_dst, z_neg = z[batch.src_idx], z[batch.dst_idx], z[batch.neg_idx]  # type: ignore
         pos_out = self.link_predictor(z_src, z_dst)
         neg_out = self.link_predictor(z_src, z_neg)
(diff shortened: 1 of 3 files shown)


shenyangHuang:
> # Purpose
> When running multi-hop sampling, there will often be times when the nbr mask is (partially) full. The temporal multi-head attention module will take this mask into account when computing the attention tensor:
> 
> https://github.com/tgm-team/tgm/blob/5b4949e5cb06b5fe94e8f292b9412c908d7f0795/tgm/nn/attention.py#L64-L76
> 
> But the key value heads propagate the features as is:
> 
> https://github.com/tgm-team/tgm/blob/5b4949e5cb06b5fe94e8f292b9412c908d7f0795/tgm/nn/attention.py#L54-L57
> 
> Since nbr tensors are initialized with `torch.empty()` these garbage values mess with network gradients and lead to `nans`.
> 
> ## Solution
> Safely initialize all features with `torch.zeros`.
> 
> ## Relevant Prs
> Close #52

good catch, is this from the TGB (PyG) implementation? if so, we can submit this as PR for the PyG code to fix it too

shenyangHuang (approved):

PR #58, opened 2025-06-24
TGAT MRR

### LINKED ISSUE #57: TGAT MRR CudaMemory Issue

Jacob-Chmura:
# Purpose
The purpose of this PR is trying to replicate and patch #57. I extracted the TGAT mrr based example to try and isolate the problem.

### Performance
**Note: our validation MRR after a single epoch is 0.0740. The TGB leaderboard expects 0.131 ± 0.008 (after 10 epochs). A randomly initialized TGAT model (garbage weights) achieves MRR = 0.0438, so something is happening**.

Probably training longer would help. As would using actual static node features (or gaussian random) instead of zeros. There are probably also actual bugs.
 
## Key Changes
- TGAT mrr example
- update nbr sampler to make it compatible with non-equal positive and negative links. Used the heurstic fake link time we previously discussed

## Relevant Prs
Close #57

## Out of scope
- Debugging actual performance


examples/linkproppred/TGB/tgat.py
```diff
@@ -0,0 +1,273 @@
...
+                )
+                for i in range(num_layers)
+            ]
+        )
+
+    def forward(self, batch: DGBatch) -> Tuple[torch.Tensor, torch.Tensor]:
+        device = batch.src.device
+        z = torch.zeros(len(batch.unique_nids), self.embed_dim, device=device)
+
+        for hop in reversed(range(self.num_layers)):
```
shenyangHuang, line 89: reversed?

shenyangHuang: maybe let's walk through this together when we meet

tgm/hooks.py
```diff
@@ -305,11 +305,24 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+                times = [batch.time.repeat(2)]  # Real link times
                 if hasattr(batch, 'neg'):
                     batch.neg = batch.neg.to(device)
                     seed.append(batch.neg)
+
+                    # This is a heuristic. For our fake (negative) link times,
+                    # we pick random time stamps within temporal window of the batch.
+                    # Using random times on the whole graph will likely produce information
+                    # leakage, making the prediction easier than it should be.
+                    fake_times = torch.randint(
```
shenyangHuang, line 318: yeah I think this is fine for now. In the ideal case, you should produce timestamp for fake edges based on the true edge that it corresponds to, this is a sufficient condition for us now. We can also address this later on

shenyangHuang (approved):

examples/linkproppred/TGB/tgat.py
Jacob-Chmura: Sounds good. This is just how message passing is implemented. You always have to start at your boundary (highest hop)

PR #63, opened 2025-06-25
Uniform Multi-Hop Sampler

### LINKED ISSUE #12: Multi-Hop Uniform Sampling

Jacob-Chmura:
# Purpose
The purpose of this PR is to add multi-hop support to the COO based uniform neighbor sampler.

### Performance
TODO: Need to validate correctness once #58  is in before writing extensive tests.
 
## Key Changes
- Storage engine only answers 1-hop queries now (the multi-hop aggregation is left to the hook (https://github.com/tgm-team/tgm/blob/77eabb295ca739cefd01e8e2f9db00fa07943dd8/tgm/hooks.py#L170). Note that the driver code is _exactly_ the same between uniform sampler and recency sampler. The only difference is how we manage state and query the nbr ids. This is a good sign, and we could probably eventually write a base class hook (`BaseNeighborSamplerHook`) which manages the joint state and keeps the API fixed for all neighbour sampler implementations
- Cleaned up the storage 1-hop nbr query design (and changed to API in the base class)

## Relevant Prs
Close #12 

## Out of scope
- Non-trivial performance optimizations


tgm/hooks.py
```diff
@@ -193,20 +193,51 @@ def num_nbrs(self) -> List[int]:
...
+            else:
+                mask = batch.nbr_mask[hop - 1].bool()  # type: ignore
+                seed_nodes = batch.nbr_nids[hop - 1][mask].flatten()  # type: ignore
+                seed_times = batch.nbr_times[hop - 1][mask].flatten()  # type: ignore
+
+            # TODO: Storage needs to use the right device
+            nbr_nids, nbr_times, nbr_feats, nbr_mask = dg._storage.get_nbrs(
+                seed_nodes,
+                num_nbrs=num_nbrs,
+                slice=DGSliceTracker(end_time=dg._slice.end_time),
```
Jacob-Chmura: Actually, might have to be `dg._slice.end_idx` since we don't want it to _look forward_ to the next batch (even if tiemstamps match)

examples/linkproppred/tgn.py
```diff
@@ -278,7 +278,7 @@ def forward(
         )
         time_feat = self.time_encoder(torch.zeros(len(batch.nids[hop]), device=device))
         nbr_time_feat = self.time_encoder(
-            batch.nbr_times[hop] - batch.time.unsqueeze(dim=1).repeat(3, 1)
+            batch.time.unsqueeze(dim=1).repeat(3, 1) - batch.nbr_times[hop]
```
shenyangHuang: why is reversed?

tgm/_storage/backends/array_backend.py
```diff
@@ -87,42 +85,51 @@ def get_nbrs(
...
-        for i, nbrs_set in enumerate(nbrs.values()):
-            node_nbrs = list(nbrs_set)
+        src_list = edges[0].tolist()
+        dst_list = edges[1].tolist()
+        eid_list = event_ids.tolist()
+
+        # This is a loop over the entire graph up to (not including) the current batch end time
+        # which results in quadratic cost for a single epoch. Consider raises a warning to let the
+        # user know that this is not the right backend for this.
+        nbrs: Dict[int, List[Tuple[int, int]]] = {node: [] for node in seed_nodes_set}
```
shenyangHuang, line 95: this is the main cost part

tgm/_storage/backends/array_backend.py
```diff
@@ -87,42 +85,51 @@ def get_nbrs(
...
+        for s, d, i in zip(src_list, dst_list, eid_list):
+            if s in nbrs:
+                nbrs[s].append((i, d))
+            if d in nbrs:
+                nbrs[d].append((i, s))
+
+        B = len(seed_nodes)
+        nbr_nids = torch.full((B, num_nbrs), -1, dtype=torch.long, device=device)
```
shenyangHuang, line 103: worst case these are O(n), where n is the number of nodes in the graph.

tgm/_storage/backends/array_backend.py
```diff
@@ -87,42 +85,51 @@ def get_nbrs(
...
                 continue
-            if n_nbrs != -1 and len(node_nbrs) > n_nbrs:
-                node_nbrs = random.sample(node_nbrs, k=n_nbrs)
 
-            nbr_nids_, nbr_times_, nbr_feats_ = [], [], []
-            for event_idx, edge_idx in node_nbrs:
-                nbr_nids_.append(self._data.edge_index[event_idx, edge_idx].item())
-                nbr_times_.append(self._data.timestamps[event_idx])
+            # Subsample if we have more neighbours than was queried
+            if num_nbrs != -1 and len(node_nbrs) > num_nbrs:
```
shenyangHuang, line 114: there is also some online way of keeping neighbors right? some probabilistic approach that you mentioned? online way for unbiased sampling I mean. potentially a future enhancement

tgm/hooks.py
shenyangHuang: not sure I understand, to discuss

tgm/hooks.py
```diff
@@ -229,73 +260,90 @@ def __init__(self, num_nodes: int, num_nbrs: List[int]) -> None:
...
+                    seed_nodes = torch.cat([batch.src, batch.dst, batch.neg])
+                else:
+                    seed_nodes = torch.cat([batch.src, batch.dst])
+            else:
+                # TODO: Need to use nbr_nids of last hop as seed nodes for current hop
+                # But also must take into account the nbr mask from the previous hop
+                prev_hop_seed = batch.nids[-1]  # type: ignore
+                prev_hop_nbrs = batch.nbr_nids[-1][batch.nbr_mask[-1].bool()]  # type: ignore
+
+                # TODO: What is the expected behaviour when prev hop nbrs is empty?
```
shenyangHuang: if previous hop is empty, then the next hop is also empty

shenyangHuang: for most methods, they care about the sampled neighborhood for all nodes in the batch at the end. If some nodes are disconnected, i.e. no neighbor, it will just message pass to itself only.

shenyangHuang: to discuss

tgm/hooks.py
Jacob-Chmura: `end_time=dg._slice.start_time`

shenyangHuang (approved):

PR #64, opened 2025-06-25
Performance Testing Scaffolding

### LINKED ISSUE #5: Performance Testing and Instrumentation
### Purpose
The purpose of this feature is to add performance benchmark in the flavour of `pytest-benchmark` that can be added to our CI pipeline. 

**Goals**:
- PR's that touch the critical path can be evaluated for the performance impact
- Gives us observability and understanding of where we are fast and where we are slow
- It gives us the information to judge tradeoffs on data structures or implementations that improve performance

### What do we want to benchmark
We ideally want decent coverage across our critical section for the most relevant models and configs. For instance:

- Data loading / initialization
- Data discretization / coarsening for DTDG
- Raw iteration/data loading speed
- Neighbor sampling hooks
- Device hooks and cpu-gpu transfer
- Model forward and backdrop

We should run benchmarks on both CPU and GPU, though, currently only the last 2 points are actually relevant for GPU (device transfer and model forward).

Additionally, we should aim to test on graphs of various sizes and properties. Ideally, a small / medium / large sort of setup, where size is determined by things like number of timestamps, number of nodes, density of events, number of node events etc. Hopefully a subset of the TGB datasets will be more than enough for this.

### CI Integration
Small CPU benchmarks can probably run on PR as they shouldn't take more than a minute. The larger benchmarks will probably require a github CI runner with (with GPU). Plan is to set that up at Mila (#22). We can probably have it setup so that only the tests relevant to a particular code section trigger a benchmark (e.g. no need to run model tests if `nn` hasn't changed.)

Jacob-Chmura:
# Purpose
The purpose of this PR is to add scaffolding to set up performance benchmarks in the flavour of `pytest-benchmark` that can be added to our CI pipeline. 

**Goals**:
- PR's that touch the critical path can be evaluated for the performance impact
- Gives us observability and understanding of where we are fast and where we are slow
- It gives us the information to judge tradeoffs on data structures or implementations that improve performance

## Key Changes
- add `pytest-benchmark` as dev dependency
- Created several pytest markers for different-sized datasets. This enables developers to easily run performance benchmarks locally on small TGB data: `uv run pytest "small" --m --benchmark-only `, while we can run the `"medium"` and `"large"` datasets on the cluster.
- Added the simplest possible benchmark test for just loading the data into memory.

## Relevant Prs
Close #5 

## Out of scope
- Add the remainder of the test suite (discretization, data loading, hook, model forward/backwards, cpu/gpu, etc. see #5 )
- Parsing output benchmark files to show clear performance numbers
- Using a shared directory to pre-process all our datasets (#92 )

### DIFF
--- a/test/performance/test_iteration.py
+++ b/test/performance/test_iteration.py
@@ -84,12 +84,16 @@ def create_hook_manager(hooks):
 
 @pytest.mark.benchmark(group='data_loader_cpu_hooks')
 @pytest.mark.parametrize('dataset', DATASETS)
-@pytest.mark.parametrize('batch_size', [200, 'M'])
+@pytest.mark.parametrize('batch_size', [200, 'D'])
 @pytest.mark.parametrize('hook_key', list(HOOK_CONFIGS.keys()))
 def test_data_loader_cpu_hooks(
     benchmark, dataset, batch_size, hook_key, preloaded_graphs
 ):
     if dataset not in preloaded_graphs:
+        pytest.skip()
+
+    # Skip temporal batch unit with tgbn-trade which has coarse ("Y") granularity
+    if dataset == 'tgbn-trade' and isinstance(batch_size, str):
         pytest.skip()
 
     data = preloaded_graphs[dataset]['data']
@@ -102,21 +106,17 @@ def test_data_loader_cpu_hooks(
     else:
         loader = DGDataLoader(dg, batch_unit=batch_size, hook_manager=hook_manager)
 
-    total_events = 0
-    loader_iter = iter(loader)
+    def run_full_loader():
+        for _ in loader:
+            continue
 
-    def run_epoch_wrapper():
-        nonlocal total_events
-        batch = next(loader_iter)
-        total_events += len(batch.src)
+    benchmark(run_full_loader)
 
-    benchmark(run_epoch_wrapper)
-
-    throughput = (total_events / benchmark.stats['mean']) / 1e6
+    throughput = (len(dg) / benchmark.stats['mean']) / 1e6
     benchmark.extra_info.update(
         {
             'throughput_M_events_per_sec': throughput,
-            'num_events': total_events,
+            'num_events': len(dg),
         }
     )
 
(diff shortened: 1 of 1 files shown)


.github/workflows/performance.yml
```diff
@@ -0,0 +1,114 @@
...
+        with:
+          path: /tmp/.uv-cache
+          key: uv-${{ runner.os }}-${{ hashFiles('uv.lock') }}
+          restore-keys: |
+            uv-${{ runner.os }}-${{ hashFiles('uv.lock') }}
+            uv-${{ runner.os }}
+      - name: Install dependencies
+        run: uv sync --group dev
+
+      - name: Run Performance Benchmarks
```
Jacob-Chmura: This slurm script should probably be extracted to it's own file.

test/performance/conftest.py
```diff
@@ -0,0 +1,53 @@
...
+    log_dir.mkdir(parents=True, exist_ok=True)
+
+    # Save the log directory path for easy parsing in the Github action
+    latest_path_file = log_base / 'latest_path.txt'
+    latest_path_file.write_text(f'{log_dir}\n{ci_run_dir}')
+
+    return {
+        'log_dir': log_dir,
+        'project_root': Path(__file__).resolve().parents[2],
+    }
```
Jacob-Chmura: These can be simplified / removed

test/performance/test_graph_loading.py
Jacob-Chmura, line 1: TODO: Discuss which tests need to be added

test/integration/test_gclstm.py
```diff
@@ -24,7 +24,7 @@ def test_gclstm_linkprop_pred(slurm_job_runner, dataset):
 
 
 @pytest.mark.integration
-@pytest.mark.parametrize('dataset', ['tgbn-trade'])
+@pytest.mark.parametrize('dataset', ['tgbn-genre'])
```
shenyangHuang, line 27: genre will take longer since it is larger

test/integration/test_persistant_forecast.py
```diff
@@ -0,0 +1,19 @@
...
+@pytest.mark.parametrize('dataset', ['tgbl-genre'])
+@pytest.mark.slurm(
+    resources=[
+        '--partition=main',
+        '--cpus-per-task=2',
+        '--mem=4G',
+        '--time=0:10:00',
+    ]
+)
+def test_persistant_forecast_nodeprop_pred(slurm_job_runner, dataset):
```
shenyangHuang, line 14: persistent forecast the result is deterministic on a given dataset, we can directly check if the computed result is the same as what we expect.

shenyangHuang (commented):
looks good, will review again when more tests are added.

test/performance/test_construction.py
Jacob-Chmura, line 1: Arguably not worth our time, this is really fast, and mostly bottlenecked by disc io. Still, we _do_ have some pre-processing in our `DGData` constructor, so might as well leave it here.

PR #66, opened 2025-06-26
Update TGB TimeDelta Constants

### LINKED ISSUE #65: TGB Data Default Time Granularities do not match the paper

Jacob-Chmura:
# Purpose
In preparation for #40, I had a look at our TGB time granularities:

https://github.com/tgm-team/tgm/blob/5b4949e5cb06b5fe94e8f292b9412c908d7f0795/tgm/timedelta.py#L78-L99

These do not match the reported time granularities in the original data, which I have verified by manually loading in timestamps.

## Solution
I fixed the time granularities. I also removed `'tkgl-' and 'thgl-' graphs as we don't support loading them at the moment.

## Relevant Prs
Fix #65


test/test_timedelta.py
```diff
@@ -168,3 +168,20 @@ def test_time_delta_is_coarser_try_compare_ordered():
 
     with pytest.raises(ValueError):
         td1.is_coarser_than('r')
+
+
+def test_tgb_native_time_deltas():
+    exp_dict = {
+        'tgbl-wiki': TimeDeltaDG('s'),
```
shenyangHuang: looks good to me, thanks for manually verifying

Issue #35, opened 2025-06-17
RDL: Native Support for RelBench Datasets


Jacob-Chmura:

benjaminnNgo:
1. rel-hm
2. rel-f1
3. rel-salt
4. rel-event

PR #69, opened 2025-06-26
Non-Ordered TGB Timestamp Remap 

### LINKED ISSUE #40: Temporal Coarsening Time ID Remape

Jacob-Chmura:
# Purpose
The purpose of this PR is to add the logic for remapping temporal timestamps when the user specifies a non-ordered time delta type that conflicts with _tgb-native_ time units. 

For simplicity, I have forwarded the custom `time_delta` directly into `DGData.from_any()` so that future data sources (e.g. #35) can re-use this logic without polluting `kwargs` dict. 

## Key Changes
- Time delta logic is more straightforward in the `DGraph` constructor. We forward the argument and handle all conversions in `DGData` before initializing the storage
- The `DGData.from_any()` explicitly takes an optional custom `time_delta` which is used to remap timestamps in the appropriate class constructor (currently only TGB)
- Made `DGraph.time_delta` a private attribute with a getter to make it clear that the user should not set a new time_delta on the graph (this won't reflect an update in the storage and hence is a _bad state_).

### Things to keep in mind
- When working with custom datasets, we do not remap timestamps, as they lack native semantics. Instead, we just _assign_ the time delta to the time tensor.
- We don't support changing the time delta on an underlying storage. This keeps with the read-only semantics. The alternative is to create a deep copy (#67), but hopefully, we don't need this.
We throw a ValueError when trying to coarsen a graph. This requires either merging events (#67) and/or having float-timestamps with is currently not expected behaviour. We should revisit this to ensure it is not necessary.

## Relevant Prs
Close #40 

## Out of scope
- cleaning up TGB mock objects in the unit tests (they are bloated) #70 
- coarsening a graph (requires event fusion, dropping events, or something else). Here we just handle time stamp remap.


test/test_data.py
```diff
@@ -682,12 +683,160 @@ def test_from_tgb_with_node_events(mock_dataset_cls, split, expected_indices):
...
+    mock_dataset.val_mask = val_mask
+    mock_dataset.test_mask = test_mask
+    mock_dataset.num_edges = num_events
+
+    mock_dataset_cls.return_value = mock_dataset
+
+    custom_td = TimeDeltaDG('s')
+    mock_td = {'tgbn-foo': TimeDeltaDG('Y')}
+    with patch.dict('tgm.timedelta.TGB_TIME_DELTAS', mock_td):
+        data = DGData.from_tgb(name='tgbn-foo', time_delta=custom_td, split=split)
```
shenyangHuang: so here we overwrite the default time delta from tgb

shenyangHuang: this basically behaves like temporal coarsening? the timestamps would be remapped? might be good to chat about this

tgm/data.py
```diff
@@ -431,6 +439,21 @@ def _check_tgb_import() -> tuple['LinkPropPredDataset', 'NodePropPredDataset']:
...
+            tgb_time_delta = TGB_TIME_DELTAS[name]
+
+            if time_delta.is_coarser_than(tgb_time_delta):
+                raise ValueError(
+                    f"Tried to use a time_delta ({time_delta}) which is coarser than the TGB native time granularity ({tgb_time_delta}). This is undefined behaviour, either pick a finer time granularity or use an ordered time_delta ('r')"
+                )
+
+            # TODO: int conversion may need some thought
+            time_factor = int(tgb_time_delta.convert(time_delta))
+            timestamps *= time_factor
```
shenyangHuang, line 453: the timestamps will become float after this point?
timestamps should always be integers if possible. 
let's discuss this

shenyangHuang (commented):
to discuss at meeting

shenyangHuang (approved):

Issue #82, opened 2025-07-01
PermissionError in unit tests executed on a machine with Window OS


benjaminnNgo:
![Image](https://github.com/user-attachments/assets/62da0a12-ac3a-43f7-a6eb-01a0514698fe)

Jacob-Chmura:
This could be relevant: https://github.com/allenai/olmocr/issues/74

Thanks for raising

benjaminnNgo:
Hi @Jacob-Chmura!
Fixing it right now. Will create PR soon.

PR #84, opened 2025-07-01
Fixed unit tests failure due to file access permission issue on Window OS

benjaminnNgo:
# Purpose

Current unit tests are not compatible with Window OS due to the use of `NamedTemporaryFile() `. `NamedTemporaryFile()` keeps a file open and locked for exclusive access. Trying to open the temporary file with `open()` cause `PermisionError`.

# Solution

As suggested here https://github.com/allenai/olmocr/issues/74, an additional flag `delete=False` is added to each function call `NamedTemporaryFile() `. The side effect of this flag is that the temporal file will not be removed automatically. Hence, `os.remove()` is used to delete the temporary file manually.

# Related PR

Fix https://github.com/tgm-team/tgm/issues/82

### DIFF
--- a/test/unit/test_data.py
+++ b/test/unit/test_data.py
@@ -1,4 +1,5 @@
 import csv
+import os
 import tempfile
 from unittest.mock import MagicMock, patch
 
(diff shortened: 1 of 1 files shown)


Jacob-Chmura:
We may want to consider adding window's runners for our CI pipeline, to catch things like this

PR #87, opened 2025-07-02
Time Discretization

### LINKED ISSUE #86: Discretize Operation on DGraph

Jacob-Chmura:
# Purpose
The purpose of this PR is to implement temporal coarsening. We expose a new method on the `DGraph` level which creates a **deep-copy** of the underlying storage using a coarsened time granularity.

## Things to think about
- Do we preserve the graph `num_nodes` and it's implicit semantics? If the result of my coarsening is such that certain nodes no longer exist, I currently keep them as part of the graph (this has implication for static node features shape, for instance). This does not technically violate any invariants but is different to our general workflow (where number of nodes matches max node id across all events on the graph)
- We are setting up to basically clone the underlying storage on the downsampled graph. But if the user changes the storage backend they want, we'll still be constructing the one that the original graph had.
- What happens if we coarsen a sliced graph? Currently, this creates the deep copy of the full storage. But, according to this API, the user could subgraph and then coarsen. In this case, the `discretize()` method in the backend needs to also take a `DGSlice` as an argument. And, this means we have to handle the edge case of empty graphs as well (since the sliced graph can end up empty).

## Key Changes
- `DGStorage` backend API has a new function for discretizing the storage
- `DGraph` API has a new method that forwards to the storage to create a new coarsened graph
- Add info to error message if user tries to use too coarse a granualrity when constructing the graph. They should discretize after construction.
- https://github.com/tgm-team/tgm/blob/de226fb7730231445788d84c182538a61ed275d1/test/unit/test_data.py#L812 this test method got improperly merged due to automatic conflict handling. This is fixed here.

## TODO
- confirm if we need the points of confusion above
- add tests
- consider refactor to make it easier to add reduction ops in the future

## Relevant Prs
Close #86 

## Out of scope
- vectorizing the discretization implementation using tensor diffs


test/unit/test_data.py
Jacob-Chmura, line 1: These changes are patching issues that arised from incorrect auto-merge on previous PRs. This need not be reviewed.

tgm/_storage/backends/array_backend.py
```diff
@@ -19,6 +20,64 @@ def __init__(self, data: DGData) -> None:
...
+                node_mask
+            ]
+            node_ids = self._data.node_ids[node_mask]  # type: ignore
+            dynamic_node_feats = None
+            if self._data.dynamic_node_feats is not None:
+                dynamic_node_feats = self._data.dynamic_node_feats[node_mask]
+
+        static_node_feats = None
+        if self._data.static_node_feats is not None:  # Need a deep copy
+            static_node_feats = self._data.static_node_feats.clone()
```
Jacob-Chmura, line 78: Note: technically we could have way less unique node ids. If we want to pedantic, we should coarsen (reduce) the number of unique nodes. E.g. if we coarsen from second-wise to yearly data and end up with just 3 unique nodes, the static node feats be sized `3 x static_dim`, and not `|original graph| x static_dim`

shenyangHuang:
> * Do we preserve the graph `num_nodes` and it's implicit semantics? If the result of my coarsening is such that certain nodes no longer exist, I currently keep them as part of the graph (this has implication for static node features shape, for instance). This does not technically violate any invariants but is different to our general workflow (where number of nodes matches max node id across all events on the graph)

No matter how you coarsen, the number of nodes will never reduce. Theoretically this would be the case. Because when you coarsen, all you do is remove redundant edges which would originally have finer / different timestamps, now they are remapped to have the same timestamps. Therefore, if the node is in the graph then it will remain there.

shenyangHuang (approved):

PR #88, opened 2025-07-03
Enable iteration by Events for Non-ordered Graph

### LINKED ISSUE #85: Iterate by Events even when Graph has not-ordered time delta

Jacob-Chmura:
# Purpose
The purpose of this PR is to enable event-based iteration on an ordered graph. The current implementation simple enables the type of iteration where we previously raised a `ValueError`.

## Things to consider
- This iteration will be slow on CSR storage. When we get to that point, we should make it clear to users, with a warning message, for instance.

## TODO
- test to ensure no invariants are broken
- double check this is sufficient to enable #37 to go through

## Relevant Prs
Close #85


test/unit/test_data.py
Jacob-Chmura, line 1: Safe to ignore this file

shenyangHuang (approved):

PR #60, opened 2025-06-24
Dict-based Dedup Hook

Jacob-Chmura:
#### Case (`master`)
Screenshot from 2025-06-24 15-55-49

#### Control (`perf/dict_idx`)
Screenshot from 2025-06-24 15-55-57

PR #89, opened 2025-07-03
Adding TGCN for nodeproppred

benjaminnNgo:
## Objective

Adding the implementation of [T-GCN](https://arxiv.org/abs/1811.05320) and a node property prediction example for T-GCN

## New scripts

- [tgm/nn/recurrent/tgcn.py](https://github.com/tgm-team/tgm/blob/79-tgcn-nodePropPrediction/tgm/nn/recurrent/tgcn.py)
- [examples/nodeproppred/tgcn.py](https://github.com/tgm-team/tgm/blob/79-tgcn-nodePropPrediction/examples/nodeproppred/tgcn.py)
- [test/integration/test_gclstm.py](https://github.com/tgm-team/tgm/blob/79-tgcn-nodePropPrediction/test/integration/test_gclstm.py)

### Command used:
```
python .\examples\nodeproppred\tgcn.py --dataset tgbn-trade --time-gran Y --batch-time-gran Y --epochs 100 --seed 720

python .\examples\nodeproppred\gclstm.py --dataset tgbn-trade --time-gran Y --batch-time-gran Y --epochs 100 --seed 720
```
- Cross Entropy Loss & validation NDCG over epochs:

test_tgcn

- Achieved test result (no early stopping): 0.2521 NDCG on `tgbn-trade` (`GCLSTM` achieved 0.2462 NDCG)

#### TODO for another PR:

- [ ] Add link prediction example for TGCN

### DIFF
--- a/examples/nodeproppred/tgcn.py
+++ b/examples/nodeproppred/tgcn.py
@@ -1,4 +1,4 @@
-r"""python -u tgcn.py --dataset tgbn-trade --time-gran r --batch-time-gran r
+r"""python -u tgcn.py --dataset tgbn-trade --time-gran Y --batch-time-gran Y
 python -u tgcn.py --dataset tgbn-genre --time-gran s --batch-time-gran D\
 example commands to run this script.
 """
@@ -59,7 +59,7 @@ class TGCN_Model(nn.Module):
         edge_index = torch.stack([batch.src, batch.dst], dim=0)
         edge_weight = batch.edge_weight if hasattr(batch, 'edge_weight') else None  # type: ignore
         z, h_0 = self.encoder(node_feat, edge_index, edge_weight, h_0)
-        z_node = z[batch.global_to_local[batch.node_ids]]  # type: ignore
+        z_node = z[batch.global_to_local(batch.node_ids)]  # type: ignore
         pred = self.decoder(z_node)
         return pred, h_0
 
@@ -211,11 +211,10 @@ for epoch in range(1, args.epochs + 1):
     latency = end_time - start_time
 
     val_results, h_0 = eval(val_loader, model, static_node_feats, h_0)
-
     print(
         f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
         + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
     )
 
 test_results, h_0 = eval(test_loader, model, static_node_feats, h_0)
-print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
+print('Test:', ' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))


Jacob-Chmura:
> Not sure why the test check on GitHub failed

Safe to ignore these tests. There was an issue in the test cases due to auto-merging conflicts in some PRs that I pushed. I have adressed these and the fix should be in master shortly.

test/integration/test_tgcn.py
```diff
@@ -0,0 +1,26 @@
...
+@pytest.mark.slurm(
+    resources=[
+        '--partition=main',
+        '--cpus-per-task=2',
+        '--mem=4G',
+        '--time=3:00:00',
+        '--gres=gpu:a100l:1',
+    ]
+)
+def test_gclstm_nodeprop_pred(slurm_job_runner, dataset):
```
Jacob-Chmura:
```suggestion
def test_tgcn_nodeprop_pred(slurm_job_runner, dataset):
```

examples/nodeproppred/tgcn.py
```diff
@@ -0,0 +1,221 @@
+r"""python -u gclstm.py --dataset tgbn-trade --time-gran r --batch-time-gran r
+python -u gclstm.py --dataset tgbn-genre --time-gran s --batch-time-gran D\
+example commands to run this script.
+"""
```
Jacob-Chmura:
```suggestion
"""
```
I think this can be removed

examples/nodeproppred/tgcn.py
```diff
@@ -0,0 +1,221 @@
...
+    def forward(
+        self,
+        batch: DGBatch,
+        node_feat: torch.Tensor,
+        h_0: torch.Tensor | None = None,
+    ) -> Tuple[torch.Tensor, ...]:
+        edge_index = torch.stack([batch.src, batch.dst], dim=0)
+        edge_weight = batch.edge_weight if hasattr(batch, 'edge_weight') else None  # type: ignore
+        z, h_0 = self.encoder(node_feat, edge_index, edge_weight, h_0)
+        z_node = z[batch.nid_to_idx[batch.node_ids]]  # type: ignore
```
Jacob-Chmura: After merging #60, the batch attribute to use is actually `global_to_local`.
```suggestion
        z_node = z[batch.global_to_local[batch.node_ids]]  # type: ignore
```

examples/nodeproppred/tgcn.py
```diff
@@ -0,0 +1,221 @@
...
+    def __init__(self, node_dim: int, embed_dim: int) -> None:
+        super().__init__()
+        self.recurrent = TGCN(in_channels=node_dim, out_channels=embed_dim)
+        self.linear = nn.Linear(embed_dim, embed_dim)
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        edge_index: torch.Tensor,
+        edge_weight: torch.Tensor,
```
Jacob-Chmura:
```suggestion
        edge_weight: torch.Tensor | None, 
```

examples/nodeproppred/tgcn.py
```diff
@@ -0,0 +1,221 @@
...
+from tgm.util.seed import seed_everything
+
+parser = argparse.ArgumentParser(
+    description='TGCN Example',
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
+parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
+parser.add_argument('--dataset', type=str, default='tgbn-genre', help='Dataset name')
+parser.add_argument('--device', type=str, default='cpu', help='torch device')
+parser.add_argument('--epochs', type=int, default=100, help='number of epochs')
```
Jacob-Chmura: Just double checking if we want to use `100` or `10` as default number of epochs

tgm/nn/recurrent/tgcn.py
```diff
@@ -0,0 +1,157 @@
...
+        H: torch.Tensor | None = None,
+    ) -> torch.Tensor:
+        f"""Forward pass.
+
+        Args:
+            X (PyTorch Tensor): Node features.
+            edge_index (PyTorch Long Tensor): Graph edge indices.
+            edge_weight (PyTorch Long Tensor, optional): Edge weight vector.
+            H (PyTorch Tensor, optional): Hidden state matrix for all nodes.
+            C (PyTorch Tensor, optional): Cell state matrix for all nodes.
```
Jacob-Chmura:
```suggestion
```

Jacob-Chmura: Unless we need this tensor in which case example should be updated

examples/nodeproppred/tgcn.py
ntgbaoo: I would say `100` epochs

shenyangHuang:
test performance of 0.3654 is quite sensible. 
for `tgbn-trade` TGN achieves 0.374 test. So I would say the performance is quite reasonable. 
see [here](https://tgb.complexdatalab.com/docs/leader_nodeprop/#tgbn-trade) for the leaderboard.

tgm/nn/recurrent/tgcn.py
```diff
@@ -0,0 +1,156 @@
...
+
+    Reference: https://arxiv.org/abs/1811.05320
+    """
+
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        improved: bool = False,
+        cached: bool = False,
```
shenyangHuang, line 26: when would you want to turn this one? what would be the benefit?

examples/nodeproppred/tgcn.py
shenyangHuang: for DTDG methods usually takes more epochs

shenyangHuang (commented):
performance looks good.

tgm/nn/recurrent/tgcn.py
ntgbaoo: `improved` and `cached` are both used to configure GCNConv. By default, they are set to `False`. From what I am understanding from reading the [documentation from PyG](https://pytorch-geometric.readthedocs.io/en/2.5.2/generated/torch_geometric.nn.conv.GCNConv.html).

- If we set `improved = True` , the self-loops are added `A+2I` instead of `A+I`. This increases the weight of self-loops, giving each node’s own features more influence during aggregation.
- If we set `cached = True`, the layer computes the normalized adjacency matrix only once, and reuses it. This speeds up training and inference but is only suitable for transductive learning. Scenarios where `cached` can be set to `True` include tasks such as traffic forecasting, where graph structure is assumed to be static and only node features are dynamic. `cached` should not be set to `False` when graph structure dynamically changes.

Thanks for pointing this out. These are all useful information that we should include in the comment section.

benjaminnNgo:
I added further information about `cached` and `improved` in the comment section and increased the default number of epochs as you suggested, @shenyangHuang. Please take a look whenever you have a chance.

tgm/nn/recurrent/tgcn.py
shenyangHuang: yeah thanks for adding the comments here, much more clear now

Jacob-Chmura (approved):

shenyangHuang (approved):

shenyangHuang:
The new performance makes sense to me, I got similar range from GCN yesterday for nodeproppred.

PR #95, opened 2025-07-03
Fix Recursive Batch Attribute Hook Application

### LINKED ISSUE #93: TGB Negative Batch List Device Bug

Jacob-Chmura:
# Purpose
Issue in TGB TGAT example came up while doing integration testing on gpu. I originally thought this was an issue in the hooks, but it seems like just a problem in the examples. 

## Solution
I ended up adding a helper for applying generic tensor functions recusrvely to nested tensor structures. This can be reused for other hooks in the future. It's strictly not necessary, but basically a free refactor on top of the core fix, which is:

https://github.com/tgm-team/tgm/blob/d0bfc7c0ff4269a124a84fd7ad749669b0dc45a3/examples/linkproppred/TGB/tgat.py#L177

## Relevant Prs
Fix #93

### DIFF
--- a/tgm/hooks.py
+++ b/tgm/hooks.py
@@ -157,7 +157,7 @@ class NegativeEdgeSamplerHook:
     # TODO: Historical vs. random
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
         size = (round(self.neg_sampling_ratio * batch.dst.size(0)),)
-        batch.neg = torch.randint(self.low, self.high, size, device=dg.device)  # type: ignore
+        batch.neg = torch.randint(self.low, self.high, size)  # type: ignore
         return batch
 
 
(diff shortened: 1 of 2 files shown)


tgm/hooks.py
```diff
@@ -378,3 +385,25 @@ def _update(self, batch: DGBatch) -> None:
             if batch.edge_feats is not None:
                 self._nbrs[src_nbr][2].append(batch.edge_feats[i])
                 self._nbrs[dst_nbr][2].append(batch.edge_feats[i])
+
+
+def _apply_to_tensors_inplace(obj: Any, fn: Any) -> Any:
```
ntgbaoo, line 413: Do we need to do a unit test for this function? I can work on it to help with code coverage

Jacob-Chmura: Sure! We can change the method to make it public and add a test case for each branch (each type of nested data structure.

Feel free to commit directly to this branch and we can merge it when you're done.

PR #96, opened 2025-07-03
Vectorized Discrization Implementation

### LINKED ISSUE #90: Vectorize Temporal Coarsening

Jacob-Chmura:
# Purpose
The purpose of this PR is to make an implementation change in our COO discretize method to use vectorization instead of native python loop.

Specifically, I just used a basic `diff array` technique to extract the first event position that we need to capture; first unique edge pair (if edge event), or node (if node event) in each bucket.

### Outcome
TODO

## Relevant Prs
Close #90

### DIFF
--- a/tgm/_storage/backends/array_backend.py
+++ b/tgm/_storage/backends/array_backend.py
@@ -1,5 +1,5 @@
 import random
-from typing import Any, Dict, List, Literal, Optional, Set, Tuple
+from typing import Dict, List, Literal, Optional, Set, Tuple
 
 import torch
 from torch import Tensor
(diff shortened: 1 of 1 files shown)


tgm/_storage/backends/array_backend.py
```diff
@@ -38,19 +36,12 @@ def discretize(
...
-                    prev_bucket = bucket
-                if node_or_edge not in seen:
-                    seen.add(node_or_edge)
-                    keep.append(i)
-            return torch.tensor(keep, dtype=torch.long)
+            ids_flat = ids.view(ids.size(0), -1)
+
+            combined = torch.cat([event_buckets.unsqueeze(1), ids_flat], dim=1)
+            keep_mask = torch.ones(combined.size(0), dtype=torch.bool)
+            keep_mask[1:] = (combined[1:] != combined[:-1]).any(dim=1)
```
shenyangHuang, line 43: if I understand correctly, you always keep the first edge in a bucket since there it must be new and kept. For remaining edges, you keep the unique edges

PR #97, opened 2025-07-03
Better Error Messages

### LINKED ISSUE #21: Verify and refine user-side warnings and input validation

Jacob-Chmura:
# Purpose
The purpose of this PR is to have better error messages for the user throughout the codebase, where it is currently lacking in information. Most of our errors are already pretty helpful, but I am adding some more in prep for alpha release.

## Key Changes
- make materialized graph edges into contiguous tensors. This should have performance improvements in the code. It was triggering an internal torch c++ warning for operating on strided data, which is why I addressed it here.
- more verbose error messaging throughout
- silenced sort warning in `DGData`. This is because it makes the implementation much easier and avoids double sorting when user specifies node events. Unless user specifies raw tensor data, we create copies of the underlying tensors so I don't see the mutation being an issue in the research setting.

## Relevant Prs
Fix #21

### DIFF
--- a/tgm/data.py
+++ b/tgm/data.py
@@ -29,42 +29,39 @@ class DGData:
     static_node_feats: Tensor | None = None  # [num_nodes, D_node_static]
 
     def __post_init__(self) -> None:
+        def _assert_is_tensor(x: Any, name: str) -> None:
+            if not isinstance(x, Tensor):
+                raise TypeError(f'{name} must be a Tensor, got: {type(x)}')
+
         def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
             int_types = [torch.int8, torch.int16, torch.int32, torch.int64, torch.uint8]
             if x.dtype not in int_types:
                 raise TypeError(f'{name} must have integer dtype but got: {x.dtype}')
 
         # Validate edge index
-        if not isinstance(self.edge_index, Tensor):
-            raise TypeError(f'edge_index must be Tensor, got: {type(self.edge_index)}')
+        _assert_is_tensor(self.edge_index, 'edge_index')
+        _assert_tensor_is_integral(self.edge_index, 'edge_index')
         if self.edge_index.ndim != 2 or self.edge_index.shape[1] != 2:
             raise ValueError(
                 f'edge_index must have shape [num_edges, 2], got: {self.edge_index.shape}',
             )
-        _assert_tensor_is_integral(self.edge_index, 'edge_index')
 
         num_edges = self.edge_index.shape[0]
         if num_edges == 0:
             raise ValueError('empty graphs not supported')
 
         # Validate edge event idx
-        if not isinstance(self.edge_event_idx, Tensor):
-            raise TypeError(
-                f'edge_event_idx must be Tensor, got: {type(self.edge_event_idx)}'
-            )
+        _assert_is_tensor(self.edge_event_idx, 'edge_event_idx')
+        _assert_tensor_is_integral(self.edge_event_idx, 'edge_event_idx')
         if self.edge_event_idx.ndim != 1 or self.edge_event_idx.shape[0] != num_edges:
             raise ValueError(
                 'edge_event_idx must have shape [num_edges], '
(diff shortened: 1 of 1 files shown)


tgm/_storage/backends/array_backend.py
```diff
@@ -126,6 +126,8 @@ def get_edges(self, slice: DGSliceTracker) -> Tuple[Tensor, Tensor, Tensor]:
         edges = self._data.edge_index[edge_mask]
         src, dst = edges[:, 0], edges[:, 1]
         time = self._data.timestamps[self._data.edge_event_idx[edge_mask]]
+
+        src, dst, time = src.contiguous(), dst.contiguous(), time.contiguous()
```
shenyangHuang, line 130: interesting, potentially some other tensors can be continguous too?

tgm/data.py
```diff
@@ -94,30 +117,36 @@ def __post_init__(self) -> None:
...
+            raise TypeError(
+                f'timestamps must be a Tensor, got: {type(self.timestamps)}'
+            )
         if (
             self.timestamps.ndim != 1
             or self.timestamps.shape[0] != num_edges + num_node_events
         ):
-            raise ValueError('timestamps must have shape [num_edges + num_node_events]')
+            raise ValueError(
+                'timestamps must have shape [num_edges + num_node_events], '
```
shenyangHuang, line 141: actually does the user ever create the combined edge event and node event timestamp tensor explicitly? Don't we always create it ourselves by combining node and edge events? Then the error should be raised when we check for node and edge event timestamp tensor, if those are correct then this must be correct. What I am saying is the error message is not explicit enough for user to fix as it can come from edge or node event timestamps because it is the sum here.

tgm/data.py
```diff
@@ -94,30 +117,36 @@ def __post_init__(self) -> None:
...
+                f'got {num_edges} edges, {num_node_events} node_events, shape: {self.timestamps.shape} '
+            )
         if not torch.all(self.timestamps >= 0):
-            raise ValueError('timestamps must be non-negative integers')
+            raise ValueError('timestamps must all be non-negative')
```
shenyangHuang: we allow floating point timestamps currently?

tgm/hooks.py
```diff
@@ -23,8 +23,12 @@ class HookManager:
     def __init__(self, dg: DGraph, hooks: List[DGHook]) -> None:
         if not isinstance(hooks, list):
             raise TypeError(f'Invalid hook type: {type(hooks)}')
-        if not all(isinstance(h, DGHook) for h in hooks):
-            raise TypeError('All items in hook list must follow DGHook protocol')
+        bad_hook_names = [type(h).__name__ for h in hooks if not isinstance(h, DGHook)]
+        if len(bad_hook_names):
+            raise TypeError(
+                f'Some hooks do not correctly implement the DGHook protocol: {bad_hook_names}, '
```
shenyangHuang: instead of saying some hooks, should we raise error of which specific hook doesn't follow the protocol to be even more specific? especially if for some reason, the user created like three custom hook, say their own custom negative sampling hook, neighbor sampling hook and then another positonal encoding hook or sth, won't know which is causing the error

shenyangHuang: we do have the `bad_hook_names` already afterall

shenyangHuang (changes requested):
request small changes to make the error messages even more specific, looks great overall

tgm/_storage/backends/array_backend.py
Jacob-Chmura: This is the only part of our API where we return a dense tensor (since features are currently being returned sparse coo)

tgm/data.py
Jacob-Chmura: I can revert, technically, this would probably fail. But I changed this error messages since we don't actually check for integers

tgm/hooks.py
Jacob-Chmura: This is precisely what I am doing.

Jacob-Chmura: Maybe I'll change the word `Some` to `These` to make that more clear

PR #100, opened 2025-07-04
Bump version

### LINKED ISSUE #98: Bump Alpha Release Version

Jacob-Chmura:
# Purpose
Updating the version number from `0.0.1-alpha` to `0.1.0a0` which is the standardized alpha tag under semantic versioning. Secure `py-tgm` as our pip package name.

## Key Changes
- change version tags
- update some missing fields in our `pyproject.toml`

## Relevant Prs
Fix #98


pyproject.toml
```diff
@@ -1,15 +1,9 @@
...
-
-[tool.uv.pip]
-torch-backend = "auto"
-
 [project]
-name = "tgm"
-version = "0.0.1-alpha"
+name = "py-tgm"
+version = "0.1.0a0"
 description = "Efficient and Modular ML on Dynamic Graphs"
```
shenyangHuang: should change the title to say Temporal Graph everywhere? given our library is TGM so we should stick with temporal graph

pyproject.toml
```diff
@@ -1,15 +1,9 @@
...
 readme = "README.md"
+license = {text = "MIT"}
```
shenyangHuang, line 6: not familiar with license choices, whichever makes sense to you can go here.

pyproject.toml
```diff
@@ -43,6 +37,19 @@ examples = [
...
+[tool.uv.pip]
+torch-backend = "auto"
+
+[tool.uv.project]
+root-package = "tgm"
+
+[tool.flit.module]
+name = "tgm"
+
 [tool.mypy]
```
shenyangHuang, line 53: people installing from pip won't need mypy right?

shenyangHuang (approved):

pyproject.toml
Jacob-Chmura: will do

pyproject.toml
Jacob-Chmura: Yep this is for our own dev tooling, under optional dev dependancy.

PR #110, opened 2025-07-06
Debug/mrr

Jacob-Chmura:


pyproject.toml
```diff
@@ -44,9 +44,6 @@ build-backend = "flit_core.buildapi"
 [tool.uv.pip]
 torch-backend = "auto"
 
-[tool.uv.project]
```
shenyangHuang, line 47: what does this do?

examples/linkproppred/TGB/tgat.py
```diff
@@ -89,21 +87,23 @@ def forward(self, batch: DGBatch) -> Tuple[torch.Tensor, torch.Tensor]:
         for hop in reversed(range(self.num_layers)):
             seed_nodes = batch.nids[hop]
             nbrs = batch.nbr_nids[hop]
-            nbr_mask = batch.nbr_mask[hop]
+            nbr_mask = batch.nbr_mask[hop].bool()
```
shenyangHuang, line 90: for the layers, potentially we should add type check on the inputs, just to prevent any other type bugs in the future

shenyangHuang: might help us debug as well

shenyangHuang:
also, merged TGN example from #109

PR #119, opened 2025-07-10
GCN TGB MRR

shenyangHuang:
GCN is now training correctly on TGB, MRR improves after each epoch
- [ ] need to check final performance after 100 epoch


examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+)
+parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
+parser.add_argument('--dataset', type=str, default='tgbl-wiki', help='Dataset name')
+parser.add_argument('--device', type=str, default='cpu', help='torch device')
+parser.add_argument('--epochs', type=int, default=10, help='number of epochs')
+parser.add_argument('--lr', type=float, default=0.0001, help='learning rate')
+parser.add_argument('--dropout', type=str, default=0.1, help='dropout rate')
+parser.add_argument('--n-layers', type=int, default=2, help='number of GCN layers')
+parser.add_argument('--embed-dim', type=int, default=128, help='embedding dimension')
+parser.add_argument('--bsize', type=int, default=200, help='batch size')
```
Jacob-Chmura: Should we paratermize the snapshot batch size here as well? (`h`)

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+import time
+from typing import Tuple
+
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from tgb.linkproppred.dataset_pyg import PyGLinkPropPredDataset
+from tgb.linkproppred.evaluate import Evaluator
+from torch_geometric.nn import GCNConv
```
Jacob-Chmura, line 15: This needs to be added as a dependency to `examples` group in `pyproject`

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+        opt.step()
+
+        # update the model if the prediction batch has moved to next snapshot.
+        while (
+            batch.time[-1] > (snapshot_batch.time[-1] + 1) * 3600
+        ):  # if batch timestamps greater than snapshot, process the snapshot
+            try:
+                snapshot_batch = next(iter_loader)
+            except StopIteration:
+                pass
```
Jacob-Chmura, line 165: Should this `break`? Or it is always going to correctly exit the while loop

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+    encoder: nn.Module,
+    decoder: nn.Module,
+    opt: torch.optim.Optimizer,
+    static_node_feat: torch.Tensor,
+) -> float:
+    encoder.train()
+    decoder.train()
+    total_loss = 0
+    criterion = torch.nn.MSELoss()
+    iter_loader = iter(snapshots_loader)
```
Jacob-Chmura:
```suggestion
    snapshots_iterator = iter(snapshots_loader)
```

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+    decoder: nn.Module,
+    eval_metric: str,
+    evaluator: Evaluator,
+    static_node_feat: torch.Tensor,
+    z: torch.Tensor,
+) -> dict:
+    encoder.eval()
+    decoder.eval()
+    perf_list = []
+    iter_loader = iter(snapshots_loader)
```
Jacob-Chmura:
```suggestion
    snapshots_iterator = iter(snapshots_loader)
```

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+dataset = PyGLinkPropPredDataset(name=args.dataset, root='datasets')
+eval_metric = dataset.eval_metric
+neg_sampler = dataset.negative_sampler
+evaluator = Evaluator(name=args.dataset)
+dataset.load_val_ns()
+dataset.load_test_ns()
+
+train_dg = DGraph(
+    args.dataset,
+    time_delta=TimeDeltaDG('r'),
```
Jacob-Chmura, line 233:
```suggestion
```

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+train_snapshots = DGraph(
+    args.dataset,
+    time_delta=TimeDeltaDG('s'),
+    split='train',
+    device=args.device,
+).discretize('h')
+
+val_dg = DGraph(
+    args.dataset,
+    time_delta=TimeDeltaDG('r'),
```
Jacob-Chmura, line 247:
```suggestion
```

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+val_snapshots = DGraph(
+    args.dataset,
+    time_delta=TimeDeltaDG('s'),
+    split='val',
+    device=args.device,
+).discretize('h')
+
+test_dg = DGraph(
+    args.dataset,
+    time_delta=TimeDeltaDG('r'),
```
Jacob-Chmura, line 261:
```suggestion

```

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+    split='train',
+    device=args.device,
+)
+
+train_snapshots = DGraph(
+    args.dataset,
+    time_delta=TimeDeltaDG('s'),
+    split='train',
+    device=args.device,
+).discretize('h')
```
Jacob-Chmura: Consider making the discretization unit a command line argument for compatibility with other datasets

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+val_loader = DGDataLoader(
+    val_dg,
+    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')],
+    batch_size=args.bsize,
+)
+val_snapshots_loader = DGDataLoader(
+    val_snapshots,
+    hook=[],
```
Jacob-Chmura:
```suggestion
```

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+)
+
+test_loader = DGDataLoader(
+    test_dg,
+    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')],
+    batch_size=args.bsize,
+)
+test_snapshots_loader = DGDataLoader(
+    test_snapshots,
+    hook=[],
```
Jacob-Chmura:
```suggestion
```

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+    batch_unit='h',
```
Jacob-Chmura: Iteration unit can be it's own argument as well

examples/linkproppred/TGB/gcn.py
```diff
@@ -0,0 +1,335 @@
...
+        static_node_feats,
+        z,
+    )
+    print(
+        f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
+        + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
+    )
+
+# test_results = eval(test_loader, model, test_metrics, static_node_feats)
+# print(' '.join(f'{k}={v.item():.4f}' for k, v in test_results.items()))
```
Jacob-Chmura: This should work now?

Jacob-Chmura (changes requested):
Nice, when we are done with the changes in this file, can you add it to integration test setup and trigger ci on the cluster

examples/linkproppred/TGB/gcn.py
shenyangHuang: yes, good point

examples/linkproppred/TGB/gcn.py
shenyangHuang: done

examples/linkproppred/TGB/gcn.py
shenyangHuang: it shouldn't break, it is just that this would happen when you are at the last few batches and there are not more new snapshots for the GCN to process, but it still want the function to finish running. 

(assuming a CTDG time granularity of seconds (here hour for simplicity) and daily snapshots for GCN)
Say last few batches is at Monday 1am, 2am and 3am
the last snapshot that GCN uses for prediction is Sunday snapshot
at Monday 1am batch, the loop here will be returning exception thus no more new snapshots, thus the model will use Sunday snapshot for 1am, 2am and 3am batches but you shouldn't break (since you need the prediction).

examples/linkproppred/TGB/gcn.py
shenyangHuang: yes, added similar syntax with nodeproppred

examples/linkproppred/TGB/gcn.py
shenyangHuang: no, actually it should be the same as the discretization granularity, at least for UTG examples I had in mind

examples/linkproppred/TGB/gcn.py
shenyangHuang: yes, updated

shenyangHuang:
added to integration test setup, can you check if it is correct @Jacob-Chmura

Jacob-Chmura (approved):
Lets just double check the integration tests work and then we can merge

PR #121, opened 2025-07-13
DyGFormer implementation

benjaminnNgo:


benjaminnNgo:
@shenyangHuang @Jacob-Chmura, I tried to follow DyGFormer's original implementation as closely as possible and improved the readability where I could, prioritizing the reproduction of the results reported in the paper. That said, I think there’s still room for improvement in the code. Let me know if you have any ideas on how to make it better.

Andy suggested making `NeighborCooccurrenceEncoder` a hook, but I don’t think that’s a good idea since it’s a trainable component. However, we could consider turning the `_count_nodes_freq()` function of `NeighborCooccurrenceEncoder` into a hook instead. What do you think?

Andy also suggested replacing the inputs of the `forward` function: `neighbours`, `neighbours_time`, and `neighbours_edge_feat`, with a single DGBatch object. However, from the model’s perspective, I believe `DGBatch` provides more information than `DyGFormer` actually needs to make predictions. All it requires is the edge, along with some information about the neighbours of the `src` and `dst` nodes. DyGFormer doesn't need access to the negative destination nodes given by `DGBatch`

examples/linkproppred/dygformer.py
```diff
@@ -0,0 +1,309 @@
+r"""python -u tgcn.py --dataset tgbn-trade --time-gran Y --batch-time-gran Y
+python -u tgcn.py --dataset tgbn-genre --time-gran s --batch-time-gran D\
+example commands to run this script.
+"""
```
Jacob-Chmura:
```suggestion
"""
```

examples/linkproppred/dygformer.py
```diff
@@ -0,0 +1,309 @@
...
+            batch.nbr_feats[0],
+        )
+
+        out = self.decoder(src_embeddings, dst_embeddings)
+        return out
+
+
+def _init_hooks(dg: DGraph, sampling_type: str) -> List[DGHook]:
+    if sampling_type == 'uniform':
+        nbr_hook = NeighborSamplerHook(num_nbrs=args.n_nbrs)
```
Jacob-Chmura: We should probably remove `--sampling` as an argument since this would break (`args.n_nbrs`) not defined, and dygformer specializes to recency hook

examples/linkproppred/dygformer.py
```diff
@@ -0,0 +1,309 @@
...
+            dropout,
+            max_input_sequence_length,
+            num_channels,
+            time_encoder,
+            device,
+        )
+        self.decoder = LinkPredictor(output_dim)
+
+    def forward(self, edge_idx: torch.Tensor, batch: DGBatch, X: torch.Tensor):
+        src_embeddings, dst_embeddings = self.encoder(
```
Jacob-Chmura: Nit: use same variable names as in other examples (`z_src, z_dst`), not a big deal though

examples/linkproppred/dygformer.py
```diff
@@ -0,0 +1,309 @@
...
+    type=str,
+    default='s',
+    help='raw time granularity for dataset',
+)
+parser.add_argument(
+    '--batch-time-gran',
+    type=str,
+    default='D',
+    help='time granularity to operate on for snapshots',
+)
```
Jacob-Chmura: Does the original paper discretize to daily snapshots? Or just iterate by batching of 200 edges. If the later, I would drop these and mimic our arguments for tgat: https://github.com/tgm-team/tgm/blob/eda29bb9e7c4990f15ca5646d31233a3325433c1/examples/linkproppred/tgat.py#L29

examples/linkproppred/dygformer.py
```diff
@@ -0,0 +1,309 @@
...
+    total_loss = 0
+    for batch in tqdm(loader):
+        opt.zero_grad()
+        src = batch.src
+        dst = batch.dst
+        neg = batch.neg
+        edge_idx_pos = torch.stack([src, dst], dim=1).t()
+        edge_idx_neg = torch.stack([src, neg], dim=1).t()
+        pos_out = model(edge_idx_pos, batch, node_feat)
+        neg_out = model(edge_idx_neg, batch, node_feat)
```
Jacob-Chmura: The `DygFormer_LinkPrediction` model wrapper _should_ operate directly on the batch. Hence, this should be moved to the forward pass. TGAT skeleton could be a good reference starting point: https://github.com/tgm-team/tgm/blob/eda29bb9e7c4990f15ca5646d31233a3325433c1/examples/linkproppred/tgat.py#L82

Jacob-Chmura: Nit: this may be slightly more efficient instead of transposing: `torch.stack((src, dst), dim=0)`

examples/linkproppred/dygformer.py
```diff
@@ -0,0 +1,309 @@
...
+    train_loader = DGDataLoader(
+        train_dg,
+        batch_unit=args.batch_time_gran,
+        hook=_init_hooks(dg=train_dg, sampling_type=args.sampling),
+    )
+    val_loader = DGDataLoader(
+        val_dg,
+        batch_unit=args.batch_time_gran,
+        hook=_init_hooks(dg=val_dg, sampling_type=args.sampling),
+    )
```
Jacob-Chmura: This is a current design limitation, but we don't have an api for resetting neighbor state (for recency loader). This means there could be info leakage across epochs. The band aid fix for now is to re-initialize the data loader across each epoch to force a fresh neighbor state. E.g: https://github.com/tgm-team/tgm/blob/eda29bb9e7c4990f15ca5646d31233a3325433c1/examples/linkproppred/tgat.py#L220-L231

examples/linkproppred/dygformer.py
```diff
@@ -0,0 +1,309 @@
...
+        val_results = eval(evaluator, val_loader, model, val_metrics, static_node_feats)
+        print(
+            f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
+            + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
+        )
+
+        test_results = eval(
+            evaluator, test_loader, model, test_metrics, static_node_feats
+        )
+        print('Test:', ' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
```
Jacob-Chmura: This can be moved out of the loop

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+        self,
+        node_feat_dim: int,
+        edge_feat_dim: int,
+        time_feat_dim: int,
+        channel_embedding_dim: int,
+        output_dim: int = 172,
+        patch_size: int = 1,
+        num_layers: int = 2,
+        num_heads: int = 2,
+        dropout: float = 0.1,
```
Jacob-Chmura, line 314: Are there any constraints on these (e.g. patch size being a multiple of some other spec? If yes, we should validarte the user provided dimensions, if not, ignore this comment

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+
+    Args:
+        feat_dim (int): dimension of neighbor co-occurrence features (encodings).
+        device (str): Device (cpu or gpu)
+
+    Reference: https://arxiv.org/abs/2303.13047.
+    """
+
+    def __init__(self, feat_dim: int, device: str) -> None:
+        super(NeighborCooccurrenceEncoder, self).__init__()
```
Jacob-Chmura: Nit
```suggestion
        super().__init__()
```

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+
+        Args:
+            src_neighbour_nodes_ids (Tensor): Padded list of source node's neighbour.
+            dst_neighbour_nodes_ids (Tensor): Padded list of destination node's neighbour.
+
+        Returns:
+            X (PyTorch Float Tensor): Neighbor co-occurrence features (`X^{t}_{*,C}`).
+        """
+        src_neighbour_nodes_ids_np = src_neighbour_nodes_ids.cpu().numpy()
+        dst_neighbour_nodes_ids_np = dst_neighbour_nodes_ids.cpu().numpy()
```
Jacob-Chmura, line 204: I understand we were primarily concerned with reproducing exisisting results. Once we are confident in the implementation, we should keep these as tensors on device instead of moving them between cpu/gpu

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+        )
+
+    def _count_nodes_freq(
+        self, all_sources_neighbors: np.ndarray, all_dsts_neighbors: np.ndarray
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        assert len(all_sources_neighbors.shape) == 2
+        assert (
+            all_sources_neighbors.shape[0] == all_dsts_neighbors.shape[0]
+            and all_sources_neighbors.shape[1] == all_dsts_neighbors.shape[1]
+        )
```
Jacob-Chmura:
```suggestion
        assert all_sources_neighbors.ndim == 2 and all_sources_neighbors.shape == all_dsts_neighbors.shape

```

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+
+        source_freq, dst_freq = [], []
+        for src_neighbors, dst_neighbors in zip(
+            all_sources_neighbors, all_dsts_neighbors
+        ):
```
Jacob-Chmura, line 48: My understanding is that we do symmetric ops on src/dst. Can we write an embedded helper function in here that we call once for src,dst, and once for dst,src. Then it might be easier to understand and optimize the underlying computation

Jacob-Chmura: It will probably simpler if we keep all on tensors as well (torch.unique should be drop in replacement I think)

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+        """
+
+        src, dst = edge_index[0], edge_index[1]
+        num_edge = src.shape[0]
+        src_neighbours = neighbours[:num_edge]
+        dst_neighbours = neighbours[num_edge : num_edge * 2]
+        src_neighbours_time = neighbours_time[:num_edge]
+        dst_neighbours_time = neighbours_time[num_edge : num_edge * 2]
+        src_neighbours_edge_feat = neighbours_edge_feat[:num_edge]
+        dst_neighbours_edge_feat = neighbours_edge_feat[num_edge : num_edge * 2]
```
Jacob-Chmura, line 415: We need to be very careful about the implicitly assumed neighbour ids. This works as is, but it's not currently enforced in any way. So just good to take a mental note or add a comment somewhere that assumes nbr ids are contiguously added for src/dst.

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+        # include seed nodes are neighbors themselves
+        src_neighbours = torch.cat([src.unsqueeze(dim=1), src_neighbours], dim=1)
+        dst_neighbours = torch.cat([dst.unsqueeze(dim=1), dst_neighbours], dim=1)
+
+        src_neighbours_time = torch.cat(
+            [edge_time.unsqueeze(1), src_neighbours_time], dim=1
+        )
+        dst_neighbours_time = torch.cat(
+            [edge_time.unsqueeze(1), dst_neighbours_time], dim=1
+        )
```
Jacob-Chmura: Nit:

In my opinion, 
`src_neighbors = torch.cat([src[:, None], src_neighbors], dim=1)` is slightly simpler to read than `torch.cat([src.unsqueeze(dim=1), src_neighbours], dim=1)` but not a big deal

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+            src_neighbours_edge_feat.shape[2],
+            device=self.device,
+            dtype=src_neighbours_edge_feat.dtype,
+        )
+        src_neighbours_edge_feat = torch.cat([padding, src_neighbours_edge_feat], dim=1)
+        dst_neighbours_edge_feat = torch.cat([padding, dst_neighbours_edge_feat], dim=1)
+
+        # Get node feat and time feat using Time Encoder
+        src_neighbours_node_feats = self._get_node_features(X, src_neighbours)
+        dst_neighbours_node_feats = self._get_node_features(X, dst_neighbours)
```
Jacob-Chmura: So, this appears to me that we only ever use the global node features (X) here. But this is just reading the node features for the current batch nbrs. So, the good news is we can refactor this to avoid passing around the global node tensor and instead work directly on the batch. I suggest we refactor this accordingly if possible, and we won't need the `self._get_node_features()` method anymore

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,602 @@
...
+            dst_neighbours_time_features_patches,
+            dst_co_occurence_features_patches,
+        ) = self._get_patches(
+            dst_neighbours_node_feats,
+            dst_neighbours_edge_feat,
+            dst_neighbours_time_feats,
+            dst_co_occurrence_feats,
+        )
+
+        # Use projection to align the patch encoding dimension for both dst and src
```
Jacob-Chmura, line 483: A lot of things are symmetric on src and dst. Could we concat them and process them in a large batch?

examples/linkproppred/dygformer.py
ntgbaoo: Agreed. Resolved

examples/linkproppred/dygformer.py
ntgbaoo: Actually, I didn't realize that I discretized to daily snapshots. Changed to mimic TGAT.

examples/linkproppred/dygformer.py
ntgbaoo: I see. Thank you for pointing this out.

examples/linkproppred/dygformer.py
ntgbaoo: Yes. This will be moved out of the loop eventually. I tested every epoch to debug actually.

tgm/nn/model/dygformer.py
ntgbaoo: I agreed. I refactored `_get_patches` as follows:
```
def _get_patches(self,feat: torch.Tensor)-> torch.Tensor:
        list_patches = []
        for patch_id in range(self.num_patches):
            start_idx = patch_id * self.patch_size
            end_idx = patch_id * self.patch_size + self.patch_size
            list_patches.append(
                feat[:, start_idx:end_idx, :]
            )

        patches_feats = torch.stack(
            list_patches, dim=1
        ).reshape(feat.shape[0], self.num_patches, self.patch_size * feat.shape[2])

        return patches_feats
```

and the usage

```
# Get patches for each features of src and dst
        neighbours_node_feats = self._get_patches(torch.cat([src_neighbours_node_feats,dst_neighbours_node_feats],dim=0))
        neighbours_edge_feats = self._get_patches(torch.cat([src_neighbours_edge_feat,dst_neighbours_edge_feat],dim=0))
        neighbours_time_feats = self._get_patches(torch.cat([src_neighbours_time_feats,dst_neighbours_time_feats],dim=0))
        co_occurrence_feats = self._get_patches(torch.cat([src_co_occurrence_feats,dst_co_occurrence_feats],dim=0))
        
      src_neighbours_node_features_patches = neighbours_node_feats[:batch_size]
      src_neighbours_edge_features_patches = neighbours_edge_feats[:batch_size]
      src_neighbours_time_features_patches = neighbours_time_feats[:batch_size]
      src_co_occurence_features_patches = co_occurrence_feats[:batch_size]

      dst_neighbours_node_features_patches = neighbours_node_feats[batch_size:]
      dst_neighbours_edge_features_patches = neighbours_edge_feats[batch_size:]
      dst_neighbours_time_features_patches = neighbours_time_feats[batch_size:]
      dst_co_occurence_features_patches = co_occurrence_feats[batch_size:]
```

tgm/nn/model/dygformer.py
ntgbaoo: I tried to change all operations to have them work on a tensor [in this function](https://github.com/tgm-team/tgm/blob/6-DygformerImplementation/tgm/nn/model/dygformer.py#L114C1-L186C83).
However, it runs significantly slower. 

I think this specific part makes it slower: https://github.com/tgm-team/tgm/blob/6-DygformerImplementation/tgm/nn/model/dygformer.py#L148C13-L152C14

Do you have any suggestions on optimizing this @Jacob-Chmura?

tgm/nn/model/dygformer.py
ntgbaoo: I am not sure why [Format and Linting / ruf](https://github.com/tgm-team/tgm/actions/runs/16821311063/job/47648595306?pr=121) is failing. I have pre-commit installed and the code is refactored by the tool every time I commit. It should not fail right?

tgm/nn/model/dygformer.py
Jacob-Chmura: Actually, I think it makes some sense that copying the implementation directly to torch will be slightly slower due to cpu-gpu sync and kernel launch overhead. I would expect that a fully vectorized approach is competitive (one where we do not loop over batch index). But this may require riskier changes that are out of scope.

TLDR, we can just keep numpy for now as its closest to the implementatino that we know is correct.  @benjaminnNgo

examples/linkproppred/dygformer.py
```diff
@@ -0,0 +1,282 @@
...
+        return pos_out, neg_out
+
+
+def _init_hooks(dg: DGraph) -> List[DGHook]:
+    nbr_hook = RecencyNeighborHook(
+        num_nbrs=[args.max_sequence_length - 1],  # 1 remaining for seed node itself
+        num_nodes=dg.num_nodes,
+        edge_feats_dim=dg.edge_feats_dim,
+    )
+    neg_hook = NegativeEdgeSamplerHook(low=0, high=dg.num_nodes)
```
Jacob-Chmura: This is subtle, but can we use the following logic: https://github.com/tgm-team/tgm/blob/bd9872766938cbd95828d6413fac509f40384078/examples/linkproppred/TGB/tgat.py#L233-L235

which ensures that our negative ids are sampled within the _dst_ node id range (impacts bipartite graphs where node id ranges have implicit meaning, e.g. first 10k node ids are all users, last 10k are all products).

test/integration/test_dygformer.py
```diff
@@ -0,0 +1,26 @@
...
+@pytest.mark.slurm(
+    resources=[
+        '--partition=main',
+        '--cpus-per-task=2',
+        '--mem=8G',
+        '--time=3:00:00',
+        '--gres=gpu:a100l:1',
+    ]
+)
+def test_tgat_recency_sampler_linkprop_pred(slurm_job_runner, dataset):
```
Jacob-Chmura:
```suggestion
def test_dygformer_linkprop_pred(slurm_job_runner, dataset):
```

examples/linkproppred/dygformer.py
Jacob-Chmura, line 1: Can we add a tgb eval example and get the numbers?

ntgbaoo: Hi @Jacob-Chmura,
I tried to add `TGB` example for `DyGFormer`. I am currently having some issues with time and space complexity. Here is my [evaluation function.](https://github.com/tgm-team/tgm/blob/6-DygformerImplementation/examples/linkproppred/TGB/dygformer.py#L214C1-L255C23)

Unlike other models, which take in the edge lists and return node embeddings of all nodes. DyGFormer returns `src` and `dst` node embeddings given an edge between them, and neighbours' information about two nodes. Hence, `DyGFormer` need to use both `encoder` and `decoder` for every negative batch in `batch.neg_batch_list` for each `src` node, which makes it run quite slow.

I also tried to concat everything in `batch.neg_batch_list` into a single batch and perform forward pass once (this is actually how the original code does), but this results in memory issues. I think we can reduce the batch size to deal with this if we want to follow in this way.

Jacob-Chmura: Without seeing the details, I agree that we should do forward pass once. Do you have a sense of peak gpu required on batch size 200?

ntgbaoo: I don't have an exact number. But I was running on a 48GiB GPU and based on the error message I got, I would say the peak is somewhere from 50GiB to 65GiB. It should be fine if we run on an 80GiB GPU.

ntgbaoo: However, I noticed that [this function, `_count_nodes_freq()`](https://github.com/tgm-team/tgm/blob/cb2781131e397bdd452b87724185ad9597022b2e/tgm/nn/model/dygformer.py#L36C1-L113C1) is now again the issue when we oncat everything in `batch.neg_batch_list`. I think we will eventually need to vectorize this function. `DyGMamba` is another model to be added also uses this function. Will benefit both if we successfully vectorize `_count_nodes_freq()`.

tgm/nn/model/dygformer.py
ntgbaoo: A note is added to the forward function to state the assumption clearly.

examples/linkproppred/dygformer.py
ntgbaoo: I see. Thank you for pointing this out.

test/integration/test_dygformer.py
```diff
@@ -12,7 +12,7 @@
         '--gres=gpu:a100l:1',
     ]
 )
-def test_tgat_recency_sampler_linkprop_pred(slurm_job_runner, dataset):
+def test_dygformer_link_prediction(slurm_job_runner, dataset):
```
Jacob-Chmura: Let's add another integration test that calls the _tgb_ variant

ntgbaoo: I added to the integration test

benjaminnNgo:
Thanks for your reviews, @Jacob-Chmura . I will hold on this PR. Let me know when I should pull a fix from https://github.com/tgm-team/tgm/pull/110.

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+from tgm.loader import DGDataLoader
+from tgm.nn import DyGFormer, Time2Vec
+from tgm.util.seed import seed_everything
+
+parser = argparse.ArgumentParser(
+    description='DyGFormers TGB Example',
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
+parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
+parser.add_argument('--dataset', type=str, default='tgbn-genre', help='Dataset name')
```
Jacob-Chmura: We want `tgbn-genre` as default? Not `tgbl-wiki`?

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+parser.add_argument(
+    '--num_heads', type=int, default=2, help='number of heads used in attention layer'
+)
+parser.add_argument(
+    '--num-channels',
+    type=int,
+    default=4,
+    help='number of channels used in attention layer',
+)
+
```
Jacob-Chmura:
```suggestion
```

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+parser.add_argument('--bsize', type=int, default=200, help='batch size')
+
+
+class LinkPredictor(nn.Module):
```
Jacob-Chmura, line 64: Can we replace this with our new link predictor: 

https://github.com/tgm-team/tgm/blob/5b135d79f0f8c742d31abdaceeb9e1814a5c9ed5/examples/linkproppred/TGB/gcn.py#L89-L98


Dyglib uses the `concat` version as well, so should be safe.

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+            num_heads,
+            dropout,
+            max_input_sequence_length,
+            num_channels,
+            time_encoder,
+            device,
+        )
+        self.decoder = LinkPredictor(output_dim)
+
+    def forward(self, batch: DGBatch) -> Tuple[torch.Tensor, torch.Tensor]:
```
Jacob-Chmura, line 113: We can keep as is, though this could probably be cleaned up, maybe done in a single encoder pass. If we get to that point, then we should refactor this example to have explicit encoder/decoder as in our other revamped examples.

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+            nbr_feats[: pos_batch_size * 2],
+        )
+        pos_out = self.decoder(z_src_pos, z_dst_pos)
+
+        neg_nbr_nids = nbr_nids[-neg_batch_size:]
+        neg_nbr_times = nbr_times[-neg_batch_size:]
+        neg_nbr_feats = nbr_feats[-neg_batch_size:]
+        src_nbr_nids = nbr_nids[:pos_batch_size]
+        src_nbr_times = nbr_times[:pos_batch_size]
+        src_nbr_feats = nbr_feats[:pos_batch_size]
```
Jacob-Chmura: To be safe, I would do `nbr-nids[pos_batch_size * 2 : ]` because with tgb, `batch.neg` is unique negatives. I think its equivalent since tgb neg sampler does not return duplicates, but important to keep in mind. Since its working, we can leave this for refactor work later.

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+    return total_loss
+
+
+@torch.no_grad()
+def eval(
+    evaluator: Evaluator,
+    loader: DGDataLoader,
+    model: nn.Module,
+    eval_metric: str,
+) -> dict:
```
Jacob-Chmura: Nit: Can we conform to the other example argument order where applicable:
https://github.com/tgm-team/tgm/blob/5b135d79f0f8c742d31abdaceeb9e1814a5c9ed5/examples/linkproppred/TGB/gcn.py#L141-L152

Also, let's just return mrr as float, since we don't care about any other metric:

https://github.com/tgm-team/tgm/blob/5b135d79f0f8c742d31abdaceeb9e1814a5c9ed5/examples/linkproppred/TGB/gcn.py#L183

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+            ).long()
+            copy_batch.nbr_nids = [batch.nbr_nids[0][all_idx]]
+            copy_batch.nbr_times = [batch.nbr_times[0][all_idx]]
+            copy_batch.nbr_feats = [batch.nbr_feats[0][all_idx]]
+
+            pos_out, neg_out = model(copy_batch)
+
+            input_dict = {
+                'y_pred_pos': pos_out.detach().cpu().numpy(),
+                'y_pred_neg': neg_out.detach().cpu().numpy(),
```
Jacob-Chmura: Also, it turns out tgb works with tensors 
```suggestion
                'y_pred_pos': pos_out,
                'y_pred_neg': neg_out,
```

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+                    torch.Tensor([idx + batch.src.shape[0]]).to(
+                        neg_batch.device
+                    ),  # dst idx
+                    neg_idx,
+                ],
+                dim=0,
+            ).long()
+            copy_batch.nbr_nids = [batch.nbr_nids[0][all_idx]]
+            copy_batch.nbr_times = [batch.nbr_times[0][all_idx]]
+            copy_batch.nbr_feats = [batch.nbr_feats[0][all_idx]]
```
Jacob-Chmura: Can you clarify what this is doing? Also, I think this makes sense to put internally into the model rather than the training loop?

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+                'eval_metric': [eval_metric],
+            }
+            perf_list.append(evaluator.eval(input_dict)[eval_metric])
+
+    metric_dict = {}
+    metric_dict[eval_metric] = float(np.mean(perf_list))
+    return metric_dict
+
+
+if __name__ == '__main__':
```
Jacob-Chmura:
```suggestion
```

Nit, for consistancy

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+        STATIC_NODE_FEAT = train_dg.static_node_feats
+    else:
+        STATIC_NODE_FEAT = torch.randn(
+            (test_dg.num_nodes, args.node_dim), device=args.device
+        )
+
+    _, dst, _ = train_dg.edges
+    train_neg_hook = NegativeEdgeSamplerHook(low=int(dst.min()), high=int(dst.max()))
+    val_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
+    test_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')
```
Jacob-Chmura: Can inline the registry as per the revamped examples:

https://github.com/tgm-team/tgm/blob/5b135d79f0f8c742d31abdaceeb9e1814a5c9ed5/examples/linkproppred/TGB/gcn.py#L212-L217

examples/linkproppred/TGB/dygformer.py
```diff
@@ -0,0 +1,330 @@
...
+            start_time = time.perf_counter()
+            loss = train(train_loader, model, opt)
+            end_time = time.perf_counter()
+            latency = end_time - start_time
+        with hm.activate('val'):
+            val_results = eval(evaluator, val_loader, model, eval_metric)
+            print(
+                f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
+                + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
+            )
```
Jacob-Chmura: After switching `eval` to return float, we can use the following print: 
https://github.com/tgm-team/tgm/blob/5b135d79f0f8c742d31abdaceeb9e1814a5c9ed5/examples/linkproppred/TGB/gcn.py#L276-L278

examples/linkproppred/dygformer.py
Jacob-Chmura, line 1: We can drop this file as we move to MRR-only link pred examples: #196 , or you can keep it here and I'll remove it in the subsequent pr

test/integration/test_dygformer.py
```diff
@@ -0,0 +1,53 @@
...
+
+python "$ROOT_DIR/examples/linkproppred/TGB/dygformer.py" \
+    --dataset {dataset} \
+    --device cuda \
+    --epochs 1 \
+    --sampling recency \
+    --n-nbrs 20 20
+    """
+    state = slurm_job_runner(cmd)
+    assert state == 'COMPLETED'
```
Jacob-Chmura: Can drop this
```suggestion
```

test/integration/test_dygformer.py
```diff
@@ -0,0 +1,53 @@
...
+        '--mem=8G',
+        '--time=3:00:00',
+        '--gres=gpu:a100l:1',
+    ]
+)
+def test_dygformer_link_prediction(slurm_job_runner, dataset):
+    cmd = f"""
+echo "Downloading dataset: {dataset}"
+echo "y" | python -c "from tgb.linkproppred.dataset import LinkPropPredDataset; LinkPropPredDataset('{dataset}')"
+
```
Jacob-Chmura: After the new CI changes, we no longer need to do this for each integration test :)
```suggestion
```

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,585 @@
...
+from typing import Callable, Tuple
+
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from ..time_encoding import Time2Vec
+
+PADDED_NODE_ID = -1
```
Jacob-Chmura: We have it now in our constants module :)

```suggestion
from tgm.constants import PADDED_NODE_ID
```

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,585 @@
...
+                torch.from_numpy(src_counts[src_inverse_indices])
+                .float()
+                .to(self.device)
+            )
+            src_mapping_dict = dict(zip(src_unique_keys, src_counts))
+
+            dst_unique_keys, dst_inverse_indices, dst_counts = np.unique(
+                dst_neighbors, return_inverse=True, return_counts=True
+            )
+            # Fequency of each destination's neighbor within destination's neighbors
```
Jacob-Chmura: We can fix these spelling mistakes to avoid getting flagged in later prs, thanks
```suggestion
            # Frequency of each destination's neighbor within destination's neighbors
```

tgm/nn/model/dygformer.py
Jacob-Chmura, line 1: Great work, I won't dive into details of this file since we know it works. Thanks for marking the bottleneck clearly for future optimization.

examples/linkproppred/TGB/dygformer.py
ntgbaoo: Agreed. Will take a mark as #@TODO in the code and come back to it

test/integration/test_dygformer.py
ntgbaoo: Nice

test/integration/test_dygformer.py
ntgbaoo: @Jacob-Chmura, we should keep this TGB example for integration test, right? We should remove non-tgb ones

Jacob-Chmura: Yep, accidently marked the wrong one

examples/linkproppred/TGB/dygformer.py
ntgbaoo: `all_idx ` will contain the index of `src`, `dst` and negative nodes to retrieve neighbors from `batch.nbr_nids[0]`.. I agree that this code is kinda confusing, but I don't have a better way to make it look nicer. 
I don't think we can move it into the model itself, as this is done in inference with TGB only. We don't need to do this in training

test/integration/test_dygformer.py
```diff
@@ -0,0 +1,25 @@
...
+    ]
+)
+def test_dygformer_tgb_recency_sampler_linkprop_pred(slurm_job_runner, dataset):
+    cmd = f"""
+python "$ROOT_DIR/examples/linkproppred/TGB/dygformer.py" \
+    --dataset {dataset} \
+    --device cuda \
+    --epochs 1 \
+    --sampling recency \
+    --n-nbrs 20 20
```
Jacob-Chmura: Missed this
```suggestion
```

tgm/nn/model/dygformer.py
```diff
@@ -0,0 +1,585 @@
...
+        )
+
+        source_freq, dst_freq = [], []
+        for src_neighbors, dst_neighbors in zip(
+            all_sources_neighbors, all_dsts_neighbors
+        ):
+            src_unique_keys, src_inverse_indices, src_counts = np.unique(
+                src_neighbors, return_inverse=True, return_counts=True
+            )
+            # Fequency of each source's neighbor within source's neighbors
```
Jacob-Chmura:
```suggestion
            # Frequency of each source's neighbor within source's neighbors
```

Jacob-Chmura (approved):
Just a few more minor things I forgot to mark in my previous review

PR #124, opened 2025-08-02
adding first basic test for neighbor samplers

shenyangHuang:
- [x]  adding the following basic test 

```
 #############                    ###########
    # Alice (0) # ->    t = 1     -> # Bob (1) #
 #############                    ###########
         |
         v
       t = 2
         |
         v
#############                    ############
# Carol (2) # ->   t = 3      -> # Dave (3) #
#############                    ############
         |
         v
       t = 4
         |
         v
  ############# 
    # Alice (0) #  
 #############                   

```

- [x]  add a recency buffer test

Testing if new neighbors are registered correctly into the buffer while old neighbors are pushed out

- [x]  add a two hop test 

Test if two hop neighbors are correctly sampled and past neighbors are pushed out of recency sampler correctly

## Findings

1. `uniform` sampler samples from current batch, need to be changed
2. `recency` sampler when going over batches of 2 edges (with same source node), only updates 1 of the destination instead of both

Test to add from GNNflow

https://github.com/jasperzhong/GNNFlow/blob/main/tests/test_temporal_sampler.py


test/unit/test_neighbor_sampler.py
```diff
@@ -0,0 +1,363 @@
...
+        nbr_hook = NeighborSamplerHook(num_nbrs=n_nbrs)
+    elif sampling_type == 'recency':
+        nbr_hook = RecencyNeighborHook(
+            num_nbrs=n_nbrs,
+            num_nodes=dg.num_nodes,
+            edge_feats_dim=dg.edge_feats_dim,
+        )
+    else:
+        raise ValueError(f'Unknown sampling type: {sampling_type}')
+    return [nbr_hook]
```
Jacob-Chmura: Can we directly initialize the relevant sampler in the correspending function instead of this.

test/unit/test_neighbor_sampler.py
```diff
@@ -0,0 +1,363 @@
...
+    assert hasattr(batch, 'nbr_times')
+    assert hasattr(batch, 'nbr_feats')
+    assert hasattr(batch, 'nbr_mask')
+
+    nids = np.array(batch.nids)
+    nbr_nids = np.array(batch.nbr_nids)
+    nbr_times = np.array(batch.nbr_times)
+    nbr_feats = np.array(batch.nbr_feats)
+    nbr_mask = np.array(batch.nbr_mask)
+    return [nids, nbr_nids, nbr_times, nbr_feats, nbr_mask]
```
Jacob-Chmura: Can we keep them as torch tensors and use `torch.testing.assert_equal` as I have throughout our test suite

test/unit/test_neighbor_sampler.py
```diff
@@ -0,0 +1,363 @@
...
+
+
+def _init_basic_sample_graph():
```
Jacob-Chmura: This should be a pytest fixture that we can then easily re-use across tests.

test/unit/test_neighbor_sampler.py
```diff
@@ -0,0 +1,363 @@
...
+    assert nbr_nids[0][1][0] == 2
+    assert nbr_times.shape == (1, 2, 1)
+    assert nbr_times[0][0][0] == 3
+    assert nbr_times[0][1][0] == 2
+    assert nbr_feats.shape == (1, 2, 1, 1)  # 1 feature per edge
+    assert nbr_feats[0][0][0][0] == 5.0
+    assert nbr_feats[0][1][0][0] == 2.0
+    assert nbr_mask.shape == (1, 2, 1)
+
+    # starting test for uniform sampler
```
Jacob-Chmura: Each test should test an isolated unit. Let's break up the uniform and recency sampler tests. Also, they should go in the appropriate test files that I already have created, and which already have some basic api testing done.

Jacob-Chmura: We can use pytest to capture the output (from the working recency sampler) and use that to check against both uniform and recency (since we know both should match on this graph))

test/unit/test_neighbor_sampler.py
```diff
@@ -0,0 +1,363 @@
...
+    assert _batch_eq_nbrs(batch_2, batch)
+
+    batch = next(iter(loader))
+    assert _batch_eq_nbrs(batch_3, batch)
+
+    batch = next(iter(loader))
+    assert _batch_eq_nbrs(batch_4, batch)
+
+
+def _init_recency_buffer_graph():
```
Jacob-Chmura: Please make a pytest fixture as above

test/unit/test_neighbor_sampler.py
Jacob-Chmura, line 1: After addressing the comments here, can we add test cases that include negative edges. TGB preferred. You can mock the TGB negative edge sampler to give you obvious negatives. This should ensure we correctly manage the _skewed_ `src,dst,neg` neighbors in our samplers.

test/unit/test_neighbor_sampler.py
shenyangHuang: addressed

test/unit/test_neighbor_sampler.py
shenyangHuang: this is what I have tested today yes, the assert now is correct

test/unit/test_neighbor_sampler.py
shenyangHuang: all comments addressed. Good point, but I would wait until later to test this especially we might still introduce changes to TGBNegative sample, i.e. timestamp of negatives.

tgm/hooks.py
Jacob-Chmura, line 1: I think these are accidental, can we revert this file?

Issue #122, opened 2025-07-24
Better description for DGBatch


benjaminnNgo:
If a user wants to see how DGBatch looks like, what attributes it has. Currently, print(DGBatch):

<img width="1537" height="231" alt="Image" src="https://github.com/user-attachments/assets/d6520dde-4c10-4173-956f-f5e88027dd0d" />

This output doesn't include the attributes added by hooks at run time

A __str__() needs to be added to have a better description. Ideally, the format is like this:

<img width="1095" height="32" alt="Image" src="https://github.com/user-attachments/assets/fbac3921-cf07-4113-86ec-6e8bdd2e3513" />

and should include attributes added by hooks at run time

PR #125, opened 2025-08-02
Added __str__ for better description for DGBatch

benjaminnNgo:
## Objective

Added def `def __str__()` to DGBatch for a better description of all attributes of DGBatch added by Hook(s).

## Output

```
DGBatch(src = [167],dst = [167],time = [167],dynamic_node_feats = None,edge_feats = [167, 172],node_times = None,node_ids = None,neg = [167],nids = list([501]),times = list([501]),nbr_nids = list([501, 500]),nbr_times = list([501, 500]),nbr_feats = list([501, 500, 172]),nbr_mask = list([501, 500]),unique_nids = [254],global_to_local = function)
```

**Related PR**: https://github.com/tgm-team/tgm/issues/122

### DIFF
--- a/tgm/graph.py
+++ b/tgm/graph.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import pathlib
-from collections.abc import Iterable
+from collections.abc import Iterable, Sized
 from dataclasses import dataclass
 from functools import cached_property
 from typing import Any, List, Literal, Optional, Set, Tuple
@@ -448,13 +448,20 @@ class DGBatch:
                 unique_type = set()
                 for element in object:
                     unique_type.add(_get_description(element))
-                description = type(object).__name__ + '(' + '|'.join(unique_type) + ')'
+                if isinstance(object, Sized):
+                    obj_len = f' x{str(len(object))}'
+                else:
+                    obj_len = ''
+
+                description = (
+                    type(object).__name__ + '(' + '|'.join(unique_type) + obj_len + ')'
+                )
             else:
                 description = type(object).__name__
-                
+
             return description
 
         descriptions = []
         for attr, value in vars(self).items():
             descriptions.append(f'{attr} = {_get_description(value)}')
-        return 'DGBatch(' + ','.join(descriptions) + ')'
+        return 'DGBatch(' + ', '.join(descriptions) + ')'


tgm/graph.py
```diff
@@ -437,3 +438,24 @@ class DGBatch:
...
+
+    def __str__(self) -> str:
+        def _get_description(object: Any) -> str:
+            description = ''
+            if isinstance(object, torch.Tensor):
+                description = str(list(object.shape))
+            elif isinstance(object, Iterable):
+                unique_type = set()
+                for element in object:
+                    unique_type.add(_get_description(element))
```
Jacob-Chmura, line 450: So if we have a list, the string repr we provide will just show the unique elements?

tgm/graph.py
```diff
@@ -437,3 +438,24 @@ class DGBatch:
...
+                description = type(object).__name__ + '(' + '|'.join(unique_type) + ')'
+            elif object is not None:
```
Jacob-Chmura: I think this and the else case can be collapsed since type(None).__name__ will just give back NoneType?

tgm/graph.py
ntgbaoo: yes. I don't think we should print them. It will be too much. We can print out the length of the list as well.

Jacob-Chmura: Yeah, I agree that printing out all elements is excessive. I was just thinking if a user has a hook that expects a list (with duplicates) and we print out unique elements, they may think something is missing from the batch.

Jacob-Chmura (approved):

tgm/graph.py
ntgbaoo: I updated the description as follows:
```
DGBatch(src = [167], dst = [167], time = [167], dynamic_node_feats = NoneType, edge_feats = [167, 172], node_times = NoneType, node_ids = NoneType, neg = [167], nids = list([0]|[501] x2), times = list([0]|[501] x2), nbr_nids = list([0, 600]|[501, 500] x2), nbr_times = list([0, 600]|[501, 500] x2), nbr_feats = list([0, 600, 172]|[501, 500, 172] x2), nbr_mask = list([0, 600]|[501, 500] x2), unique_nids = [254], global_to_local = function) 
```

`|` represents `or`. Do you think this description is easy to understand, @Jacob-Chmura?

Jacob-Chmura: To be honest, it is a bit verbose. But for the time being we can work with it, as it could help us debug. But for end-user, this may not be super useful.

PR #130, opened 2025-08-08
Negative Time in Negative Hooks

### LINKED ISSUE #128: Move Negative Time Assignment to Negative Sampler
Idea being we can use the ground truth edge times in our negatives (for training), thereby making the training signal stronger, potentially speeding up traijing.

Jacob-Chmura:
Close #128

### DIFF
--- a/tgm/hooks.py
+++ b/tgm/hooks.py
@@ -139,7 +139,7 @@ class DeduplicationHook:
 
 class NegativeEdgeSamplerHook:
     requires: Set[str] = set()
-    produces = {'neg'}
+    produces = {'neg', 'neg_time'}
 
     r"""Sample negative edges for dynamic link prediction.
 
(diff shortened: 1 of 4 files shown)


tgm/hooks.py
```diff
@@ -352,22 +354,8 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
                 if hasattr(batch, 'neg'):
                     batch.neg = batch.neg.to(device)
                     seed.append(batch.neg)
+                    times.append(batch.neg_time)  # type: ignore
```
Jacob-Chmura, line 357: It's fine for now, but we would probably benefit eventually from grouping hooks into categories (e.g. negative sampling, neighbour sampling). We implicitly assume certain attributes are added together (e.g. `batch.neg, batch.neg_times`). The right abstraction here could enable the hook manager to check whether batch has negatives, and if it does, to ensure _all_ the required negative-related attributed are there. Similarly, for neighbours, where we assume nbr ids, nbr times and nbr mask are all added.

tgm/hooks.py
```diff
@@ -210,6 +211,22 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+
+        # This is a heuristic. For our fake (negative) link times,
+        # we pick random time stamps within [batch.start_time, batch.end_time].
+        # Using random times on the whole graph will likely produce information
+        # leakage, making the prediction easier than it should be.
+
+        # Use generator to local constrain rng for reproducibility
+        gen = torch.Generator(device=dg.device)
+        gen.manual_seed(0)
+        batch.neg_time = torch.randint(  # type: ignore
```
shenyangHuang, line 223: flag this as an issue to come back to later, I think this is a good way of dealing with this but also the idea yesterday on 200x the graph during message passing also makes sense. Either way, this ties into the inference part.

tgm/hooks.py
shenyangHuang: yes, definitely, as we are adding more hooks, the core structure will become more clear as well.

shenyangHuang (approved):
Looks good, thanks Jacob

PR #134, opened 2025-08-12
Clean Up Known Dataset Construction API

### LINKED ISSUE #117: Clean up direct DGData and  Dgraph from known datasets

Jacob-Chmura:
Close #117

### DIFF
--- a/tgm/data.py
+++ b/tgm/data.py
@@ -502,20 +502,9 @@ class DGData:
         )
 
     @classmethod
-    def from_any(
-        cls,
-        data: str | pathlib.Path | 'pandas.DataFrame',  # type: ignore
-        time_delta: TimeDeltaDG | None = None,
-        **kwargs: Any,
+    def from_known_dataset(
+        cls, data: str, time_delta: TimeDeltaDG | None = None, **kwargs: Any
     ) -> DGData:
-        if isinstance(data, (str, pathlib.Path)):
-            data_str = str(data)
-            if data_str.startswith('tgbl-') or data_str.startswith('tgbn-'):
-                return cls.from_tgb(name=data_str, time_delta=time_delta, **kwargs)
-            if data_str.endswith('.csv'):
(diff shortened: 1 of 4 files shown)


tgm/data.py
```diff
@@ -502,20 +502,9 @@ def _check_tgb_import() -> tuple['LinkPropPredDataset', 'NodePropPredDataset']:
         )
 
     @classmethod
-    def from_any(
-        cls,
-        data: str | pathlib.Path | 'pandas.DataFrame',  # type: ignore
-        time_delta: TimeDeltaDG | None = None,
-        **kwargs: Any,
+    def from_known_dataset(
```
Jacob-Chmura, line 505: And here we will dispatch to ogb, etc.

tgm/graph.py
```diff
@@ -19,7 +19,7 @@ class DGraph:
 
     def __init__(
         self,
-        data: DGStorage | DGData | str | pathlib.Path | 'pd.DataFrame',  # type: ignore
+        data: DGStorage | DGData | str,
```
Jacob-Chmura, line 22: Simplify the constructor. User either created DGData, or uses an known dataset (e.g. 'tgb-xxx'). If you want more flexibility, use class constructors (e.g. from_pandas)

tgm/graph.py
```diff
@@ -203,16 +194,8 @@ def from_tgb(
...
-            name=name,
-            split=split,
-            time_delta=time_delta,  # type: ignore
-        )
-
-        return cls(
-            data=data,
-            device=device,
-        )
+        data = DGData.from_tgb(name=name, split=split, time_delta=time_delta)  # type: ignore
```
shenyangHuang, line 197: would this be a little bit redundant? let's keep for now, we can also make it private such that only constructor would use it

Jacob-Chmura: Yeah, it is redundant, keeping for now

PR #135, opened 2025-08-12
WIP: Discretize API POC

### LINKED ISSUE #91: Enable Discretization Directly from Constuctor

Jacob-Chmura:
Close #91

### DIFF
--- a/tgm/data.py
+++ b/tgm/data.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import csv
 import pathlib
 from dataclasses import dataclass
-from typing import Any, List
+from typing import Any, List, Literal, Set, Tuple
 
 import numpy as np
 import torch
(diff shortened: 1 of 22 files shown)


tgm/data.py
```diff
@@ -477,20 +558,6 @@ def _check_tgb_import() -> tuple['LinkPropPredDataset', 'NodePropPredDataset']:
         if dataset.node_feat is not None:
             static_node_feats = torch.from_numpy(dataset.node_feat)
 
-        # Remap timestamps if a custom non-ordered time delta was supplied
```
Jacob-Chmura, line 480: Double check if we want to keep this, drop it, or move it over to `discretize` in situations where we map into a _more_ fine grain time delta. E.g. (mapping from minutes to seconds just multiples timestamps by 60).

tgm/data.py
```diff
@@ -170,6 +170,93 @@ def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
...
+        discretize_time_delta: TimeDeltaDG | str | None,
+        reduce_op: str | None,
+    ) -> DGData:
+        if isinstance(native_time_delta, str):
+            native_time_delta = TimeDeltaDG(native_time_delta)
+        if isinstance(discretize_time_delta, str):
+            discretize_time_delta = TimeDeltaDG(discretize_time_delta)
+
+        if discretize_time_delta is None or native_time_delta == discretize_time_delta:
+            return self  # Note: Does not do a copy
```
Jacob-Chmura, line 185: Safe to do since users do not call thsi directly

tgm/data.py
```diff
@@ -502,20 +569,18 @@ def _check_tgb_import() -> tuple['LinkPropPredDataset', 'NodePropPredDataset']:
...
+    ) -> Tuple[DGData, TimeDeltaDG]:
+        if data_name.startswith('tgbl-') or data_name.startswith('tgbn-'):
+            data = cls.from_tgb(data_name, **kwargs)
+            native_time_delta = TGB_TIME_DELTAS[data_name]
         else:
-            return cls.from_pandas(data, **kwargs)
+            raise ValueError(f'Unsupported dataset identifier: {data_name}')
+
+        data = data.discretize(native_time_delta, discretize_time_delta, reduce_op)
+        return data, native_time_delta
```
Jacob-Chmura: TODO: We should be returning `discretize_time_delta` if it was specified, otherwise `native_time_delta`. The will become the actual time delta on the graph.

All known datasets will be forced to provide a native time delta associated with the graph

tgm/graph.py
```diff
@@ -19,23 +19,31 @@ class DGraph:
 
     def __init__(
         self,
-        data: DGStorage | DGData | str | pathlib.Path | 'pd.DataFrame',  # type: ignore
-        time_delta: TimeDeltaDG | str = 'r',
+        data: DGStorage | str,
```
Jacob-Chmura: No longer support creating graph from DGData. Up for debate I suppose. If we want to keep it hidden from user, the idea here is that user can only directly construct DGraph from a known dataset (with string like 'tgb-wiki'). If they have custom datasets, they should use `from_csv` or `from_raw` or `from_pandas`.

The `DGDStorage` option only happens internally, when the library createa a new view on an existing storage. 

This somewhat obfuscates the notion of `discretize_time_delta`, but the logic is rather simple:
- If data is a storage, then the core library is calling the consturctor. In this case, we expect to forward the time delta from the existing graph, so it must be specified.
- If data is a string, then the `discretize_time_delta` is optional. If user specifies it, then they are attempting to discretize a known dataset. If they ignore it, we use native time granularity of the data.

tgm/graph.py
```diff
@@ -70,8 +80,16 @@ def from_csv(
...
+            native_time_delta (TimeDeltaDG | str): Native Time Delta for the graph. Defaults to 'r' for ordered edgelist.
+            discretize_time_delta (TimeDeltaDG | str | None): If specified, determines the time to discretize the underlying data to.
+            reduce_op (str | None): If discretize_time_delta is specified, this determines what reduction to apply to the underlying data.
             device (str | torch.device): Device to store the graph on. Defaults to 'cpu'.
+
+        Note:
+            - If `discretize_time_delta` is specified:
+                1. `native_time_delta` must be an absolute time delta, **not** `'r'`.
+                2. The native time delta must be **finer** than `discretize_time_delta`.
+                3. A valid `reduce_op` is required.
```
Jacob-Chmura, line 94: TODO: Copy this to other docstrings

tgm/graph.py
```diff
@@ -180,88 +202,36 @@ def from_pandas(
...
+            data_name (str): Name of the TGB dataset.
             split (str): Split of the dataset to use. Defaults to 'all'.
-            time_delta (TimeDeltaDG | None): Time delta for the graph. If None, uses the default for the dataset.
+            discretize_time_delta (TimeDeltaDG | str | None): If specified, determines the time to discretize the underlying data to.
+            reduce_op (str | None): If discretize_time_delta is specified, this determines what reduction to apply to the underlying data.
             device (str | torch.device): Device to store the graph on. Defaults to 'cpu'.
         """
-        data = DGData.from_tgb(
-            name=name,
+        data, time_delta = DGData.from_known_dataset(
```
Jacob-Chmura, line 261: For API consistency, it may make more sense to leave the discrtization to the caller. I.e. we discretize here instead if under the hood (same as pandas/csv). Not a huge deal since the user never needs to call it though.

PR #138, opened 2025-08-13
TGAT Debug Multi-Hop

Jacob-Chmura:


examples/linkproppred/TGB/tgat.py
```diff
@@ -79,57 +80,55 @@ def __init__(
...
+        for j in range(1, self.num_layers + 1):
+            for i in range(self.num_layers - j + 1):
+                num_nodes = z[j - 1][i].size(0)
+                num_nbr = batch.nbr_nids[j - 1].shape[-1]
+                out = self.attn[j - 1](
+                    node_feat=z[j - 1][i],
+                    time_feat=self.time_encoder(torch.zeros(num_nodes, device=device)),
+                    nbr_node_feat=z[j - 1][i + 1].reshape(num_nodes, num_nbr, -1),
+                    edge_feat=batch.nbr_feats[i],
+                    nbr_mask=batch.nbr_nids[i] == -1,
```
Jacob-Chmura: This -1 is the _sentinal_ node id. Will refactor this and merge into master seperately.

examples/linkproppred/TGB/tgat.py
```diff
@@ -157,16 +156,13 @@ def train(
...
-        z_dst = z[batch.global_to_local(batch.dst)]
-        z_neg = z[batch.global_to_local(batch.neg)]
+        z_src, z_dst, z_neg = torch.chunk(z, 3)
 
         pos_out = decoder(z_src, z_dst)
         neg_out = decoder(z_src, z_neg)
 
-        loss = F.binary_cross_entropy_with_logits(pos_out, torch.ones_like(pos_out))
-        loss += F.binary_cross_entropy_with_logits(neg_out, torch.zeros_like(neg_out))
+        loss = F.binary_cross_entropy(pos_out, torch.ones_like(pos_out))
```
Jacob-Chmura, line 162: Consider making our link decoders output logits again. It's slightly more numerically stable.

examples/linkproppred/TGB/tgat.py
```diff
@@ -185,15 +181,18 @@ def eval(
...
-
+        id_map = {nid.item(): i for i, nid in enumerate(batch.nids[0])}
         for idx, neg_batch in enumerate(batch.neg_batch_list):
             dst_ids = torch.cat([batch.dst[idx].unsqueeze(0), neg_batch])
             src_ids = batch.src[idx].repeat(len(dst_ids))
 
-            z_src = z[batch.global_to_local(src_ids)]
-            z_dst = z[batch.global_to_local(dst_ids)]
+            src_idx = torch.tensor([id_map[n.item()] for n in src_ids], device=z.device)
+            dst_idx = torch.tensor([id_map[n.item()] for n in dst_ids], device=z.device)
```
Jacob-Chmura, line 190: The problem was that the base case of our loop uses `batch.nids[0]` instead of `batch.unique_ids`. We cannot searchsorted `batch.nids[0]` since it is not sorted. We should consider whether we want to depricate or update the dedup api accordingly.

examples/linkproppred/TGB/tgat.py
```diff
@@ -227,9 +226,7 @@ def eval(
 if train_dg.static_node_feats is not None:
     static_node_feats = train_dg.static_node_feats
 else:
-    static_node_feats = torch.randn(
-        (test_dg.num_nodes, args.embed_dim), device=args.device
-    )
+    static_node_feats = torch.zeros((test_dg.num_nodes, 1), device=args.device)
```
Jacob-Chmura, line 226: Dyglib used 1-dim zero features. We can test zeros vs. gaussian.

examples/linkproppred/TGB/tgat.py
```diff
@@ -283,11 +281,16 @@ def _init_hooks(
...
-        batch_size=args.bsize,
+    shared_nbr = RecencyNeighborHook(
+        test_dg.num_nodes, args.n_nbrs, test_dg.edge_feats_dim
     )
+    _, dst, _ = test_dg.edges
+    neg_hook = NegativeEdgeSamplerHook(low=int(dst.min()), high=int(dst.max()))
+    foo_loader = DGDataLoader(train_dg, hook=[neg_hook, shared_nbr], batch_size=2000)
+    for _ in tqdm(foo_loader):
+        pass
+    neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
```
Jacob-Chmura: This is only here to "share" the recency sampler state with the train loader. This can also be in isolation updated in master before merging this pr.

tgm/nn/attention.py
Jacob-Chmura, line 1: More intuitive to let the masking of the special value (previously 0, now -1) outsdie of our model layers. That way user is exposed and aware of it.

tgm/hooks.py
Jacob-Chmura, line 1: This is the unlimited memory recency sampler, which also does update() at the beginning of hte batch (potential leakage).

examples/linkproppred/TGB/tgat.py
shenyangHuang: sounds good

examples/linkproppred/TGB/tgat.py
shenyangHuang: okay, thats why we are using the `id_map` for now

examples/linkproppred/TGB/tgat.py
shenyangHuang: yes, can be an input argument

shenyangHuang: this is more on the research side though, as example, this will suffice for now

examples/linkproppred/TGB/tgat.py
```diff
@@ -264,9 +261,10 @@ def _init_hooks(
 
 
 encoder = TGAT(
-    edge_dim=train_dg.edge_feats_dim or args.embed_dim,
+    node_dim=1,
```
shenyangHuang: we should not hard code `node_dim`
I would put `static_node_feats.shape[1]` for example

examples/linkproppred/TGB/tgat.py
shenyangHuang: agreed, let's have the shared hook PR merged before this one.

tgm/hooks.py
shenyangHuang: I would always move the update at the end of the batch, at least before we merge.

tgm/nn/attention.py
shenyangHuang: yes, once we decided, we can also show an example of this in the tutorial

tgm/hooks/hooks.py
```diff
@@ -228,7 +229,7 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
 
 
 class RecencyNeighborHook(StatefulHook):
-    requires: Set[str] = set()
+    requires = {'neg'}
```
Jacob-Chmura: TODO: Fix the hook manager to implicitly put nbrs after negatives.

tgm/hooks/hooks.py
```diff
@@ -270,10 +267,7 @@ def num_nbrs(self) -> List[int]:
         return self._num_nbrs
 
     def reset_state(self) -> None:
-        self._nbr_nids.fill_(-1)
-        self._nbr_times.zero_()
-        self._nbr_feats.zero_()
-        self._nbr_ptr.zero_()
+        self._history = defaultdict(lambda: deque(maxlen=self._max_nbrs))
```
Jacob-Chmura, line 270: This max capacity works fine, and can inform the circular buffer implementation.

examples/linkproppred/TGB/tgat.py
```diff
@@ -188,13 +181,15 @@ def eval(
...
+        id_map = {nid.item(): i for i, nid in enumerate(batch.nids[0])}
         for idx, neg_batch in enumerate(batch.neg_batch_list):
             dst_ids = torch.cat([batch.dst[idx].unsqueeze(0), neg_batch])
             src_ids = batch.src[idx].repeat(len(dst_ids))
 
-            z_src = z[batch.global_to_local(src_ids)]
-            z_dst = z[batch.global_to_local(dst_ids)]
+            src_idx = torch.tensor([id_map[n.item()] for n in src_ids], device=z.device)
+            dst_idx = torch.tensor([id_map[n.item()] for n in dst_ids], device=z.device)
+            z_src = z[src_idx]
```
shenyangHuang, line 191: great, single forward pass instead of one for pos and one for neg,

tgm/hooks/hooks.py
shenyangHuang: yep

tgm/hooks/hooks.py
```diff
@@ -254,14 +255,10 @@ def __init__(
         self._max_nbrs = max(num_nbrs)
 
         # We need edge_feats_dim to pre-allocate the right shape for self._nbr_feats
-        self._nbr_nids = torch.full(
-            (num_nodes, self._max_nbrs), PADDED_NODE_ID, dtype=torch.long
+        self._edge_feats_dim = edge_feats_dim
```
shenyangHuang, line 258: we can call this recency for now and replace it later.

PR #139, opened 2025-08-13
Dev/dgdataset

Jacob-Chmura:


tgm/dataset.py
```diff
@@ -0,0 +1,271 @@
...
+
+    @property
+    def val(self) -> Optional[T]:
+        return self._val
+
+    @property
+    def test(self) -> Optional[T]:
+        return self._test
+
+    def get_splits(self) -> Tuple[Optional[T], Optional[T], Optional[T]]:
```
shenyangHuang, line 34: what is the output, `DGraph`s?

tgm/dataset.py
```diff
@@ -0,0 +1,271 @@
...
+    def _load_splits(
+        self,
+    ) -> Tuple[Optional[DGData], Optional[DGData], Optional[DGData]]: ...
+
+    @property
+    @abstractmethod
+    def native_time_delta(self) -> TimeDeltaDG: ...
+
+
+class TGBDataset(PreSplitDGDataset):
```
shenyangHuang, line 140: so we don't need `from_tgb` any more?

PR #140, opened 2025-08-14
Simplify Discretize API

Jacob-Chmura:
# Purpose
Simplify the discretization semantics on DGraph construction. **Note: we do not support discretization a known graph (TGB) implicitly from constructor.**

## Key Changes
- minor formatting and simplification in examples
- `DGData.from_known_datasets` now returns the _native_ time granularity of the dataset.
- If user tries to specify a `time_delta` while loading a known dataset, we raise `UserWarning` to tell them that we are ignoring it. We tell them to instead call `.discretize` with their requested time delta after graph creation.
- Drop `DGraph.from_TGB` class method

## Out of Scope
Direct discretization from construction. This is still to be settled. Now, at least, the semantics are more consistent across known datasets and TGB:
- When user creates the graph, it will _always_ use a _native time delta_. In the case of TGB, it is pre-defined. In the case of custom dataset (e.g. from_csv), the default is 'r'.
- **Then**, user can discretize this native time delta. In both cases, this will be an explicit method call.

## Problems remaining
- Discritization is fundmentally a `DGData` op. We can discretize if and only if we have this information, otherwise, we essentially need to rebuild the data. Currently, we get away with it because `DGStorageArrayBackend` happens to keep a reference to the underlying DGData object. Hence, the discretization is straightforward. But imagine we have TCSR. Then, discretizing it is not as straightforward.
- From an efficiency perspective, this is quite poor, in the sense that, if a user knows they want to discretize, processing the full dataset only to then rebuild a subset of it is not great. This is particularly true for backends where the construction is not O(1).

## Proposed path forward
I suggest we merge this, as it is a straightforward API improvement that makes our current intentions more clear. Next, I suggest that we think about `DGDataset` #139 and it's implication on the current design. After this, we can see and decide:
- If we want to support _implicit discretization from constructor_: Then `discretize()` has to move out of backend and into `DGData`. We may need to change argument names (e.g. `time_delta -> discretize_time_delta`, add `reduce_op`) so that the user understands what is happening under the hood. This is what is currently implemented in #135 but, as you can see it's a big change, and not clear the benefits. Hence why I am breaking things up.
- If we do not want to support _implicit discretization from constructor_: Then we will be good to go. Once perforamance test are in #64 then we can measure the impact of discretization after construction. Off the top of my head, it seems reasonable to assume that the cost would be ~2x peak memory (because we have the original and the discretized version of `DGData` in memory at once, and roughly ~2-3x latency (we have to reconstruct a new storage, and we have to loop through the graph once to actually discretize the DGData)


test/unit/test_data.py
```diff
@@ -769,50 +777,8 @@ def _get_exp_times():
     mock_dataset_cls.assert_called_once_with(name='tgbn-trade')
 
 
-@patch('tgb.linkproppred.dataset.LinkPropPredDataset')
```
Jacob-Chmura, line 772: No longer implicitly discretize from DGraph constuctor. Instead, user has to manually discretize if loading a _known dataset_.

tgm/graph.py
```diff
@@ -20,25 +20,33 @@ class DGraph:
...
+                    'Call .discretize() after creating the DGraph to change it.'
+                )
+            data, time_delta = DGData.from_known_dataset(data, **kwargs)
+            data = DGStorage(data)
         elif isinstance(data, DGData):
-            self._storage = DGStorage(data)
-        elif isinstance(data, str):
-            data = DGData.from_known_dataset(data, time_delta, **kwargs)
-            self._storage = DGStorage(data)
+            time_delta = time_delta or TimeDeltaDG('r')  # Default to ordered
```
Jacob-Chmura, line 38: This is mostly used in our unit tests. But it remains a valid default for custom user (not pre-defined) datasets.

test/unit/test_data.py
```diff
@@ -690,8 +693,9 @@ def _get_exp_times():
         mask = getattr(dataset, f'{split}_mask')
         return times[mask]
 
-    data = DGData.from_tgb(name='tgbl-wiki', split=split)
+    data, native_time_delta = DGData.from_tgb(name='tgbl-wiki', split=split)
```
shenyangHuang, line 696: why does `from_tgb` return a native_time_delta

examples/nodeproppred/persistant_forecast.py
```diff
@@ -80,9 +80,9 @@ def eval(loader: DGDataLoader, model: PersistantForecaster) -> dict:
 args = parser.parse_args()
 seed_everything(args.seed)
 
-train_dg = DGraph(args.dataset, time_delta=args.time_gran, split='train')
-val_dg = DGraph(args.dataset, time_delta=args.time_gran, split='val')
-test_dg = DGraph(args.dataset, time_delta=args.time_gran, split='test')
+train_dg = DGraph(args.dataset, split='train').discretize(args.time_gran)
```
shenyangHuang, line 83: to discuss

shenyangHuang: maybe we just iterate by batch time grain instead of performing discretization

tgm/graph.py
shenyangHuang: yes for sure, could check a more intuitive name later

test/unit/test_data.py
Jacob-Chmura: This is an artifact from our previous API. We returned the native time delta from this method so that `from_known_dataset` could decide whether or not to discretize the graph (based on user provided `time_delta`). 

This will not be in the final design.

examples/nodeproppred/persistant_forecast.py
Jacob-Chmura: Yes. We should only ever discretize when we want to explicitly reduce the granularity of the graph. If we just want to change how we iterate, the `DGDataLoader` API handles this entirely.

shenyangHuang (approved):

PR #144, opened 2025-08-14
Prevent TGB Direct DGraph Construction

### LINKED ISSUE #143: Prevent Explicit DGraph TGB Construction

Jacob-Chmura:
Close #143


examples/nodeproppred/tgcn.py
```diff
@@ -160,41 +159,22 @@ def eval(
...
-test_dg = DGraph(
-    args.dataset,
-    time_delta=TimeDeltaDG(args.time_gran),
-    split='test',
-    device=args.device,
-)
+train_data, time_delta = DGData.from_tgb(args.dataset, split='train')
+train_dg = DGraph(train_data, time_delta, device=args.device).discretize(args.time_gran)
+
+val_data, time_delta = DGData.from_tgb(args.dataset, split='val')
```
shenyangHuang, line 165: we need to update tutorial with this as well

shenyangHuang (changes requested):
maybe quickly change the tutorial syntax as well

examples/nodeproppred/tgcn.py
Jacob-Chmura: I intentionally did not update the docs for 2 reasons.

Reason 1: Our public docs host the latest documentation. These should only be updated when a new build is cut. Otherwise, the user might `pip install py-tgm` and find inconsistenancies with the version we documented and the version we released. Eventually, when we are in the routine of doing proper releases, we can have seperate docs for nightly vs stabel builds.

Reason 2: This is not permenant. The subsequent PRs will continue to change the API dramatically, and I rather just update the docs when we have stabilized it.

shenyangHuang (approved):

PR #147, opened 2025-08-14
Move Discretize to DGData

Jacob-Chmura:


tgm/data.py
```diff
@@ -174,6 +174,87 @@ def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
...
                     self.dynamic_node_feats = self.dynamic_node_feats[node_order]
 
+    def discretize(
+        self, time_delta: TimeDeltaDG | str | None, reduce_op: str = 'first'
+    ) -> DGData:
+        if isinstance(time_delta, str):
+            time_delta = TimeDeltaDG(time_delta)
+
+        if time_delta is None or self.time_delta == time_delta:
+            return self  # TODO: Does not do a copy, clarify memory ownership
```
Jacob-Chmura: TODO: This should return a copy. Or else this could be problematic:

```python
data = DGData(.., time_delta='s')
discretized = data.discretize('Y')

assert data.time_delta == 's' # would be False
```

shenyangHuang: yes, it should return a copy since discretize modified the data and time granularity

examples/linkproppred/TGB/gcn.py
```diff
@@ -229,16 +229,25 @@ def eval(
 dataset.load_test_ns()
 
 train_data = DGData.from_tgb(args.dataset, split='train')
+train_data_discretized = train_data.discretize(args.snapshot_time_gran)
 train_dg = DGraph(train_data, train_data.time_delta, device=args.device)
```
ntgbaoo, line 233: Is `train_data.time_delta`/`val_data.time_delta`/`test_data.time_delta` the same with `args.time_gran`?

examples/linkproppred/TGB/gcn.py
```diff
@@ -229,16 +229,25 @@ def eval(
 dataset.load_test_ns()
 
 train_data = DGData.from_tgb(args.dataset, split='train')
```
ntgbaoo, line 231: Do we need to call `.discretize(args.time_gran)` first?

examples/nodeproppred/gclstm.py
```diff
@@ -162,22 +162,17 @@ def eval(
...
-test_data = DGData.from_tgb(args.dataset, split='test')
-test_dg = DGraph(test_data, test_data.time_delta, device=args.device).discretize(
-    args.time_gran
-)
+test_data = DGData.from_tgb(args.dataset, split='test').discretize(args.time_gran)
+test_dg = DGraph(test_data, test_data.time_delta, device=args.device)
 
-num_nodes = DGraph(args.dataset).num_nodes
+full_data = DGData.from_tgb(args.dataset)
+num_nodes = DGraph(full_data).num_nodes
```
ntgbaoo, line 175: nit: if we only need to get the total number of nodes. Would it be memory beneficial if we do this?
`num_nodes = len(test_dg.nodes | val_dg.nodes | train_dg.nodes))`

examples/linkproppred/TGB/gcn.py
Jacob-Chmura: Good point. I don't really see why we need `args.time_gran` at all. Unless I'm missing something, all that we need is the snapshot time gran. Frankly, I don't immediately see why we need discretization at all. Snapshot time gran seems to just determine our _iteration_, which is supposed to be specified with `batch_unit` in the data loader.

@shenyangHuang Can you elaborate on your implementation. Is it sufficient to keep a single copy of the graph (in native tgb granularity). And then have 2 data loaders:
- One by batch of edges (batch_unit='r', batch_size=200)
- One by snapshots (batch_unit='args.snapshot_time_gran')

examples/nodeproppred/gclstm.py
Jacob-Chmura: Yeah, in fact we can just take the max id since we assume the presence of node id `x` implies all node ids `y < x`. In any case, this will change once the `split()` api is done, and we can revisit then

examples/linkproppred/TGB/gcn.py
shenyangHuang: @Jacob-Chmura @benjaminnNgo It would be best to walk you both through these examples next time to understand everything more clearly. 

We do need `args.time_gran` and `args.snapshot_time_gran` both 
You are right that we are iterating by 2 dataloader:
- batch of edges (size 200 and standard in TGB), in default time gran that comes with the dataset usually in seconds
- one by **`discretized snapshots`**, here it is not just iterating by snapshot but you would group all edges within this snapshot into a single edge_index for the model to see. This means that any redundant edges must be removed / collasped. 

Using one extreme example to clarify. If for a daily snapshot, the original data has 200 of the same edge (occurring at different seconds in a day), the discretized snapshot will contain exactly a single edge and it just says the timestamp is that day (no more information about seconds). While if you data load with this, you would still spit out 200 edges in second wise granularity, there is no data modification.

shenyangHuang: Let's discuss this in a separate meeting next week, we should discuss amongst ourselves to be on the same page. Also time delta is unique to TG so we can think about the best approach together.

examples/linkproppred/tgn.py
```diff
@@ -404,7 +404,7 @@ def _init_hooks(dg: DGraph, sampling_type: str) -> List[DGHook]:
 )
 
 # Get global number of nodes for TGN Memory
-full_data, _ = DGData.from_tgb(args.dataset)
+full_data = DGData.from_tgb(args.dataset)
```
shenyangHuang, line 407: nice, no more native time gran xd

examples/linkproppred/TGB/gcn.py
shenyangHuang: I can go through the examples and check for any error as well once discretization API changes are done.

test/unit/test_data.py
```diff
@@ -800,3 +800,142 @@ def _get_exp_times():
...
+        edge_index,
+        edge_feats,
+        static_node_feats=static_node_feats,
+        time_delta='m',
+    )
+    new_granularity = TimeDeltaDG('h')
+    coarse_data = data.discretize(new_granularity, reduce_op='first')
+
+    assert coarse_data.time_delta == new_granularity
+    assert data.time_delta == TimeDeltaDG('m')
```
shenyangHuang, line 821: at first confused this with month. For string names, we could make it more unique later on. Right now, `M` is month, and `m` is minute, but it might not be very intuitive to new users to know. maybe `min` for minute? to change later.

tgm/data.py
```diff
@@ -174,6 +175,97 @@ def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
...
+                f'finer than the self.time_delta: {self.time_delta}'
+            )
+
+        valid_ops = ['first']
+        if reduce_op not in valid_ops:
+            raise ValueError(
+                f'Unknown reduce_op: {reduce_op}, expected one of: {valid_ops}'
+            )
+
+        # Note: If we simply return a new storage this will 2x our peak memory since the GC
```
shenyangHuang, line 201: for most things, this is a mask. Except for dealing with edge feature and node event feature where you have to use reduce. Potentially also counting edge weight, i.e. if from a fine time to coarse time, this edge is reduced to a single edge but it originally repeated 200 times, you can add 200 as an edge weight feature. We can flush out the reduce op in a separate issue.

tgm/graph.py
```diff
@@ -41,55 +41,6 @@ def __init__(
         self._device = torch.device(device)
         self._slice = DGSliceTracker()
 
-    def discretize(
```
shenyangHuang, line 44: nice, overall good to move IO / data operations out of `DGraph`

shenyangHuang (approved):
Approve, note to discuss the UTG examples next time @Jacob-Chmura @benjaminnNgo

PR #149, opened 2025-08-16
132 graph property prediction

benjaminnNgo:
## Objective 

Adding Graph Property Prediction tasks examples

**The integration tests of this PR may fail. Update from PR: https://github.com/tgm-team/tgm/pull/194#pullrequestreview-3171338706 is needed to fix the fail.**

## Excluded from this PR

- The days defined as `gap` between snapshots, used when generating labels in [GraphPulse](https://openreview.net/forum?id=DZqic2sPTY), are excluded from these examples. This is because the use case here focuses specifically on transaction networks, and the gap parameter may not be relevant in other domains
- Incorporations of node features into the graph-level embedding. Similar to `gap`, this may be only useful in a specific domain. To generalize the example, this is excluded.

## Testing evidence
**TGCN training log and performance on one token taken from [MiNT ](https://arxiv.org/abs/2406.10426)**

- Test AP = `0.6667`
- Test AUC = `0.8524`

TGCN-graphproppred

**Persistent forecast example**
- Test AP = `0.6538`
- Test AUC = `0.5750`

## Relevant PR

Closed https://github.com/tgm-team/tgm/issues/132


benjaminnNgo:
@shenyangHuang, @Jacob-Chmura. Can you guys take a quick look on [persistant_forecast.py](https://github.com/tgm-team/tgm/blob/132-Graph-Property-Prediction/examples/graphproppred/persistant_forecast.py) and let me know what do you guys think? This is just the daft, I will need to pull updates from Jacob regarding discretize when he is done.

I have a couple questions:
- Currently, I placed my datasets `.csv` file at `examples/graphproppred/test_token.csv` (you can't find it because I didn't push it yet). Should I create a directory named `sample_dataset` under graphproppred and place our sample data there? 
- I know Jacob is working on `.split()`. I am just wondering if I can pass train/val/test ratio for that?

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,168 @@
...
+import torch
+from torchmetrics import Metric, MetricCollection
+from torchmetrics.classification import BinaryAUROC, BinaryAveragePrecision
+from tqdm import tqdm
+
+from tgm import DGBatch, DGraph
+from tgm.loader import DGDataLoader
+
+parser = argparse.ArgumentParser(
+    description='TGCN Graph Property Example',
```
Jacob-Chmura:
```suggestion
    description='Persistant Forecast Graph Property Example',
```

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def edge_count(snapshot: DGBatch):
+    # return number of edges of current snapshot
+    return snapshot.src.shape[0]
+
+
+def node_count(snapshot: DGBatch):
+    # return number of ndoes of current snapshot
+    return torch.unique(torch.cat([snapshot.src, snapshot.dst])).numel()
```
Jacob-Chmura: Actually, the `DeduplicationHook` manages unique node ids, so we already have a build-in property for this:

https://github.com/tgm-team/tgm/blob/088faf210a2f46a398d02114605371211cf8ef97/tgm/hooks.py#L134

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,168 @@
...
+        current_measurement = self.comparator(batch)
+
+        if self.prev_measurement is None:
+            pred = torch.randint(0, 2, ())  # random guess for the first prediction
+        else:
+            pred = (
+                torch.tensor(1)
+                if current_measurement > self.prev_measurement
+                else torch.tensor(0)
+            )
```
Jacob-Chmura: Nit
```suggestion
            pred = torch.tensor(int(current_measurement > self.prev_measurement))
```

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,168 @@
...
+    loader: DGDataLoader,
+    y_true: torch.Tensor,
+    model: PersistantForecast,
+    metrics: Metric,
+    ignore_last_snapshot=False,
+) -> dict:
+    all_preds = []
+    number_of_snapshot = len(loader)
+    for idx, snapshot in tqdm(enumerate(loader)):
+        if not (ignore_last_snapshot and idx == number_of_snapshot - 1):
```
Jacob-Chmura: The data loader has an argument `drop_last`. Can we just use that to handle this last snapshot logic?

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,168 @@
...
+
+def eval(
+    loader: DGDataLoader,
+    y_true: torch.Tensor,
+    model: PersistantForecast,
+    metrics: Metric,
+    ignore_last_snapshot=False,
+) -> dict:
+    all_preds = []
+    number_of_snapshot = len(loader)
```
Jacob-Chmura: Warning: This is not completely reliable once #150 goes in. Specifically, if `on_empty='skip'` which is the default. We should instead store a `Counter` and increment it accordingly.

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,168 @@
...
+            labels.append(label)
+            prev_snapshot_metric = curr_snapshot_metric
+    return torch.tensor(labels, dtype=torch.int64)
+
+
+# ETL step
+def preproccess_raw_data(dataframe: pd.DataFrame) -> pd.DataFrame:
+    # time offset
+    start_time = dataframe['timestamp'].min()
+    dataframe['timestamp'] = dataframe['timestamp'].apply(lambda x: x - start_time)
```
Jacob-Chmura: Can be vectorized instead of applying row-wise lambda: `df['timestamp'] = df['timestamp'] - df['timestamp'].min()`

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # normalize edge weight
+    max_weight = float(dataframe['value'].max())
+    min_weight = float(dataframe['value'].min())
+    dataframe['value'] = dataframe['value'].apply(
+        lambda x: 1 + (9 * ((float(x) - min_weight) / (max_weight - min_weight)))
+    )
```
Jacob-Chmura: Likewise, can be vectorized:
```python
# normalize edge weight to [1, 10]
vmin, vmax = df["value"].min(), df["value"].max()
df["value"] = 1 + 9 * (df["value"] - vmin) / (vmax - vmin)
```

any reason why we normalize to this range instead of [0, 1]?

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,168 @@
...
+    # Key generator
+    node_id_map = {}
+    dataframe['from'] = dataframe['from'].apply(
+        lambda x: node_id_map.setdefault(x, len(node_id_map))
+    )
+    dataframe['to'] = dataframe['to'].apply(
+        lambda x: node_id_map.setdefault(x, len(node_id_map))
+    )
+
+    dataframe['value'] = dataframe['value'].apply(lambda x: [x])
```
Jacob-Chmura: Nit: move this up to preprocess 'value' all together

examples/graphproppred/tgcn.py
Jacob-Chmura, line 1: Ignoring this file for now

examples/graphproppred/persistant_forecast.py
ntgbaoo: I tried to set `drop_last = True`, but it doesn't work
```
drop_last (bool, optional): set to ``True`` to drop the last incomplete batch,
    if the dataset size is not divisible by the batch size. If ``False`` and
    the size of the dataset is not divisible by the batch size, then the last batch
    will be smaller. (default: ``False``)
```
Seems like it only drops the last incomplete batch?

examples/graphproppred/persistant_forecast.py
ntgbaoo: I will take note of this. I would probably set `on_empty= None`

examples/graphproppred/persistant_forecast.py
ntgbaoo: Good question. I believe there are some reasons behind it from the blockchain transaction networks. This way of normalization has been carried over from this work [Chartalist](https://proceedings.neurips.cc/paper_files/paper/2022/hash/e245189a86310b6667ac633dbb922d50-Abstract-Datasets_and_Benchmarks.html).

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,202 @@
...
+    all_preds = []
+    number_of_snapshot = len(loader)
+    for idx, snapshot in tqdm(enumerate(loader)):
+        if not (ignore_last_snapshot and idx == number_of_snapshot - 1):
+            pred = model(snapshot)
+            all_preds.append(pred)
+    y_pred = torch.Tensor(all_preds).float()
+    indexes = torch.zeros(y_pred.size(0), dtype=torch.long, device=y_pred.device)
+
+    # TODO: For debug purpose only. will be removed before merge to main
```
Jacob-Chmura:
```suggestion
```

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,202 @@
...
+            train_ratio=args.train_ratio,
+            val_ratio=args.val_ratio,
+            test_ratio=args.test_ratio,
+        )
+    )
+
+    full_dg = DGraph(full_data, device='cpu')
+    train_dg = DGraph(train_data, device='cpu')
+    val_dg = DGraph(val_data, device='cpu')
+    test_dg = DGraph(test_data, device='cpu')
```
Jacob-Chmura: Can just make use of default
```suggestion
    full_dg = DGraph(full_data)
    train_dg = DGraph(train_data)
    val_dg = DGraph(val_data)
    test_dg = DGraph(test_data)
```

examples/graphproppred/tgcn.py
```diff
@@ -0,0 +1,339 @@
...
+    opt: torch.optim.Optimizer,
+    metrics: Metric,
+) -> Tuple[float, torch.Tensor, torch.Tensor]:
+    model.train()
+    total_loss = 0
+    h_0 = None
+    criterion = torch.nn.BCELoss()
+    all_preds = []
+    for idx, snapshot in enumerate(tqdm(loader)):
+        # TODO: Consider skipping empty batches natively, when iterating by time (instead of events)
```
Jacob-Chmura: Just so I understand, the reason we are not doing `on_empty=skip` (the default) is because we are using the length of the dataloaders to split the labels after calling `label_generator_next_binary_classification` on the full dataset?

Can't we skip empty batches and still get the correct splits based on train/val/test ratios of the underlying `DGData` objects (instead of length of their loaders)?

ntgbaoo: I think I would set `on_empty=raise`, as it is quite important to choose `batch-time-gran` so that there is no empty snapshot.
But I am just wondering how we can get correct splits based on train/val/test ratios of the underlying `DGData` objects (instead of length of their loaders)?

benjaminnNgo:
Yes, the dataset is hosted there. But I like your idea of having `test_token.csv` hosted on another GitHub and downloading from it when running CI.

shenyangHuang:
quick thoughts:
- **The days defined as gap between snapshots**, I am not sure gap is specific to transaction network, it is more a problem formulation of the graph property prediction task, I would quickly discuss this before merging this PR. The `gap` in particular might result in leakage for models that uses the labels directly. when we predict for daily, the label is for next 1 week, then for the label of the last few day of the training set might contain information about the first few days to validation set. 
- **node feature**. Node feature is also not a domain specific feature. Similar to other examples in the library right now, if there is `static_node_feats` available in the `dg` then we should use it, otherwise not.

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,205 @@
...
+        lambda x: node_id_map.setdefault(x, len(node_id_map))
+    )
+    dataframe['to'] = dataframe['to'].apply(
+        lambda x: node_id_map.setdefault(x, len(node_id_map))
+    )
+
+    return dataframe
+
+
+class PersistantForecast:
```
shenyangHuang, line 93: maybe we would add this to `nn` similar to Edgebank ?

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,205 @@
...
+    )
+    train_loader = DGDataLoader(
+        train_dg, batch_unit=args.batch_time_gran, on_empty='raise'
+    )
+    val_loader = DGDataLoader(val_dg, batch_unit=args.batch_time_gran, on_empty='raise')
+    test_loader = DGDataLoader(
+        test_dg, batch_unit=args.batch_time_gran, on_empty='raise'
+    )
+
+    train_snapshots = len(train_loader)
```
shenyangHuang: I would not call them `train_snapshots` etc., as there is no graph data / snapshot contained in the various. call them `num_train_snapshots` instead or directly use `len(train_loader)` later on.

examples/graphproppred/persistant_forecast.py
```diff
@@ -0,0 +1,205 @@
...
+    val_snapshots = len(val_loader)
+    test_snapshots = len(test_loader) - 1
+
+    labels = label_generator_next_binary_classification(
+        loader=full_loader, snapshot_measurement=edge_count
+    )  # shape: number of snapshots - 1
+
+    train_labels = labels[:train_snapshots]
+    val_labels = labels[train_snapshots : train_snapshots + val_snapshots]
+    test_labels = labels[
```
shenyangHuang: I would do 

`test_labels = labels[len(train_loader) + len(val_loader): -1]`

examples/graphproppred/tgcn.py
```diff
@@ -0,0 +1,339 @@
...
+    )
+    train_loader = DGDataLoader(
+        train_dg, batch_unit=args.batch_time_gran, on_empty='raise'
+    )
+    val_loader = DGDataLoader(val_dg, batch_unit=args.batch_time_gran, on_empty='raise')
+    test_loader = DGDataLoader(
+        test_dg, batch_unit=args.batch_time_gran, on_empty='raise'
+    )
+
+    train_snapshots = len(train_loader)
```
shenyangHuang: confusing variable name, see other comments

examples/graphproppred/tgcn.py
```diff
@@ -0,0 +1,339 @@
...
+    val_snapshots = len(val_loader)
+    test_snapshots = len(test_loader) - 1
+
+    labels = label_generator_next_binary_classification(
+        loader=full_loader, snapshot_measurement=edge_count
+    )  # shape: number of snapshots - 1
+
+    train_labels = labels[:train_snapshots]
+    val_labels = labels[train_snapshots : train_snapshots + val_snapshots]
+    test_labels = labels[
```
shenyangHuang: same change here

examples/graphproppred/tgcn.py
```diff
@@ -0,0 +1,339 @@
...
+    ]
+
+    if train_dg.static_node_feats is not None:
+        static_node_feats = train_dg.static_node_feats
+    else:
+        static_node_feats = torch.randn(
+            (test_dg.num_nodes, args.node_dim), device=args.device
+        )
+    model = TGCN_Model(
+        node_dim=args.node_dim, embed_dim=args.embed_dim, num_classes=1
```
shenyangHuang: `node_dim=static_node_feats.shape[1]`

examples/graphproppred/tgcn.py
```diff
@@ -0,0 +1,339 @@
...
+        self.out = nn.Linear(in_dim, out_dim)
+
+    def forward(self, node_embed):
+        h = self.lin_node(node_embed)
+        h = h.relu()
+        h = self.out(h)
+        return h.sigmoid()
+
+
+def eval(
```
shenyangHuang: `@torch.no_grad()` see https://github.com/tgm-team/tgm/blob/c3b6eafc5eb510b52301dca0154e2f62efa072da/examples/linkproppred/gclstm.py#L122C1-L123C1

shenyangHuang (dismissed):

examples/graphproppred/persistant_forecast.py
ntgbaoo: I thought about it. But I actually checked [`EdgeBank`](https://github.com/tgm-team/tgm/blob/main/tgm/nn/memory/edgebank.py#L6C1-L6C25) and [`PersistantForecaster`](https://github.com/tgm-team/tgm/blob/main/examples/nodeproppred/persistant_forecast.py#L36), and saw that we don't have `nn` for them. I think it makes sense to not have `nn` as they are not really neural networks.

examples/graphproppred/persistant_forecast.py
ntgbaoo: Agreed. Revised names.

examples/graphproppred/persistant_forecast.py
ntgbaoo: Agreed. Resolved

examples/graphproppred/tgcn.py
ntgbaoo: Good catch! Thank you.

examples/graphproppred/tgcn.py
ntgbaoo: Right. Thank you!

Jacob-Chmura (approved):

PR #150, opened 2025-08-17
Handle Empty Batches

### LINKED ISSUE #14: Consider Dropping Empty Temporal Batches or Constraining Time Granularity

Jacob-Chmura:
# Purpose
The purpose of this PR is to add support to handle empty batches (num_events = 0) during iteration. We currently support the following policies:
- `on_empty='skip'`: (default) continues the iteration, skipping the empty batch
- `on_empty='raise'`: Raises a ValueError when an empty batch is encoutered
- `on_empty=None`: Yields the empty batch normally (what we currently have implemented in master).

## Design Tradeoffs
There are several obvious implementations that I did _not_ go with, and for good reason. The main difficulty is that why don't know _which_ batches will be empty ahead of time. This means we cannot do online iteration since the `torch.DataLoader` would not now anything about the batches we want to skip, and would re-yield them to the user.

### Option 1: Pre-compute the batches that are empty
This is possible. I didn't try it, because it's not ideal to have to iterate the entire graph at construction time. Not to mention each query requires materializing the batch (or a call to the storage backend) to actually see if our view is empty. (recall, our notion of empty is not that the start/end time have no overlap, but it's that no _events_ occured in an interval). Also, this would require storing an extra array, whereas loader should be lightweight.

### Option 2: Subclass `torch.IterableDataset` instead of `torch.DataLoader`
This is more akin to the behaviour we expect, we read from a data stream and don't know exactly which batch will be yielded ahead of time. This change has several downsides, mostly related to the API. Iterable loaders don't have `len()` and must be _reset_ after use manually, making them cumbersome for the user. Also, it's silly to use this when we have all our data up front.

### Option 3: Wrap an instance of the dataloder
This is the most straightforward. Idea would be to have the internal loader (with the main logic) wrapped in a user-facing class. The main downside here is that our DGDataloader would no longer subclass `torch.DataLoader`. And so, we would have to right a bit of boilerplate if we wanted it to be a drop in replacemement for torch loader.

### Option 4: Mixin Class
This is what I went with. We maintain subclassing `torch.DataLoader`, to get all their benefits out of the box. We define a private generic utility class (`SkippableDataLoaderMixin`) which holds all the logic for skipping batches. There is some finnacky multiple inheritence things that come up, but I don't think it's too bad. If this comes to bite us, we can go back to option 3 and keep the same API.

## Performance
There is obviously a runtime check that would only be avoided if we went with option 1 and pre-computed the batches that are empty. But this would make our end-to-end latency a bit worse, and seems not ideal. That being said, our  empty check does not come for free either, and we pay the cost on every batch.

It is important to realize that we cannot (currently) check if a graph is empty in `O(1)`. This is because we need to call the backend and see how many events actually exist in a time interval. If we have a cache miss, this requires binary searching the time stamps to get the interval range. Therefore, I this implementation **materializes the graph internally**. It calls all the hooks, and then skips the batch if it's empty. Technically, this might not be any better since materialization (even on an empty graph) requires the exact same binary search (plus other things).

If we find that this is not acceptable, then we can do a few things
- Mess around with method resolution order so that the `is_batch_empty` check happens on the unmaterialized graph. 
- Materialize the graph internally (without running all the hooks) and then do the check

Currently, the perforamnce will be exactly the same as what we have, since our current logic of materializing and going through all the hooks (and then in the example files skipping empty batches) is close to what's happening internally. Empty batches probably don't come up often (?) and it only dould occurs when iterating by time. 

## Relevant Prs
Close #14

### DIFF
--- a/tgm/loader.py
+++ b/tgm/loader.py
@@ -63,7 +63,7 @@ class DGDataLoader(_SkippableDataLoaderMixin, torch.utils.data.DataLoader):  # t
         dg: DGraph,
         batch_size: int = 1,
         batch_unit: str = 'r',
-        on_empty: Literal['skip', 'raise', None] = None,
+        on_empty: Literal['skip', 'raise', None] = 'skip',
         hook: HookManager | DGHook | List[DGHook] | None = None,
         **kwargs: Any,
     ) -> None:
--- a/tgm/loader.py
+++ b/tgm/loader.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
-from typing import Any, List
+from abc import ABC, abstractmethod
+from typing import Any, Iterator, List, Literal
 
 import torch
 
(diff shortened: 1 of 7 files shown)


tgm/loader.py
```diff
@@ -67,9 +106,15 @@ def __init__(
             slice_start = range(start_idx, stop_idx - batch_size, batch_size)
         else:
             slice_start = range(start_idx, stop_idx, batch_size)
-        super().__init__(slice_start, 1, shuffle=False, collate_fn=self, **kwargs)  # type: ignore
+
+        super().__init__(
```
shenyangHuang, line 110: just double checking, on_empty=None will be equivalent to what we currently has

Jacob-Chmura: yes

PR #152, opened 2025-08-18
Data Split API

### LINKED ISSUE #30: Avoiding Loading dataset multiple times for Test / Val in Hooks and Graph Construction
Currently when running 

```
train_dg = DGraph(args.dataset, time_delta=TimeDeltaDG('r'), split='train')
val_dg = DGraph(args.dataset, time_delta=TimeDeltaDG('r'), split='val')
test_dg = DGraph(args.dataset, time_delta=TimeDeltaDG('r'), split='test')

train_data = train_dg.materialize(materialize_features=False)
val_loader = DGDataLoader(
    val_dg,
    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')],
    batch_size=args.bsize,
)
test_loader = DGDataLoader(
    test_dg,
    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')],
    batch_size=args.bsize,
)
```

The TGB dataset will be loaded four times, three times for DGraph for each split and one time for negative sampler.

Jacob-Chmura:
Close #30

### DIFF
--- a/examples/linkproppred/gclstm.py
+++ b/examples/linkproppred/gclstm.py
@@ -153,13 +153,12 @@ def eval(
 args = parser.parse_args()
 seed_everything(args.seed)
 
-train_data = DGData.from_tgb(args.dataset, split='train').discretize(args.time_gran)
+
+data = DGData.from_tgb(args.dataset).discretize(args.time_gran)
+train_data, val_data, test_data = data.split()
+
 train_dg = DGraph(train_data, train_data.time_delta, device=args.device)
-
-val_data = DGData.from_tgb(args.dataset, split='val').discretize(args.time_gran)
 val_dg = DGraph(val_data, val_data.time_delta, device=args.device)
-
-test_data = DGData.from_tgb(args.dataset, split='test').discretize(args.time_gran)
 test_dg = DGraph(test_data, test_data.time_delta, device=args.device)
 
 train_loader = DGDataLoader(
--- a/examples/linkproppred/gclstm.py
+++ b/examples/linkproppred/gclstm.py
@@ -157,9 +157,9 @@ seed_everything(args.seed)
 data = DGData.from_tgb(args.dataset).discretize(args.time_gran)
 train_data, val_data, test_data = data.split()
 
-train_dg = DGraph(train_data, train_data.time_delta, device=args.device)
-val_dg = DGraph(val_data, val_data.time_delta, device=args.device)
-test_dg = DGraph(test_data, test_data.time_delta, device=args.device)
+train_dg = DGraph(train_data, device=args.device)
+val_dg = DGraph(val_data, device=args.device)
+test_dg = DGraph(test_data, device=args.device)
 
 train_loader = DGDataLoader(
     train_dg,
--- a/examples/linkproppred/gclstm.py
+++ b/examples/linkproppred/gclstm.py
@@ -146,8 +146,11 @@ args = parser.parse_args()
 seed_everything(args.seed)
 
 
-data = DGData.from_tgb(args.dataset).discretize(args.time_gran)
-train_data, val_data, test_data = data.split()
+train_data, val_data, test_data = DGData.from_tgb(args.dataset).split()
+
+train_data = train_data.discretize(args.time_gran)
+val_data = val_data.discretize(args.time_gran)
+test_data = test_data.discretize(args.time_gran)
 
 train_dg = DGraph(train_data, device=args.device)
 val_dg = DGraph(val_data, device=args.device)
--- a/examples/linkproppred/gclstm.py
+++ b/examples/linkproppred/gclstm.py
@@ -153,14 +153,20 @@ def eval(
 args = parser.parse_args()
 seed_everything(args.seed)
 
-train_data, time_delta = DGData.from_tgb(args.dataset, split='train')
-train_dg = DGraph(train_data, time_delta, device=args.device).discretize(args.time_gran)
+train_data = DGData.from_tgb(args.dataset, split='train')
+train_dg = DGraph(train_data, train_data.time_delta, device=args.device).discretize(
+    args.time_gran
+)
 
-val_data, time_delta = DGData.from_tgb(args.dataset, split='val')
-val_dg = DGraph(val_data, time_delta, device=args.device).discretize(args.time_gran)
+val_data = DGData.from_tgb(args.dataset, split='val')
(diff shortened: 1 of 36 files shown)


tgm/split.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __post_init__(self) -> None:
+        total = self.train_ratio + self.val_ratio + self.test_ratio
+        if abs(total - 1.0) > 1e-6:
+            raise ValueError(
+                f'train_ratio + val_ratio + test_ratio must sum to 1.0, got {total}'
+            )
+
+    def apply(self, data: 'DGData') -> Tuple['DGData', 'DGData', 'DGData']:  # type: ignore
+        return data, data, data
```
ntgbaoo: What is the behaviour of RatioSplit if the ratios passed are summed up to less than 1 (e.g 0.5/0.15/0.15)?

tgm/data.py
```diff
@@ -482,48 +496,28 @@ def _check_pandas_import(min_version_number: str | None = None) -> None:
...
-                err_msg = 'User requires tgb to initialize a DGraph from a tgb dataset '
-                raise ImportError(err_msg)
-
-        LinkPropPredDataset, NodePropPredDataset = _check_tgb_import()
+    def from_tgb(cls, name: str, **kwargs: Any) -> DGData:
+        try:
+            from tgb.linkproppred.dataset import LinkPropPredDataset
+            from tgb.nodeproppred.dataset import NodePropPredDataset
+        except ImportError:
+            raise ImportError('TGB required to load TGB data, try `pip install py-tgb`')
```
ntgbaoo, line 508: Is it 'TGM required to load TGB data, try `pip install py-tgb`'?

tgm/split.py
ntgbaoo, line 1: So users need to define the implementation of `apply()` functions themselves?

tgm/data.py
Jacob-Chmura: Im trying to say, the user needs _TGB_ in order to use TGB datasets

tgm/split.py
Jacob-Chmura: I just put in stubs for now, we will implement the apply functions before merging.

If a user wants to define their own split strategy, then yes, they'll have to implement apply.

tgm/split.py
Jacob-Chmura: True, ill see what libraries like sklearn do regarding these data splits, and follow suite

Jacob-Chmura: Similarly, if test time is < val time in the timesplitter

tgm/data.py
```diff
@@ -30,6 +31,8 @@ class DGData:
 
     static_node_feats: Tensor | None = None  # [num_nodes, D_node_static]
 
+    _split_strategy: SplitStrategy | None = None
```
Jacob-Chmura, line 34: Should add repr(False), equal(False)

tgm/data.py
```diff
@@ -124,13 +127,17 @@ def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
...
-            ):
+            if self.static_node_feats.ndim != 2:
                 raise ValueError(
-                    'static_node_feats must have shape [num_nodes, D_node_static], '
-                    f'got {num_nodes} nodes and shape {self.static_node_feats.shape}'
+                    f'static_node_feats must be a 2D tensor of shape [N, D_node_static], '
+                    f'but got ndim={self.static_node_feats.ndim} with shape {self.static_node_feats.shape}'
+                )
+
+            if self.static_node_feats.shape[0] < num_nodes:
```
Jacob-Chmura, line 136: We changed this from strict equality check to accomodate the fact that post `data.split()` the _train/val/test_ will have a static node feature reference which is a _superset_ of the nodes on that single split. 

Of course, this, as everywhere else, assumes the number of nodes is determined by the max node it.

tgm/data.py
```diff
@@ -583,3 +577,11 @@ def _check_tgb_import() -> tuple['LinkPropPredDataset', 'NodePropPredDataset']:
             dynamic_node_feats=dynamic_node_feats,
             static_node_feats=static_node_feats,
         )
+
+        split_bounds = {
+            'train': (int(train_times.min().item()), int(train_times.max().item())),
+            'val': (int(val_times.min().item()), int(val_times.max().item())),
+            'test': (int(test_times.min().item()), int(test_times.max().item())),
+        }
```
Jacob-Chmura: It is sufficient to just store the the upper and lower bound edge times for each split. Alternatively, we could store the entire edge timestamp mask, but this would require more complex logic in the `TGBSplit` application

tgm/split.py
```diff
@@ -0,0 +1,151 @@
...
+            edge_start_time, edge_end_time = self.split_bounds[split_name]
+            edge_mask = (data.timestamps[data.edge_event_idx] >= edge_start_time) & (
+                data.timestamps[data.edge_event_idx] <= edge_end_time
+            )
+
+            node_mask = None
+            if data.node_ids is not None:
+                node_times = data.timestamps[data.node_event_idx]
+                if edge_mask.any():
+                    node_mask = (node_times >= (edge_start_time - 1)) & (
```
Jacob-Chmura, line 141: Using the weird tgb - 1 logic here akin to: https://github.com/tgm-team/tgm/blob/5bec4b724b62cd0a40c7b6ca519071edfeb54826/tgm/data.py#L531-L542

Not ideal that it's duplicated

examples/linkproppred/gclstm.py
```diff
@@ -145,14 +145,13 @@ def eval(
 args = parser.parse_args()
 seed_everything(args.seed)
 
-train_data = DGData.from_tgb(args.dataset, split='train').discretize(args.time_gran)
-train_dg = DGraph(train_data, train_data.time_delta, device=args.device)
 
-val_data = DGData.from_tgb(args.dataset, split='val').discretize(args.time_gran)
-val_dg = DGraph(val_data, val_data.time_delta, device=args.device)
+data = DGData.from_tgb(args.dataset).discretize(args.time_gran)
```
shenyangHuang: interesting. 

if you do 
```
train_data, _ , _  = data.split()
train_data.discretize()
```

is it equivalent to 
```
data.discretize()
train_data, _ , _  = data.split()
```

shenyangHuang: should we standardize this also across examples?

PR #156, opened 2025-08-19
Privatize DGStorage Construction

### LINKED ISSUE #155: Make DGraph storage construction private

Jacob-Chmura:
# Purpose
The purpose of this PR is to simplify the `DGraph` constuctor by moving the _internal_ `DGStorage` based construction into a private class method.

This makes the API extremely clear to the user:
```python
dg = DGraph(data: DGData, device: str | torch.device)
```

The time delta is now fully manged by `DGData`, and we avoid confusion for passing `time_delta` itno DGraph (which is now only possible via view_creation.

## Changes
New private class method `from_storage` which we use internally to create our `DGraph` views. Public constructor is only for initial graph construction (from `DGData` by the user).
 
## Relevant Issues
Close #155

### DIFF
--- a/tgm/graph.py
+++ b/tgm/graph.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from collections.abc import Iterable, Sized
-from dataclasses import dataclass
+from dataclasses import dataclass, replace
 from functools import cached_property
 from typing import Any, Optional, Set, Tuple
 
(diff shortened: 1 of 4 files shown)


shenyangHuang:
make sense, given that we are moving all the time granularity related operations into DGData

tgm/graph.py
```diff
@@ -16,28 +16,16 @@
...
-        elif isinstance(data, DGStorage):
-            if time_delta is None:
-                raise RuntimeError(
-                    'Internally tried creating DGraph on existing storage without specifying a time_delta'
-                )
+    def __init__(self, data: DGData, device: str | torch.device = 'cpu') -> None:
+        if not isinstance(data, DGData):
+            raise TypeError(f'DGraph must be initialized with DGData, got {type(data)}')
+
+        if isinstance(data.time_delta, str):
```
shenyangHuang: If `time_delta` come from `DGData.time_delta`, why would it be a string? shouldn't `DGData.time_delta` always be initialized as `TimeDeltaDG` already.

shenyangHuang (approved):

tgm/graph.py
Jacob-Chmura: You are right, technically this is something we handle in `post_init`. But the type on `DGData` is technically `str | TimeDeltaDG`. We can simplify this, but will require to remember that the `TimeDelta` must be _normalized_ in `DGData`.

Alternatively, we just allow `TimeDelta()` to be constructed from another timedelta. Similar to torch:
```python
device = torch.device('cpu')

torch.device('cpu')
torch.device(device) # ok
```

PR #163, opened 2025-08-20
Hook State API Change

### LINKED ISSUE #160: Add statefulness property to hooks

Jacob-Chmura:
### Purpose
The purpose of this PR is to add the minimal API change required to satisfy the hook state management planned for beta release.
Namely, the ability to reset state on all hooks.

### Changes
- hook manager has `reset_state()` api, which resets all hooks.
- `DGHook` protocol enforces that each hook now implements `has_state: bool` property, and `reset_state()` method. For stateless hooks, these are no-ops.
- To reduce boilerplate, I created a `StatelessHook` protocol which has these as no-ops.

### Out of scope
Going to modify the examples in subsequent PR. This also requires thining about our dataloading API which I wrote to implicitly allow `HookManager` construction: https://github.com/tgm-team/tgm/blob/5bec4b724b62cd0a40c7b6ca519071edfeb54826/tgm/loader.py#L72

If the user is to completely manage state via hook manager, it makes sense from an API perspective to expose direct hook manager injection instead. This will require more changes to loader, hence, why I am splitting things up.

### Relevant Issues
Close #160

### DIFF
--- a/tgm/hooks.py
+++ b/tgm/hooks.py
@@ -12,12 +12,27 @@ from tgm._storage import DGSliceTracker
 
 @runtime_checkable
 class DGHook(Protocol):
-    requires: Set[str]
-    produces: Set[str]
-
     r"""The behaviours to be executed on a DGraph before materializing."""
 
+    requires: Set[str]
+    produces: Set[str]
+    has_state: bool
+
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch: ...
+
+    def reset_state(self) -> None: ...
+
+
+class StatelessHook:
+    requires: Set[str] = set()
+    produces: Set[str] = set()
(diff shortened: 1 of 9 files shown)


tgm/hooks.py
```diff
@@ -12,13 +12,28 @@
...
-
-    r"""The behaviours to be executed on a DGraph before materializing."""
+    has_state: bool
 
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch: ...
 
+    def reset_state(self) -> None: ...
+
+
+class StatelessHook:
```
shenyangHuang, line 26: if we have this then why don't we have a `statefulHook` ?

shenyangHuang (approved):

PR #168, opened 2025-08-22
Update Temporal Attention

### LINKED ISSUE #167: Temporal Attention Upgrades

Jacob-Chmura:
### Purpose
The purpose of this PR is to port over the isolated changes to `TemporalAttention` module that were discovered while debugging against dyglib (#138 )

#### Implicit `out_dim`
https://github.com/tgm-team/tgm/blob/0265757eb6bead40bfbba633957ff8825802c4a1/tgm/nn/attention.py#L28
Instead of paramterizing `out_dim`, we hard code it to `node_dim + time_dim` as is standard in _temporal_ attention modules. This is not the case in general attention modules, but seems to be the case here.

#### Padding

https://github.com/tgm-team/tgm/blob/0265757eb6bead40bfbba633957ff8825802c4a1/tgm/nn/attention.py#L30-L32

If `out_dim % n_heads != 0`, we previously raised an error. Instead, we know apply zero-padding. Note, this also must happen at runtime to pad features if needed:

https://github.com/tgm-team/tgm/blob/0265757eb6bead40bfbba633957ff8825802c4a1/tgm/nn/attention.py#L55

#### Residual Connection

https://github.com/tgm-team/tgm/blob/0265757eb6bead40bfbba633957ff8825802c4a1/tgm/nn/attention.py#L90

I just forgot this

### Relevant Issues
Close #167

### DIFF
--- a/tgm/nn/attention.py
+++ b/tgm/nn/attention.py
@@ -45,30 +45,29 @@ class TemporalAttention(torch.nn.Module):
 
     def forward(
         self,
-        node_feat: torch.Tensor,  # (batch_size, node_dim)
-        time_feat: torch.Tensor,  # (batch_size, time_dim)
-        edge_feat: torch.Tensor,  # (batch_size, num_nbrs, edge_dim)
-        nbr_node_feat: torch.Tensor,  # (batch_size, num_nbrs, node_dim)
-        nbr_time_feat: torch.Tensor,  # (batch_size, num_nbrs, time_dim)
-        nbr_mask: torch.Tensor,  # (batch_size, num_nbrs)
-    ) -> torch.Tensor:  # (batch_size, out_dim)
+        node_feat: torch.Tensor,  # (B, node_dim)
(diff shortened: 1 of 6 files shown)


tgm/nn/attention.py
```diff
@@ -19,49 +19,55 @@ def __init__(
         node_dim: int,
         edge_dim: int,
         time_dim: int,
-        out_dim: int,
         dropout: float = 0.1,
     ) -> None:
         super().__init__()
-        if any((x <= 0 for x in [n_heads, node_dim, edge_dim, time_dim, out_dim])):
+        if any((x <= 0 for x in [n_heads, node_dim, edge_dim, time_dim])):
```
shenyangHuang, line 25: yes, we should enforce all of these to be non `0`.  We should standardize this across all `nn` layers, for example, `node_dim` > = 0

shenyangHuang (approved):

PR #171, opened 2025-08-24
Fixing Uniform Sampler

shenyangHuang:
- solved the current batch leak error by only sampling from `batch.time.min()-1`
- making the unit test deterministic


tgm/hooks.py
```diff
@@ -266,8 +266,8 @@ class NeighborSamplerHook(StatelessHook):
     def __init__(self, num_nbrs: List[int]) -> None:
         if not len(num_nbrs):
             raise ValueError('num_nbrs must be non-empty')
-        if not all([isinstance(x, int) and (x == -1 or x > 0) for x in num_nbrs]):
-            raise ValueError('Each value in num_nbrs must be a positive integer or -1')
+        if not all([isinstance(x, int) and (x > 0) for x in num_nbrs]):
```
Jacob-Chmura: We should update the tests to catch regressions in case we end up wanting to support this api

tgm/_storage/backends/array_backend.py
```diff
@@ -74,6 +74,7 @@ def get_nbrs(
         seed_nodes: Tensor,
         num_nbrs: int,
         slice: DGSliceTracker,
+        empty: int = -1,
```
Jacob-Chmura: We can leave this for now, but this should not be customizable I dont think. I would create a global constant that we use throughout the library (and the test suite)

test/unit/test_storage_impl.py
```diff
@@ -459,3 +360,184 @@ def test_set_dg_storage_backend_with_str():
...
+        [[1], [2], [5], [2]]
+    )  # edge feat is simply summing the node IDs at two end points
+    data = DGData.from_raw(edge_timestamps, edge_index, edge_feats)
+    return data
+
+
+def test_dg_storage_get_nbrs_one_hop(basic_sample_graph):
+    dg = DGraph(basic_sample_graph)
+    n_nbrs = 3  # 3 neighbor for each node
+    loader = DGDataLoader(dg, hook=[], batch_size=1)
```
Jacob-Chmura: I don't think we need DGraph and DGDataloader, and the storage tests shouldn't depend on them. Instead, we can directly query the backend like my other tests

tgm/hooks.py
```diff
@@ -304,7 +304,7 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
             nbr_nids, nbr_times, nbr_feats, nbr_mask = dg._storage.get_nbrs(
                 seed_nodes,
                 num_nbrs=num_nbrs,
-                slice=DGSliceTracker(end_time=int(batch.time.min())),
+                slice=DGSliceTracker(end_time=(int(batch.time.min()) - 1)),
```
Jacob-Chmura:
```suggestion
                slice=DGSliceTracker(end_time=int(batch.time.min()) - 1), 
```

tgm/_storage/backends/array_backend.py
```diff
@@ -102,18 +103,21 @@ def get_nbrs(
...
-        nbr_nids = torch.full((B, num_nbrs), -1, dtype=torch.long, device=device)
-        nbr_times = torch.zeros(B, num_nbrs, dtype=torch.long, device=device)
-        nbr_feats = torch.zeros(B, num_nbrs, self.get_edge_feats_dim(), device=device)  # type: ignore
+        nbr_nids = torch.full((B, num_nbrs), empty, dtype=torch.long, device=device)
+        nbr_times = torch.full((B, num_nbrs), empty, dtype=torch.long, device=device)
+        nbr_feats = torch.full(  # type: ignore
+            (B, num_nbrs, self.get_edge_feats_dim()),
+            float(empty),
+            device=device,  # type: ignore
+        )
```
Jacob-Chmura: I think these can still be zeros. As long as the nbr nids are marked correctly.

Jacob-Chmura (changes requested):

tgm/hooks.py
Jacob-Chmura: Also might be worthwhile adding an integration test that actually uses the unfiorm sampler, just so that we ensure no api breaking changes

shenyangHuang: added a new integration test for tgat using uniform with neighbors [5,5]

tgm/_storage/backends/array_backend.py
shenyangHuang: fixed

test/unit/test_storage_impl.py
shenyangHuang: yes, updated

PR #178, opened 2025-08-25
Add Sentinal Node Id for Marking Invalid Neighbors

### LINKED ISSUE #177: Clear Sentinal Value for "Empty" Node Id in Sampling

Jacob-Chmura:
### Purpose
The purpose of this PR is to add a global `tgm` library constant `INVALID_NODE_ID` which denotes the node id that we use for marking _invalid neighors_ during sampling. Currently, we chose `-1`. It will be annoyuing to users who have negative node ids, but we don't see this as the common situation.

#### Added `INVALID_NODE_ID`
A global constant available in `tgm/constants.py`.

#### Removed `nbr_mask` from the sampler API
The `nbr_mask` is not used to recursively prune neighborhood info. Instead, our nbr_mask can _always_ be computed at any stage by doing `batch.nbr_nids[i] == INVALID_NODE_ID`. Therefore, I think there is no need to return this mask explicitly, as it's sort of redundant.

#### Samplers
Samplers just use the placehodlers constant `INVALID_NODE_ID` instead of hardcoding `-1` for makring incomplete neighborhood searches.

#### Data Loading
I am now throwing an exception if either the edge events or node events contains node ids that match the `INVALID_NODE_ID` sentinal value.

#### Attention
This one we can think about. I decided to rename `nbr_mask` to `invalid_nbr_mask` to make it more clear that this should be a boolearn tensor marking _invalid_ node positions, rather than valid ones (this is what we used for masked softmax).

### Relevant Issues
Close #177

### DIFF
--- a/examples/linkproppred/TGB/tgat.py
+++ b/examples/linkproppred/TGB/tgat.py
@@ -11,6 +11,7 @@ from tgb.linkproppred.evaluate import Evaluator
 from tqdm import tqdm
 
 from tgm import DGBatch, DGData, DGraph
+from tgm.constants import INVALID_NODE_ID
 from tgm.hooks import (
     DGHook,
     NegativeEdgeSamplerHook,
@@ -126,7 +127,7 @@ class TGAT(nn.Module):
                 edge_feat=batch.nbr_feats[hop],
                 nbr_node_feat=nbr_feat,
                 nbr_time_feat=nbr_time_feat,
-                nbr_mask=nbr_mask,
+                invalid_nbr_mask=nbr_mask == INVALID_NODE_ID,
             )
             z[batch.global_to_local(seed_nodes)] = out
         return z
--- a/tgm/nn/attention.py
+++ b/tgm/nn/attention.py
@@ -50,7 +50,7 @@ class TemporalAttention(torch.nn.Module):
         edge_feat: torch.Tensor,  # (B, num_nbrs, edge_dim)
         nbr_node_feat: torch.Tensor,  # (B, num_nbrs, node_dim)
         nbr_time_feat: torch.Tensor,  # (B, num_nbrs, time_dim)
-        nbr_mask: torch.Tensor,  # (B, num_nbrs)
+        invalid_nbr_mask: torch.Tensor,  # (B, num_nbrs)
     ) -> torch.Tensor:  # (B, out_dim)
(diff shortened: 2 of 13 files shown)


tgm/nn/attention.py
```diff
@@ -50,7 +50,7 @@ def forward(
         edge_feat: torch.Tensor,  # (B, num_nbrs, edge_dim)
         nbr_node_feat: torch.Tensor,  # (B, num_nbrs, node_dim)
         nbr_time_feat: torch.Tensor,  # (B, num_nbrs, time_dim)
-        nbr_mask: torch.Tensor,  # (B, num_nbrs)
+        invalid_nbr_mask: torch.Tensor,  # (B, num_nbrs)
```
Jacob-Chmura: Maybe `valid_nbr_mask` is more intuitive? Then the caller has to do `node_ids != INVALID_NODE_ID`. Both might be fine

shenyangHuang:
User should never have negative node IDs, we can assume that.

examples/linkproppred/TGB/tgat.py
```diff
@@ -126,7 +127,7 @@ def forward(
                 edge_feat=batch.nbr_feats[hop],
```
shenyangHuang, line 127: what happens to TGAT forward pass if in the batch, there is no history and every neighbor is `INVALID_NODE_ID` ? Not immediately clear to me. This happens on first batch when there is no neighbors

tgm/nn/attention.py
shenyangHuang: yeah I think either is fine. One way is `node_ids[nbr_mask]` which is you mask out things you don't need. I am fine with either

benjaminnNgo:
In my DyGFormer implementation, instead of calling it an invalid node ID, I call it [`PADDED_NODE_ID`](https://github.com/tgm-team/tgm/blob/6-DygformerImplementation/tgm/nn/model/dygformer.py#L12). I am fine with `INVALID_NODE_ID`.

examples/linkproppred/TGB/tgat.py
Jacob-Chmura: Attention matrix becomes all 1e-10. Softmax attention becomes basically uniform random.

Jacob-Chmura: I think I saw some other libraries special case this and return the zero vector

Issue #106, opened 2025-07-04
TGM Hook Tutorial


shenyangHuang:
create detailed hook tutorial for tgm.
Introducing existing hooks while showing how to create custom ones. 
also discuss hook composability. 

talk about statefullness

shenyangHuang:
Also need to discuss the Hook Manager,  reset hooks, scoping with train, val, test

PR #179, opened 2025-08-25
Hook Management Registry

### LINKED ISSUE #157: Recency Sampler state need to be carried from train to val to test
In normal TGL workflow, the recency sampler should store information from train set and then carry it to validation set and then carry it to test set.

Jacob-Chmura:
### Purpose
The purpose of this PR is to finalize the hook state management API / semantics for Beta release. The primary purpose is to enable users to share and reset hooks generically across dataloaders etc.

#### Registry
- Each hook is now associated with a string `key`. The allows user to specifically register hooks for `train`, `val`, etc. There is a seperate API for registered _shared_ hooks, which will be added to _all_ keys:

https://github.com/tgm-team/tgm/blob/3224d757c49ec0077315710404eb1b537b95eb02/tgm/hooks/hook_manager.py#L34

https://github.com/tgm-team/tgm/blob/3224d757c49ec0077315710404eb1b537b95eb02/tgm/hooks/hook_manager.py#L20

#### Hook Activation
The provide a basic key-value style API for getting, setting, and activating hooks via a particular `key`. The idea is that the user can activate `train` keys during training, `val` keys during validation etc.  

https://github.com/tgm-team/tgm/blob/3224d757c49ec0077315710404eb1b537b95eb02/tgm/hooks/hook_manager.py#L50-L63

We additional provide a context manager so that a user can more clearly seperate and see which hooks are active. See, for instance, the current example:

https://github.com/tgm-team/tgm/blob/3224d757c49ec0077315710404eb1b537b95eb02/examples/linkproppred/TGB/tgat.py#L279-L283

#### Hook State Management
We still have the reset state method which now optionally takes a key (to only reset state for hooks on that specific key). Considered making this a context manager but decided this is simpler for now.

#### Additional Points
- Shared hooks are automatically added to all new _keyed hooks_.
- We only allow users to register new hooks while there is not active key (i.e. before the hot path of the code). This enables us to _pre-compute_ valid topological hook ordering outside of the data loader. See the next point.
- Since we opaquely manage keyed and shared hooks, we now need to ensure the ordering is valid (i.e. requires atributes are on the batch before produces attributes. I implemented topological sort for this purpose. We do this during registration phase.
- The device hook now has to be explicitly added :( unless we let the hook manager take a DGraph. But that's not clear as to which dgraph it should take (train/val/test). So, it's simpler to just registered device movement as a shared hook (and it could be non-shared in case for some reason user wants validation or test on a differen device, for instance).

### Other Changes
- Dataloader now takes a single `HookManager` instead of implicitly constructing it. That way, it is shared among all data loaders.
- Updated `TGAT example to show how it would work
- Created a `hooks` directory and split up some of our hook code

### To Discuss
Currently the topo sort is not really useful. This is because
- we only enable registering hooks one at a time
- we eagerly run the topological sort.

Because of this, the user will immediately get the 'missing produces' exception before topological sort even runs. Hence, they end up having to be aware of the order that they register hooks.

**Option 1**: Leave this as is. We fail fast so that as soon as a hook is registered, all the requires must already be satisfied. This is simplest, but hardest on the user.

**Option 2**: Leave as is, but add ability to register multiple hooks at once. This way, we stil fail fast, but now we actual have a reason to do the sort: the user can input multiple hooks at once in any order, and we find the right order.

**Option 3**: Don't do the topological sort eagerly. Instead, let users add hooks as they wish. Defer the topological check until we actually active / execute the hooks. We can still cache the result of the topological sort, so that it's only executed once per key. (assumign the user does not re-registry a hook, but we already structured the api such that this is not encouraged).
Let's discuss this tomorrow, otherwise, the unit tests, and exampels have been updated, and pending this detail we can merge.
My preference is either (3). If we do (3), we can add (2) as an additional features down the line. Doign only (1) and leaving (2) and or (3) for full release is valid too, but requires user to think about which order to register hooks and shared hooks.

I added a string method that nicely shows all the registered hooks and their dependency resolution order, but it's still better in my opinion if user can just blindly register hooks and the manager handles the execution / ordering.

### Relevant Issues
Close #157

### DIFF
--- a/examples/linkproppred/TGB/edgebank.py
+++ b/examples/linkproppred/TGB/edgebank.py
@@ -7,7 +7,7 @@ from tgb.linkproppred.evaluate import Evaluator
 from tqdm import tqdm
 
 from tgm import DGData, DGraph
-from tgm.hooks import TGBNegativeEdgeSamplerHook
+from tgm.hooks import HookManager, TGBNegativeEdgeSamplerHook
 from tgm.loader import DGDataLoader
 from tgm.nn import EdgeBankPredictor
 from tgm.util.seed import seed_everything
@@ -75,16 +75,15 @@ val_dg = DGraph(val_data)
 test_dg = DGraph(test_data)
 
 train_data = train_dg.materialize(materialize_features=False)
-val_loader = DGDataLoader(
-    val_dg,
-    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')],
-    batch_size=args.bsize,
-)
-test_loader = DGDataLoader(
-    test_dg,
-    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')],
-    batch_size=args.bsize,
-)
+val_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
+test_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')
+
+hm = HookManager()
+hm.register('val', val_neg_hook)
+hm.register('test', test_neg_hook)
+
+val_loader = DGDataLoader(val_dg, args.bsize, hook_manager=hm)
+test_loader = DGDataLoader(test_dg, args.bsize, hook_manager=hm)
 
 model = EdgeBankPredictor(
     train_data.src,
@@ -95,7 +94,10 @@ model = EdgeBankPredictor(
     pos_prob=args.pos_prob,
 )
 
-val_results = eval(val_loader, model, eval_metric, evaluator)
-print(' '.join(f'{k}={v:.4f}' for k, v in val_results.items()))
-test_results = eval(test_loader, model, eval_metric, evaluator)
-print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
+with hm.activate('val'):
+    val_results = eval(val_loader, model, eval_metric, evaluator)
+    print(' '.join(f'{k}={v:.4f}' for k, v in val_results.items()))
+
+with hm.activate('test'):
+    test_results = eval(test_loader, model, eval_metric, evaluator)
+    print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
--- a/examples/linkproppred/TGB/edgebank.py
+++ b/examples/linkproppred/TGB/edgebank.py
@@ -78,7 +78,7 @@ train_data = train_dg.materialize(materialize_features=False)
 val_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
 test_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')
 
-hm = HookManager()
+hm = HookManager(keys=['val', 'test'])
 hm.register('val', val_neg_hook)
 hm.register('test', test_neg_hook)
 
--- a/examples/linkproppred/TGB/gcn.py
+++ b/examples/linkproppred/TGB/gcn.py
@@ -16,7 +16,11 @@ from torch_geometric.nn import GCNConv
 from tqdm import tqdm
 
 from tgm import DGBatch, DGData, DGraph
-from tgm.hooks import NegativeEdgeSamplerHook, TGBNegativeEdgeSamplerHook
+from tgm.hooks import (
+    HookManager,
+    NegativeEdgeSamplerHook,
+    TGBNegativeEdgeSamplerHook,
+)
 from tgm.loader import DGDataLoader
 from tgm.timedelta import TimeDeltaDG
 from tgm.util.seed import seed_everything
@@ -230,7 +234,6 @@ evaluator = Evaluator(name=args.dataset)
 dataset.load_val_ns()
 dataset.load_test_ns()
 
-
 train_data, val_data, test_data = DGData.from_tgb(args.dataset).split()
 
 train_data_discretized = train_data.discretize(args.snapshot_time_gran)
@@ -245,30 +248,31 @@ test_data_discretized = test_data.discretize(args.snapshot_time_gran)
 test_dg = DGraph(test_data, device=args.device)
 test_snapshots = DGraph(test_data_discretized, device=args.device)
 
-train_loader = DGDataLoader(
-    train_dg,
-    hook=[NegativeEdgeSamplerHook(low=0, high=train_dg.num_nodes)],
-    batch_size=args.bsize,
-)
+
+_, dst, _ = train_dg.edges
+train_neg_hook = NegativeEdgeSamplerHook(low=int(dst.min()), high=int(dst.max()))
+val_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
+test_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')
+
+hm = HookManager(args.device)
+hm.register('train', train_neg_hook)
+hm.register('val', val_neg_hook)
+hm.register('test', test_neg_hook)
+
+train_loader = DGDataLoader(train_dg, args.bsize, hook_manager=hm)
+val_loader = DGDataLoader(val_dg, args.bsize, hook_manager=hm)
+test_loader = DGDataLoader(test_dg, args.bsize, hook_manager=hm)
+
 train_snapshots_loader = DGDataLoader(
-    train_snapshots,
-    hook=[NegativeEdgeSamplerHook(low=0, high=train_dg.num_nodes)],
-    batch_unit=args.snapshot_time_gran,
+    train_snapshots, batch_unit=args.snapshot_time_gran, hook_manager=hm
+)
+val_snapshots_loader = DGDataLoader(
+    val_snapshots, batch_unit=args.snapshot_time_gran, hook_manager=hm
+)
+test_snapshots_loader = DGDataLoader(
+    test_snapshots, batch_unit=args.snapshot_time_gran, hook_manager=hm
 )
 
-val_loader = DGDataLoader(
-    val_dg,
-    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')],
-    batch_size=args.bsize,
-)
-val_snapshots_loader = DGDataLoader(val_snapshots, batch_unit=args.snapshot_time_gran)
-
-test_loader = DGDataLoader(
-    test_dg,
-    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')],
-    batch_size=args.bsize,
-)
-test_snapshots_loader = DGDataLoader(test_snapshots, batch_unit=args.snapshot_time_gran)
 
 if train_dg.static_node_feats is not None:
     static_node_feats = train_dg.static_node_feats
@@ -294,23 +298,43 @@ opt = torch.optim.Adam(
 )
 
 for epoch in range(1, args.epochs + 1):
-    start_time = time.perf_counter()
-    loss, z = train_in_batches(
-        train_loader,
-        train_snapshots_loader,
-        static_node_feats,
-        encoder,
-        decoder,
-        opt,
-        conversion_rate,
-    )
-    end_time = time.perf_counter()
-    latency = end_time - start_time
-    print(f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f}')
+    with hm.activate('train'):
+        start_time = time.perf_counter()
+        loss, z = train_in_batches(
+            train_loader,
+            train_snapshots_loader,
+            static_node_feats,
+            encoder,
+            decoder,
+            opt,
+            conversion_rate,
+        )
+        end_time = time.perf_counter()
+        latency = end_time - start_time
+        print(f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f}')
 
-    val_results = eval(
-        val_loader,
-        val_snapshots_loader,
+    with hm.activate('val'):
+        val_results = eval(
+            val_loader,
+            val_snapshots_loader,
+            static_node_feats,
+            z,
+            encoder,
+            decoder,
+            eval_metric,
+            evaluator,
+            conversion_rate,
+        )
+        print(
+            f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
+            + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
+        )
+
+
+with hm.activate('test'):
+    test_results = eval(
+        test_loader,
+        test_snapshots_loader,
         static_node_feats,
         z,
         encoder,
@@ -319,20 +343,4 @@ for epoch in range(1, args.epochs + 1):
         evaluator,
         conversion_rate,
     )
-    print(
-        f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
-        + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
-    )
-
-test_results = eval(
-    test_loader,
-    test_snapshots_loader,
-    static_node_feats,
-    z,
-    encoder,
-    decoder,
-    eval_metric,
-    evaluator,
-    conversion_rate,
-)
-print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
+    print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
--- a/examples/linkproppred/TGB/gcn.py
+++ b/examples/linkproppred/TGB/gcn.py
@@ -254,7 +254,7 @@ train_neg_hook = NegativeEdgeSamplerHook(low=int(dst.min()), high=int(dst.max())
 val_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
 test_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')
 
-hm = HookManager(args.device)
+hm = HookManager(keys=['train', 'val', 'test'])
 hm.register('train', train_neg_hook)
 hm.register('val', val_neg_hook)
 hm.register('test', test_neg_hook)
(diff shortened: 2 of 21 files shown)


examples/linkproppred/TGB/tgat.py
```diff
@@ -232,37 +233,34 @@ def eval(
...
+train_neg_hook = NegativeEdgeSamplerHook(low=int(dst.min()), high=int(dst.max()))
+val_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
+test_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')
+
+hm = HookManager()
+hm.register('train', train_neg_hook)
+hm.register('val', val_neg_hook)
+hm.register('test', test_neg_hook)
+hm.register_shared(nbr_hook)
+hm.register_shared(DeviceTransferHook(args.device))
```
Jacob-Chmura: This was previously being done _under the hood_. But, this required taking a `DGraph` as input ( to derivce the right device).

I like doing it implicitly, and maybe, we can get away with it. But for now, the new api requires us to explicitly add the device hook (dedup can be added implicitly since there is not __init__ method).

tgm/hooks/hook_manager.py
```diff
@@ -0,0 +1,120 @@
...
+            raise TypeError(
+                f'Cannot register hook {type(hook).__name__}: must implement __call__(dg: DGraph, batch: DGBatch) -> DGBatch and reset_state()'
+            )
+
+    @staticmethod
+    def _topological_sort_hooks(hooks: List[DGHook]) -> List[DGHook]:
+        # Build adjacency list and then run Kahn's algorithmk jbs
+        adj_list: Dict[DGHook, List[DGHook]] = defaultdict(list)
+        for i, h1 in enumerate(hooks):
+            for j, h2 in enumerate(hooks):
```
Jacob-Chmura, line 141: The quadratic loop is fine becuase
1. The number of hooks is extremely small
2. This is not the hot path (we require users to register hooks outside of the activation loop, and we cache the right ordering at registry time)

tgm/loader.py
```diff
@@ -96,7 +96,7 @@ def __init__(
 
         self._dg = dg
         self._batch_size = batch_size
-        self._hook = HookManager.from_any(dg, hook)
+        self._hook_manager = HookManager() if hook_manager is None else hook_manager
```
Jacob-Chmura: Maybe make it required? I left it like this since maybe user just wants to iterate the data without doing any hook registration at all.

examples/linkproppred/TGB/tgat.py
shenyangHuang: Or we can just initialize the device on the `HookManager` constructor since a hook manager always need the knowledge of device.

shenyangHuang: we always need `DeviceTransferHook` right

examples/linkproppred/TGB/tgat.py
```diff
@@ -278,32 +276,27 @@ def _init_hooks(
...
-    )
-    start_time = time.perf_counter()
-    loss = train(train_loader, static_node_feats, encoder, decoder, opt)
-    end_time = time.perf_counter()
-    latency = end_time - start_time
-
-    val_results = eval(
-        val_loader, static_node_feats, encoder, decoder, eval_metric, evaluator
-    )
+    with hm.activate('train'):
```
shenyangHuang, line 277: yeah, this should be intuitive for people to learn

examples/linkproppred/TGB/tgat.py
```diff
@@ -278,32 +276,27 @@ def _init_hooks(
...
     print(
         f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
         + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
     )
 
-test_results = eval(
-    test_loader, static_node_feats, encoder, decoder, eval_metric, evaluator
-)
+    if epoch < args.epochs:  # Reset hooks after each epoch, except last epoch
+        hm.reset_state()
```
shenyangHuang, line 294: need to do this in the tutorial

shenyangHuang: added a comment in https://github.com/tgm-team/tgm/issues/106

tgm/hooks/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .base import DGHook, StatelessHook, StatefulHook
```
shenyangHuang, line 1: Note to later on break up `hooks.py`, intuitively we can break down to StatelessHook and StatefulHook

tgm/hooks/hook_manager.py
shenyangHuang: yeah looks good. Realistically, there should be at most 5-6 hooks.

tgm/hooks/hook_manager.py
```diff
@@ -0,0 +1,120 @@
...
+            raise RuntimeError(
+                'Cannot register hooks while a key is active. Register hooks before using `activate`.'
+            )
+
+        if key not in self._key_to_hooks:
+            self._key_to_hooks[key] = []
+        self._key_to_hooks[key].append(hook)
+
+        # Precompute execution order for the key
+        self._key_to_hooks[key] = self._topological_sort_hooks(
```
shenyangHuang: we run this every time we register a hook, again still fast so no issue.
Potentially, we can have an interface to allow users to register multiple hooks at once (not sure).

tgm/hooks/hook_manager.py
```diff
@@ -0,0 +1,120 @@
...
+
+        while queue:
+            u = queue.popleft()
+            ordered.append(u)
+            for v in adj_list.get(u, []):
+                indegree[v] -= 1
+                if indegree[v] == 0:
+                    queue.append(v)
+
+        if len(ordered) != len(hooks):
```
shenyangHuang, line 162: can we say explicitly we are missing xxx that we are producing?

tgm/loader.py
shenyangHuang: Yes, it is possible, we should teach them to make one. But for non ML use case, they might just want to loop over data.

examples/linkproppred/TGB/tgat.py
Jacob-Chmura: Yeah we can do this.

tgm/hooks/hook_manager.py
Jacob-Chmura: Yeah I was thinking this too. Lets leave it for full release once we are confident in this design.

But i agree, a helper wrapper would be to allow direct setting hooks with key value pairs at once

tgm/hooks/hook_manager.py
Jacob-Chmura: Yes, will do

examples/linkproppred/TGB/gcn.py
```diff
@@ -245,30 +248,31 @@ def eval(
...
 test_snapshots = DGraph(test_data_discretized, device=args.device)
 
-train_loader = DGDataLoader(
-    train_dg,
-    hook=[NegativeEdgeSamplerHook(low=0, high=train_dg.num_nodes)],
-    batch_size=args.bsize,
-)
+
+_, dst, _ = train_dg.edges
+train_neg_hook = NegativeEdgeSamplerHook(low=int(dst.min()), high=int(dst.max()))
```
Jacob-Chmura, line 253: Changed the sample range to match other examples.

examples/linkproppred/graphmixer.py
```diff
@@ -249,86 +250,93 @@ def eval(
...
-train_loader = DGDataLoader(
-    train_dg, hook=_init_hooks(train_dg, args.time_gap), batch_size=args.bsize
-)
-val_loader = DGDataLoader(
-    val_dg, hook=_init_hooks(val_dg, args.time_gap), batch_size=args.bsize
+graph_mixer_hook = GraphMixerHook(args.time_gap, num_nodes=test_dg.num_nodes)
+
+_, train_dst, _ = train_dg.edges
+_, val_dst, _ = val_dg.edges
+_, test_dst, _ = test_dg.edges
```
Jacob-Chmura, line 320: The examples are messy. We have an open issue before beta release to clean them all up.

tgm/hooks/hook_manager.py
```diff
@@ -0,0 +1,171 @@
...
+
+        # Implicitly add deduplication shared hook
+        self._shared_hooks.append(DeduplicationHook())
+
+        # Device transfer hook is a special case. It is always run first,
+        # and triggers even when no _active_key is present. This happens
+        # if a user want to run data loader on gpu, without any addition hooks (e.g. negatives, neighbors)
+        self._device_hook = DeviceTransferHook(device)
+
+    def __str__(self) -> str:
```
Jacob-Chmura, line 24: Gives us something like:

```
HookManager:
  DeviceHook: DeviceTransferHook (requires=set(), produces=set())
  Shared hooks:
    - DeduplicationHook (requires=set(), produces={'unique_nids', 'global_to_local'})
    - MockHook (requires=set(), produces=set())
  Active key: None
  Keyed hooks:
    train:
      - DeduplicationHook (requires=set(), produces={'unique_nids', 'global_to_local'})
      - MockHook (requires=set(), produces=set())
      - MockHookRequires (requires={'foo'}, produces=set())
      - MockHookWithState (requires=set(), produces=set())
    val:
      - DeduplicationHook (requires=set(), produces={'unique_nids', 'global_to_local'})
      - MockHook (requires=set(), produces=set())
      - MockHookRequires (requires={'foo'}, produces=set())

```

test/unit/test_hooks/test_hook_manager.py
```diff
@@ -57,89 +49,212 @@ def dg():
...
 
-def test_hook_manager_from_any_none(dg):
-    hook = HookManager.from_any(dg, None)
-    assert len(hook.hooks) == 1
-    assert isinstance(hook.hooks[0], DeduplicationHook)
+@pytest.mark.skip(
+    'TODO: This test only makes sense if we enable registering multiple hooks at once. '
+    'Otherwise, the "missing produced" will trigger an exception before the secon hook can '
+    'be registered. Skiping for now, and should reconsider enabling multiple hook registry.'
+)
```
Jacob-Chmura: This is important to note. We should think whether we want to support multi-hook registration.

Otherwise, the user ends up having to worry about the order themselves.

Alternatively, we don't order the hooks eagerly. Instead, we keep letting users add hooks. Once they are done, we do the topological sort on the first execution. We can cache this so that on subsequent batch iteration, we know the right order to execute.

Jacob-Chmura:
Currently the topo sort is not really useful. This is because (1) we only enable registering hooks one at a time, and (2) we eagerly run the topological sort.

Because of this, the user will immediately get the 'missing produces' exception before topological sort even runs. Hence, they end up having to be aware of the order that they register hooks.

**Option 1**: Leave this as is. We fail fast so that as soon as a hook is registered, all the requires must already be satisfied. This is simplest, but hardest on the user.

**Option 2**: Leave as is, but add ability to register multiple hooks at once. This way, we stil fail fast, but now we actual have a reason to do the sort: the user can input multiple hooks at once in any order, and we find the right order.

**Option 3**: Don't do the topological sort eagerly. Instead, let users add hooks as they wish. Defer the topological check until we actually active / execute the hooks. We can still cache the result of the topological sort, so that it's only executed once per key. (assumign the user does not re-registry a hook, but we already structured the api such that this is not encouraged).
Let's discuss this tomorrow, otherwise, the unit tests, and exampels have been updated, and pending this detail we can merge.
My preference is either (2) or (3). If we do (3), we can add (2) as an additional features down the line. Doign only (1) and leaving (2) and or (3) for full release is valid too, but requires user to think about which order to register hooks and shared hooks.

I added a string method that nicely shows all the registered hooks and their dependency resolution order, but it's still better in my opinion if user can just blindly register hooks and the manager handles the execution / ordering.

examples/linkproppred/TGB/tgn.py
```diff
@@ -477,35 +474,29 @@ def _init_hooks(
...
-        f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
-        + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
-    )
+    with hm.activate('train'):
+        start_time = time.perf_counter()
+        loss = train(train_loader, static_node_feats, encoder, decoder, opt)
+        end_time = time.perf_counter()
+        latency = end_time - start_time
+
+        with hm.activaet('val'):
```
ntgbaoo: `with hm.activate('val'):`

examples/linkproppred/TGB/edgebank.py
```diff
@@ -95,7 +94,10 @@ def eval(
...
-print(' '.join(f'{k}={v:.4f}' for k, v in val_results.items()))
-test_results = eval(test_loader, model, eval_metric, evaluator)
-print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
+with hm.activate('val'):
+    val_results = eval(val_loader, model, eval_metric, evaluator)
+    print(' '.join(f'{k}={v:.4f}' for k, v in val_results.items()))
+
+with hm.activate('test'):
+    test_results = eval(test_loader, model, eval_metric, evaluator)
+    print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
```
ntgbaoo, line 103: Without reading your detailed description in the PR, it is not very intuitive for the users to understand which hook(s) are used when `.active()` is called. Maybe it's good to give a brief explanation here. Would definitely need a good tutorial for this.

examples/linkproppred/TGB/tgat.py
```diff
@@ -278,32 +274,27 @@ def _init_hooks(
...
         f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
         + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
     )
 
-test_results = eval(
-    test_loader, static_node_feats, encoder, decoder, eval_metric, evaluator
-)
-print(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
+    if epoch < args.epochs:  # Reset hooks after each epoch, except last epoch
+        hm.reset_state()
```
ntgbaoo, line 294: So we reset all the hooks belonging to `hm`. Can we take the key as the input, and only reset hooks registered with the given key?
I am just thinking about the MiNT use case when we train with multiple datasets at each epoch. In other words, I will loop over all datasets in a single epoch. Sometimes I may want to reset hooks registered with `train`, sometimes I don't want to because I think the transfer information may be helpful. Hence, only hooks register with `val` needs to be reset.

tgm/hooks/hook_manager.py
ntgbaoo: I liked this. Very helpful for users to understand and debug

benjaminnNgo (commented):
Left minor comments. In general, I like the idea. Although it is not very intuitive for users at the first glance to see when which hook(s) is used when we call `.active(key)`, we can make this clear if we have a good tutorial. And, the concept of using `with hm.active()` is very easy to get and use.

Regarding topo sort, I prefer option 3 as well. However, if we have a tight deadline for beta release, I would say leave it as it is and let users think about it and we come back and deal with it for the full release.

examples/linkproppred/TGB/tgat.py
Jacob-Chmura: Yes, you can reset by key. This will register shared hooks as well, since, they are the same object across keys.

examples/linkproppred/TGB/edgebank.py
shenyangHuang: yes, we need to add a detailed tutorial, see issue https://github.com/tgm-team/tgm/issues/106

examples/linkproppred/TGB/edgebank.py
```diff
@@ -75,16 +75,15 @@ def eval(
...
-test_loader = DGDataLoader(
-    test_dg,
-    hook=[TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')],
-    batch_size=args.bsize,
-)
+val_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
+test_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')
+
+hm = HookManager(keys=['val', 'test'])
+hm.register('val', val_neg_hook)
```
shenyangHuang, line 82: why not just initialize the TGB hook here.

examples/linkproppred/TGB/gcn.py
```diff
@@ -245,30 +248,31 @@ def eval(
...
+val_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='val')
+test_neg_hook = TGBNegativeEdgeSamplerHook(neg_sampler, split_mode='test')
+
+hm = HookManager(keys=['train', 'val', 'test'])
+hm.register('train', train_neg_hook)
```
shenyangHuang, line 258: maybe just initialize here, 'train' key is implicitly to mean this will be for training

examples/linkproppred/TGB/gcn.py
```diff
@@ -294,23 +298,43 @@ def eval(
...
-        static_node_feats,
-        encoder,
-        decoder,
-        opt,
-        conversion_rate,
-    )
-    end_time = time.perf_counter()
-    latency = end_time - start_time
-    print(f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f}')
+    with hm.activate('train'):
```
shenyangHuang, line 301: I was thinking, we activate this in the train function? what do you think? in examples, we always ask user to feed hm to `train` and `eval` functions to let them remember to scope it correctly,

examples/linkproppred/TGB/tgn.py
```diff
@@ -477,35 +474,29 @@ def _init_hooks(
...
-        f'Epoch={epoch:02d} Latency={latency:.4f} Loss={loss:.4f} '
-        + ' '.join(f'{k}={v:.4f}' for k, v in val_results.items())
-    )
+    with hm.activate('train'):
+        start_time = time.perf_counter()
+        loss = train(train_loader, static_node_feats, encoder, decoder, opt)
+        end_time = time.perf_counter()
+        latency = end_time - start_time
+
+        with hm.activate('val'):
```
shenyangHuang, line 483: is the indent off here?

tgm/hooks/hook_manager.py
shenyangHuang: yeah looks good

shenyangHuang (approved):

Issue #151, opened 2025-08-18
TPNet Model


Jacob-Chmura:
https://github.com/lxd99/TGB_TPNet

PR #180, opened 2025-08-26
151 - TPNet implementation

benjaminnNgo:
## Objective

Add TPNet to the library

## Testing evidence

Epoch 1 validation MRR: `0.8180` (Thanks, Andy, for helping me run it)

AUC and AP results
tpnet_log

## Checklists

- [x] Added skeleton for: `RandomProjectionModule`, `FeedForwardNet`, `MLPMixer`, `TPNet`
- [x] Implement  `RandomProjectionModule`,
- [x] Implement `FeedForwardNet`, `MLPMixer`
- [x] Implement `TPNet`
- [x] Prepare link prediction example
- [x] Add the model to the integration tests

**Relevant PR:** https://github.com/tgm-team/tgm/issues/151

### DIFF
--- a/tgm/nn/model/tpnet.py
+++ b/tgm/nn/model/tpnet.py
@@ -220,9 +220,7 @@ class RandomProjectionModule(nn.Module):
 
         Returns:
         """
-        assert (
-            len(random_projections) == 2
-        ), (
+        assert len(random_projections) == 2, (
             'Expect a tuple of (now_time,random_projections)'
         )  # @TODO: Need to raise custom exception
         now_time, random_projections = random_projections
@@ -234,9 +232,7 @@ class RandomProjectionModule(nn.Module):
 
         self.now_time.data = now_time.clone()
         for i in range(1, self.num_layer + 1):
-            assert torch.is_tensor(
-                random_projections[i - 1]
-            ), (
+            assert torch.is_tensor(random_projections[i - 1]), (
                 'Not a valid state of random projection'
             )  # @TODO: Need to raise custom exception
             self.random_projections[i].data = random_projections[i - 1].clone()


examples/linkproppred/tpnet.py
```diff
@@ -0,0 +1,342 @@
...
+            dropout=dropout,
+            num_layers=num_layers,
+            num_neighbors=num_neighbors,
+            random_projections=random_projection_module,
+            device=device,
+            time_encoder=time_encoder,
+        )
+        self.rp_module = random_projection_module.to(device)
+        self.decoder = LinkPredictor(output_dim).to(
+            device
```
ntgbaoo, line 115: This is something we need to think about. Unlike other methods, which return embedding of all nodes for every forward pass, TPNet and DyGFormer return node embeddings given edges. If we make this LinkPredictor as a separate module, we need to return `z_src_pos`, `z_dst_pos`, `z_src_neg` and `z_dst_neg`

examples/linkproppred/tpnet.py
```diff
@@ -0,0 +1,342 @@
...
+random_projection_module = RandomProjectionModule(
+    num_nodes=num_nodes,
+    num_layer=args.rp_num_layers,
+    time_decay_weight=args.rp_time_decay_weight,
+    beginning_time=dgraph.start_time,
+    enforce_dim=args.enforce_dim,
+    num_edges=train_dg.num_edges,
+    dim_factor=args.rp_dim_factor,
+    device=args.device,
+)
```
ntgbaoo, line 311: I need to define this module explicitly, as I need to reset its state at the end of each epoch. Another option for this is: define this module implicitly inside `TPNet` and make `TPNet` have APIs call to reset states of this module. But one drawback of this is we need to bring all input args to init `RandomProjectionModule` to `TPNet` init agrs, which makes a giant block of init input args for `TPNet`.

test/integration/test_tpnet.py
```diff
@@ -0,0 +1,23 @@
...
+        '--mem=8G',
+        '--time=3:00:00',
+        '--gres=gpu:a100l:1',
+    ]
+)
+def test_tpnet_linkprop_pred(slurm_job_runner, dataset):
+    cmd = f"""
+python "$ROOT_DIR/examples/linkproppred/tpnet.py" \
+    --dataset {dataset} \
+    --device cuda \
```
ntgbaoo, line 19: As Andy mentioned @Jacob-Chmura, this integration test need to be run on A100 with 80GiB GPU memory.

tgm/nn/model/tpnet.py
```diff
@@ -0,0 +1,489 @@
...
+        Args:
+            X (PyTorch Float Tensor): Input tensor
+
+        Returns:
+            H : Output tensor
+        """
+        return self.ffn(X)
+
+
+class MLPMixer(nn.Module):
```
ntgbaoo, line 282: We have this defined in graphmixer example. Should we make this in `graphmixer.py` inside `tgm.model.nn`

test/integration/test_tpnet.py
Jacob-Chmura: Sounds good, currently all our integration tests run on 80gig cards anyway, since mila has more of them.

examples/linkproppred/tpnet.py
```diff
@@ -0,0 +1,342 @@
...
+parser = argparse.ArgumentParser(
+    description='TPNet LinkPropPred Example',
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+)
+parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
+parser.add_argument('--dataset', type=str, default='tgbl-wiki', help='Dataset name')
+parser.add_argument('--bsize', type=int, default=200, help='batch size')
+parser.add_argument('--device', type=str, default='cpu', help='torch device')
+parser.add_argument(
+    '--num-neighbors',
```
Jacob-Chmura, line 34: Nit, but for consistency across examples
```suggestion
    '--n-nbrs',
```

examples/linkproppred/tpnet.py
```diff
@@ -0,0 +1,342 @@
...
+)
+parser.add_argument('--node-dim', type=int, default=128, help='embedding dimension')
+parser.add_argument('--time-dim', type=int, default=100, help='time encoding dimension')
+parser.add_argument(
+    '--embed-dim', type=int, default=172, help='node representation dimension'
+)
+parser.add_argument('--num-layers', type=int, default=2, help='number of model layers')
+parser.add_argument('--dropout', type=float, default=0.1, help='dropout rate')
+parser.add_argument('--lr', type=float, default=0.0001, help='learning rate')
+parser.add_argument('--epochs', type=int, default=100, help='number of epochs')
```
Jacob-Chmura, line 71:
```suggestion
parser.add_argument('--epochs', type=int, default=10, help='number of epochs')
```

examples/linkproppred/tpnet.py
```diff
@@ -0,0 +1,342 @@
...
+            }
+            perf_list.append(evaluator.eval(input_dict)[eval_metric])
+
+    return float(np.mean(perf_list))
+
+
+args = parser.parse_args()
+seed_everything(args.seed)
+
+# loading negative sample from TGB
```
Jacob-Chmura, line 260: Nit
```suggestion
```

examples/linkproppred/tpnet.py
```diff
@@ -0,0 +1,342 @@
...
+data = DGData.from_tgb(args.dataset)
+dgraph = DGraph(data)
+
+num_nodes = dgraph.num_nodes
+edge_feats_dim = dgraph.edge_feats_dim
+
+if dgraph.static_node_feats is not None:
+    STATIC_NODE_FEAT = dgraph.static_node_feats
+else:
+    STATIC_NODE_FEAT = torch.randn((num_nodes, args.node_dim), device=args.device)
```
Jacob-Chmura: Can we pass static node feats along to train and eval as in the updated examples, instead of the global variable

tgm/nn/model/tpnet.py
Jacob-Chmura: Thanks for catching it, let's leave for post-iclr #162

tgm/nn/model/tpnet.py
```diff
@@ -0,0 +1,485 @@
...
+        Z_l_tilde = X + H_l_tilde
+
+        H_l = self.channel_norm(Z_l_tilde)
+        H_l = self.channel_feedforward(H_l)
+        Z_l = Z_l_tilde + H_l
+        return Z_l
+
+
+class TPNet(nn.Module):
+    r"""An implementation of DyGFormer.
```
Jacob-Chmura:
```suggestion
    r"""An implementation of TPNet.
```

examples/linkproppred/tpnet.py
Jacob-Chmura: I see, and it's not possible to derive this in a single forward pass? I don't suggest we change this now, but potentially for Beta, it seemse like this and Dyglib are a slightly different class of models, _link-centric_ you can say.

Intuitively, it appears that you could just reconstruct the edge index via:

```python
pos_edges = torch.cat([batch.src, batch.dst])
neg_edges = torch.cat([batch.src, batch.neg])
edge_index = torch.cat([pos_edges, neg_edges])

z_all = self.encoder(edge_index, ...)
z_src, z_dst, z_neg = z_all.chunk(..) # Something here I'm not sure about
pos_out = decoder(z_src, z_dst)
neg_out = decoder(z_src, z_neg)
````

But it's very likely I'm missing something. We can revisit these two models after iclr and see if we can unify them. Ultimately, we will just want some boundary that says, anything in `tgm.nn` accepts `DGBatch` and gives you the right output. As long as we follow that contract, the rest doesn't matter to much.

File: tgm/nn/model/tpnet.py [hunk 10]
```diff
@@ -0,0 +1,485 @@
...
+                dim=2,
+            )
+        else:
+            neighbor_combine_features = torch.cat(
+                [neighbor_node_features, neighbours_time_feats, neighbours_edge_feat],
+                dim=2,
+            )
+
+        embeddings = self.projection_layer(neighbor_combine_features)
+        embeddings.masked_fill((neighbours == 0)[:, :, None].to(self.device), 0)
```
Jacob-Chmura: Is this meant to be:
```python
embeddings.masked_fill((neighbours == PADDED_NODE_ID)[:, :, None].to(self.device), 0)
```
?

examples/linkproppred/tpnet.py
Jacob-Chmura: Understood. Makes me think we should create something like a model reset hook, then define a interface where that allows all models to get automatically have a `reset` method called at end of epoch. Future work I suppose

Jacob-Chmura (changes requested):
This looks great, don't want to change to much given we know it works, for iclr.

Left some minor comments. Can we also:
- Update readme to check off TPNET
- Update docs: https://github.com/tgm-team/tgm/blob/main/docs/api/nn.md , just add the relevant model class in the same format (I forgot to flag this for dygformer, we can add that as well). You can build the docs and ensure they don't break by running:

```sh
./scripts/build_docs.sh
```


Also, do you have a sense of where the 80gb requirement comes from? Is it the random projection model? What's eating up gpu?

examples/linkproppred/tpnet.py
ntgbaoo: This is a great idea, actually. And this should be the right way of doing a forward pass. But I believe that doing forward pass in this way could slightly change model performance. Recall the observation that changing the evaluation batch size affects model performance. Seeing an edge in a forward pass may allow the model to update its state and use that for subsequent predictions, resulting in better performance.

This makes me think that current link-centric models like DyGFormer, TPNet, etc. take the advantages of positive edges information to predict negative edges, which is practically not true as you don't know whether an edge is positive or negative. @shenyangHuang, @Jacob-Chmura what do you guys think?

This is just my hypothesis, and it may be interesting to clarify after ICLR submission.

examples/linkproppred/tpnet.py
ntgbaoo: Yeah. I like the idea. The hook will take model as init input args and taking care of reseting model's state at the beginning of each epoch. By doing that we don't need to do [this](https://github.com/tgm-team/tgm/blob/151-TPNet-Implementation/examples/linkproppred/tpnet.py#L338). But it won't solve the issue of floating module like `RandomProjectionModule`. Need to think about it more.

tgm/nn/model/tpnet.py
ntgbaoo: Took a note of this.

tgm/nn/model/tpnet.py
ntgbaoo: Oh good catch. Should be `PADDED_NODE_ID`

Issue #181, opened 2025-08-26
Custom Exceptions


Jacob-Chmura:
`HookDepedencyError`
`PaddedNodedIDError`

and more

PR #187, opened 2025-08-28
Dev/custom exceptions

Jacob-Chmura:
### Summary / Description

The purpose of this PR is to add custom excetion classes into our codebase. This provides clearer semantics and intent for the user. It also sets us up for forward compatibility: users can just catch our custom exceptions for TGM specific ops (e.g. time granularity errors).

**Related Issues:** #181 

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence
- [x] Unit tests

#### Note
This is stacked on top of #179

### DIFF
--- a/tgm/exceptions.py
+++ b/tgm/exceptions.py
@@ -1,0 +1,31 @@
+class BadHookProtocolError(Exception):
+    """Raised when a DGHook does not correctly implement the required protocol for execution by the HookManager."""
+
+
+class UnresolvableHookDependenciesError(Exception):
+    """Raised when no valid execution ordering of hooks can be found, due to conflicting or cyclic requires/produces dependencies."""
+
+
+class PaddedNodeIDError(Exception):
+    """Raised when a dataset contains node IDs that conflict with the reserved padded node placeholder ID."""
+
+
+class EmptyGraphError(Exception):
+    """Raised when attempting to instantiate an empty graph. Empty graphs are unsupported since updates are not allowed."""
+
+
+class OrderedTimeGranularityError(Exception):
+    """Raised when an operation requiring time-based granularity is attempted on a graph that only has ordered (relative) granularity.
+
+    Examples:
+        - Discretizing or coarsening an ordered graph by absolute time units.
+        - Iterating an ordered graph with non-ordered batch units (e.g. daily snapshots).
+    """
+
+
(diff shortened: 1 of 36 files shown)


tgm/exceptions.py
```diff
@@ -0,0 +1,35 @@
...
+
+class BadHookProtocolError(TGMError):
+    """Raised when a DGHook does not correctly implement the required protocol for execution by the HookManager."""
+
+
+class UnresolvableHookDependenciesError(TGMError):
+    """Raised when no valid execution ordering of hooks can be found, due to conflicting or cyclic requires/produces dependencies."""
+
+
+class PaddedNodeIDError(TGMError):
```
shenyangHuang: maybe make the comment more explicitly here, point to where the paddedID is located.

tgm/exceptions.py
```diff
@@ -0,0 +1,35 @@
...
+    """Raised when a dataset contains node IDs that conflict with the reserved padded node placeholder ID."""
+
+
+class EmptyGraphError(TGMError):
+    """Raised when attempting to instantiate an empty graph. Empty graphs are unsupported since updates are not allowed."""
+
+
+class OrderedTimeGranularityError(TGMError):
+    """Raised when an operation requiring time-based granularity is attempted on a graph that only has ordered (relative) granularity.
```
shenyangHuang, line 22: the name of the error is not very intuitive to me, it only tells you sth is wrong with ordered time granularity but not why it is caused (due to your conversation of timegran)

tgm/loader.py
```diff
@@ -79,12 +84,14 @@ def __init__(
         batch_ordered = batch_unit == 'r'
 
         if dg_ordered and not batch_ordered:
-            raise ValueError('Cannot iterate ordered dg using non-ordered batch_unit')
+            raise OrderedTimeGranularityError(
```
shenyangHuang: yeah the error should be called sth else, maybe like time conversion error?  not sure yet.

tgm/exceptions.py
shenyangHuang: also maybe just call it Invalid NodeID error?

shenyangHuang (changes requested):

Issue #191, opened 2025-08-29
Topological Sort may make neighbor sampler preceed negatives


Jacob-Chmura:
**Describe the bug**
Since neighbor sampler does not explicitly require negative, the topological sort may put the neighbor sampler first.

But if we *are* using negatives, then it's crucial that they come before the neighbor sampling.

PR #192, opened 2025-08-29
Fix Implicit Neighbor Hook Precedence

Jacob-Chmura:
### Summary / Description

Since neighbor sampler does not explicitly require negative, the topological sort may put the neighbor sampler first.
But if we *are* using negatives, then it's crucial that they come before the neighbor sampling.

To fix this, I monkey patched the topological sort by creating an edge in the DAG between negative and neighbor hooks if they both exist in our registry.

**Related Issues:** #191 

#### Type of Change

- [X] Bug fix

#### Test Evidence

- [X] Unit tests

### DIFF
--- a/tgm/hooks/hook_manager.py
+++ b/tgm/hooks/hook_manager.py
@@ -148,7 +148,7 @@ class HookManager:
                 # that the negatives come first (so that we sample neighbors for the negatives).
                 # But since neighbor sampler does not explicitly require negatives, the topological
                 # sort may put these out of order. In order to fix this, we add an extra edge
-                # into tthe DAG before sorting. Long term, we need to think about how to avoid
+                # into the DAG before sorting. Long term, we need to think about how to avoid
                 # things like this, and make it seamless for the user.
                 is_neg_hook = lambda h: 'neg' in h.produces
                 is_nbr_hook = lambda h: 'nbr_nids' in h.produces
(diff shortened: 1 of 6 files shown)


tgm/hooks/hook_manager.py
```diff
@@ -135,14 +135,26 @@ def _topological_sort_hooks(hooks: List[DGHook]) -> List[DGHook]:
...
                 if i != j and h1.produces & h2.requires:
                     # If h2 requires something h1 produces, add edge
                     adj_list[h1].append(h2)
 
+                # TODO: This is a hacky short term fix for implcit hook ordering constraints.
+                # If both a negative hook and a neighbor hook are present, it is crucial
+                # that the negatives come first (so that we sample neighbors for the negatives).
+                # But since neighbor sampler does not explicitly require negatives, the topological
+                # sort may put these out of order. In order to fix this, we add an extra edge
+                # into the DAG before sorting. Long term, we need to think about how to avoid
```
shenyangHuang, line 151: yes, I agree.

shenyangHuang: potential solution is to know which set of hook classes will not depend on other things and will always be at top such as negative sampler, it never makes sense to sample negative hooks last so maybe the sorting don't happen for these kind of hooks and they always stay on top

shenyangHuang: to address later when we add more hooks.

shenyangHuang (approved):

Issue #193, opened 2025-08-31
Default key for HookManager


benjaminnNgo:
**Describe the bug**
Currently, examples nodepredprop failed due to RuntimeError: No active key set. Use activate() context manager.

**Screenshots**

<img width="789" height="308" alt="Image" src="https://github.com/user-attachments/assets/5b0f6496-bc36-458d-98b7-5b3fc332f8c1" />

Jacob-Chmura:
My mistake, forgot to remove the hookmanager altogether from these examples.

PR #194, opened 2025-08-31
Made shared_hooks as default key in hook managers

benjaminnNgo:
### Summary / Description

Make `_shared_hooks` the default key in `HookManager` to fix nodeproppred examples, which don't need to register any hooks.

**Related Issues:** # (issue number)

#### Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence

- [X] Unit tests
- [ ] Integration tests
- [ ] Performance tests

*Related PR:* https://github.com/tgm-team/tgm/issues/193

### DIFF
--- a/tgm/hooks/hook_manager.py
+++ b/tgm/hooks/hook_manager.py
@@ -5,6 +5,7 @@ from contextlib import contextmanager
 from typing import Any, Dict, Iterator, List
 
 from tgm import DGBatch, DGraph
+from tgm.constants import DEFAULT_KEY_HOOK_MANAGER
 from tgm.exceptions import (
     BadHookProtocolError,
     UnresolvableHookDependenciesError,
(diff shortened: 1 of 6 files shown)


tgm/hooks/hook_manager.py
```diff
@@ -123,8 +122,18 @@ def execute_active_hooks(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
         if self._active_key is None:
-            raise RuntimeError('No active key set. Use activate() context manager.')
+            registered_key = [
+                key
+                for key in self._key_to_hooks.keys()
+                if key != DEFAULT_KEY_HOOK_MANAGER
+            ]
+            raise RuntimeError(
+                f'No active key set. Use activate() on following registered context {registered_key} from context manager.'
+            )
```
ntgbaoo, line 136: To be discussed here, this will raise when a user registers a hook (different from the default ones) and forgets to activate it. This may limit some use cases, such as in some parts of the code, users may want to not activate any hooks they registered.

Jacob-Chmura:
Thanks Bao. But actually, the node prop examples don't really need any hooks. I have a branch which addresses this, and will push shortly. I don't think we'll need the changes here. Deduplication is not actually something that we need for now, but it may be something we integrate before release. Of coarse, the user can always explicitly register it.

PR #195, opened 2025-08-31
Fix Examples in CI

Jacob-Chmura:


.github/scripts/run_download_tgb_job_sbatch.sh
Jacob-Chmura, line 1: The problem was that `downlaod_tgb_datasets.sh` required a good amount of RAM when the processed data was not actually cached. So we need to actually allocate resources. 

Since the CI runner has a bare-minimall allocation (so that it can run 24/7), I need to `sbatch` a new job to check tgb data.

Technically, the "check if I need to do anything" _can_ be done in the ci runner, and we can skip the entire sbatch allocation in the typical case were we already have the data...

.github/scripts/run_perf_job_sbatch.sh
```diff
@@ -14,8 +12,13 @@ module load cudatoolkit/11.7
 
 cd "$GITHUB_WORKSPACE"
 
-export TGM_CI_PERF_LOG_BASE=${TGM_CI_PERF_LOG_BASE:-$HOME/tgm_ci_perf}
+export TGM_CI_PERF_LOG_BASE=${TGM_CI_PERF_LOG_BASE:-/home/$USER/tgm_ci_perf}
```
Jacob-Chmura: Found out the $HOME is set to a temporary local path on the github runner, not actual $HOME.

.github/workflows/integration.yml
```diff
@@ -20,7 +20,8 @@ jobs:
     if: github.actor == 'jacob-chmura' || github.actor == 'shenyanghuang'
     env:
       TGM_CI_LOG_BASE: $SLURM_TMPDIR/tgm_ci
-      DATA_ROOT: $HOME/tgb_datasets
+      REAL_HOME: /home/mila/j/jacob.chmura
```
Jacob-Chmura, line 22: This is not ideal to hardcode, I'm happy to hear better solutions. As I mentioned, $HOME is not expanded to real home in the github runner.

I suppose I can create a github secret to hide, this, but this seems more maintainable

.github/workflows/integration.yml
```diff
@@ -20,7 +20,8 @@ jobs:
     if: github.actor == 'jacob-chmura' || github.actor == 'shenyanghuang'
     env:
       TGM_CI_LOG_BASE: $SLURM_TMPDIR/tgm_ci
```
Jacob-Chmura: I'm afraid `SLURM_TMPDIR` also does not expand to the fast slurm dir, making our ci slightly slower than ideal... micro optimization to eventually fix

examples/linkproppred/TGB/gcn.py
```diff
@@ -33,7 +33,7 @@
 parser.add_argument('--dataset', type=str, default='tgbl-wiki', help='Dataset name')
 parser.add_argument('--device', type=str, default='cpu', help='torch device')
 parser.add_argument('--epochs', type=int, default=10, help='number of epochs')
-parser.add_argument('--lr', type=float, default=0.0001, help='learning rate')
+parser.add_argument('--lr', type=float, default=0.001, help='learning rate')
```
Jacob-Chmura, line 36: These are the learning rates I saw in the linked GCN code: https://github.com/shenyangHuang/UTG/blob/8b53ad894fbd1b1782367affffcc49b681fe4fd1/ctdg_utg_gcn.py#L3

examples/linkproppred/TGB/gcn.py
```diff
@@ -264,14 +262,10 @@ def eval(
 test_loader = DGDataLoader(test_dg, args.bsize, hook_manager=hm)
 
 train_snapshots_loader = DGDataLoader(
```
Jacob-Chmura, line 264: This caused an exception, but also, the snapshots don't need hooks.

examples/linkproppred/TGB/tgn.py
```diff
@@ -257,71 +258,70 @@ def get_updated_memory(
...
-                edge_feat=batch.nbr_feats[hop],
-                nbr_node_feat=nbr_feat,
-                nbr_time_feat=nbr_time_feat,
-                valid_nbr_mask=nbr_mask != PADDED_NODE_ID,
-            )
-            z[batch.global_to_local(seed_nodes)] = out
-        return z
+        z = {j: {} for j in range(self.num_layers + 1)}  # z[j][i] = z of nbr^i at hop j
+
+        # Layer 0 (leaf nodes): z[0][i] = static_node_feat
```
Jacob-Chmura, line 300: I ported over the tgat layer from our working example, and changed only the base case so that the base nodea features are the sum of the static node features and node memory. The example still does not work, but it runs now.

examples/nodeproppred/gclstm.py
```diff
@@ -173,10 +172,9 @@ def eval(
 label_dim = train_dg.dynamic_node_feats_dim
 evaluator = Evaluator(name=args.dataset)
 
-hm = HookManager(keys=['train', 'val', 'test'])
-train_loader = DGDataLoader(train_dg, batch_unit=args.batch_time_gran, hook_manager=hm)
-val_loader = DGDataLoader(val_dg, batch_unit=args.batch_time_gran, hook_manager=hm)
-test_loader = DGDataLoader(test_dg, batch_unit=args.batch_time_gran, hook_manager=hm)
+train_loader = DGDataLoader(train_dg, batch_unit=args.batch_time_gran)
```
Jacob-Chmura, line 175: No need for hooks, insteaf of global-to-local, I read embeddings direclty `z_node = z[batch.node_ids]` which seems to work

scripts/trigger_integration_tests.sh
Jacob-Chmura, line 1: I'm considering making a helper script called `trigger_workflow.sh` that abstracts out the common logic, since it's not clear from the name of the script that we can also run perf tests.

.github/workflows/integration.yml
shenyangHuang: I think this is fine for now but yeah we should find a way to hide it later on

examples/linkproppred/TGB/gcn.py
shenyangHuang: they vary depend on dataset but yeah either is fine for now. smaller learning rate is slower to train but leads to potentially better performance. larger learning rate is faster but the model might not learn as well.

examples/linkproppred/TGB/tgn.py
shenyangHuang: sounds good

examples/linkproppred/graphmixer.py
```diff
@@ -119,9 +119,7 @@ def forward(
 
         # Link Decoder
         z = self.output_layer(torch.cat([z_link, z_node], dim=1))
-        z_src = z[batch.global_to_local(batch.src)]
-        z_dst = z[batch.global_to_local(batch.dst)]
-        z_neg = z[batch.global_to_local(batch.neg)]
+        z_src, z_dst, z_neg = torch.chunk(z, 3)
```
shenyangHuang, line 122: ideally we should always teach users that it will be src,  dst and whatever is leftover is neg, instead of assuming neg will always be able to be chunked. It is fine now as it is simple.

examples/nodeproppred/gclstm.py
shenyangHuang: this is because snapshot-based models have the same size embedding size as the node size so it works. Not necessarily true for all models.

PR #197, opened 2025-09-01
Better Edgebank

Jacob-Chmura:
### Summary / Description

Cleaning up edgebank example.

- Perf: 78s validation loop down to 10s locally. Current bottlenecks are now tgb neg sampler and python loop in edgebank (would require switching to tensor memory which is a decent amount of work)
- Remove non-tgb edgebank from examples and integration tests
- Clean up example file
- Add integration test for fixed memory edge bank

### DIFF
--- a/examples/linkproppred/TGB/edgebank.py
+++ b/examples/linkproppred/TGB/edgebank.py
@@ -20,10 +20,10 @@ parser = argparse.ArgumentParser(
 parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
 parser.add_argument('--dataset', type=str, default='tgbl-wiki', help='Dataset name')
 parser.add_argument('--bsize', type=int, default=200, help='batch size')
-parser.add_argument('--window_ratio', type=float, default=0.15, help='Window ratio')
-parser.add_argument('--pos_prob', type=float, default=1.0, help='Positive edge prob')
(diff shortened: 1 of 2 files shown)


examples/linkproppred/TGB/edgebank.py
```diff
@@ -19,10 +20,10 @@
 parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
 parser.add_argument('--dataset', type=str, default='tgbl-wiki', help='Dataset name')
 parser.add_argument('--bsize', type=int, default=200, help='batch size')
-parser.add_argument('--window_ratio', type=float, default=0.15, help='Window ratio')
-parser.add_argument('--pos_prob', type=float, default=1.0, help='Positive edge prob')
+parser.add_argument('--window-ratio', type=float, default=0.15, help='Window ratio')
```
ntgbaoo, line 23: Any reasons why favor `window-ratio` over `window_ratio`? I personally like `window_ratio` because this matchs how we retrieve the arguments in code `args.window_ratio`, and we can't retrieve arguments by `args.window-ratio`

Jacob-Chmura: Both are fine, I think hyphenated is more traditionally used for cli tools. I mainly changed it since all our other examples use hypthens

PR #200, opened 2025-09-01
Better graph mixer

Jacob-Chmura:


examples/linkproppred/TGB/graphmixer.py
```diff
@@ -90,39 +95,26 @@ def __init__(
...
-        # Link Decoder
         z = self.output_layer(torch.cat([z_link, z_node], dim=1))
-        z_src, z_dst, z_neg = torch.chunk(z, 3)
-        pos_out = self.link_predictor(z_src, z_dst)
-        neg_out = self.link_predictor(z_src, z_neg)
-        return pos_out, neg_out
+        return z
 
 
 class FeedForwardNet(nn.Module):
```
ntgbaoo, line 120: In my TPNet implementation, I has 2 similar modules: `FeedForwardNet` and `MLPMixer`. Do you think it is a good idea to move them in `nn/models`?

examples/linkproppred/TGB/graphmixer.py
```diff
@@ -219,43 +220,54 @@ def train(
...
 
-# Neighbor Sampler and GraphMixer Hook is shared across loaders
-nbr_hook = RecencyNeighborHook(
-    num_nbrs=[args.n_nbrs],
-    num_nodes=test_dg.num_nodes,  # Assuming node ids at test set > train/val set
-    edge_feats_dim=test_dg.edge_feats_dim,
-)
-
-
 class GraphMixerHook(StatefulHook):
```
ntgbaoo, line 271: Not a big deal but can we move the hook definition before:
```
args = parser.parse_args()
```

benjaminnNgo (approved):
I can't validate the correctness of GraphMixerHook as I am not familiar with the technical details of the model. Other than that, everything looks good to me

examples/linkproppred/TGB/graphmixer.py
Jacob-Chmura: Yeah we could. I think post iclr we'll properly handle how we plan on maintaining and supporting common layers. So for now, leaving as is.

PR #206, opened 2025-09-02
Torch Geometric as Core Dependancy

Jacob-Chmura:
### Summary / Description

Adding pyg to our core dependencies, in light of our decision to host complete model implementation in our library #162.

We may consider keeping it as an optional dependency (only needed for gcn-style models). To be determined...


shenyangHuang:
I think the idea of optional dependency so user need to use it if they want to use certain models and maybe will prompt them to install pyg

PR #210, opened 2025-09-02
Deprecate non mrr linkprop

Jacob-Chmura:
### Summary / Description

Close #201


examples/linkproppred/tgat.py
Jacob-Chmura, line 1: This and the tgn diffs can be ignored. These are the previous _tgb_ versions of the files showing up here.

benjaminnNgo (changes requested):
[This integration test](https://github.com/tgm-team/tgm/blob/dev/deprecate_non_mrr_linkprop/test/integration/test_dygformer.py#L15C1-L17C60) needs to be updated as well, so that integration tests won't fail.

test/integration/test_dygformer.py
```diff
@@ -14,7 +14,7 @@
 )
 def test_dygformer_tgb_recency_sampler_linkprop_pred(slurm_job_runner, dataset):
```
ntgbaoo: May want to rename this to `test_dygformer_linkprop_pred_recency_sampler` to be consistent across tests.

benjaminnNgo (approved):
Left one more minor comment. Other than that, everything looks good to me. Approved.

PR #212, opened 2025-09-03
Remove Undirected Assumption

Jacob-Chmura:
### Summary / Description

Both samplers assume the graph is undirected. Removing the offending lines should make the samplers work for both directed and undirected edgelists.

**Related Issues:** #175

### DIFF
--- a/test/unit/test_hooks/test_neighbor_sampler_hook.py
+++ b/test/unit/test_hooks/test_neighbor_sampler_hook.py
@@ -195,10 +195,6 @@ def test_init_basic_sampled_graph_1_hop(basic_sample_graph):
 
 
 def test_init_basic_sampled_graph_directed_1_hop(basic_sample_graph):
-    """The goal of this test is to provide a simple TG with 1-hop neighbors
-    and test the basic functionality of the neighbor sampler.
-    also make sure recency and uniform samplers return the same output.
-    """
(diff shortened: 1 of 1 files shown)


test/unit/test_hooks/test_recency_nbr_hook.py
```diff
@@ -242,6 +242,82 @@ def test_init_basic_sampled_graph_1_hop(basic_sample_graph):
...
+    batch_1 = next(batch_iter)
+    nids, nbr_nids, nbr_times, nbr_feats = _nbrs_2_np(batch_1)
+    assert nids.shape == (1, 2)
+    assert nids[0][0] == 0
+    assert nids[0][1] == 1
+    assert nbr_nids.shape == (1, 2, 1)
+    assert nbr_nids[0][0][0] == PADDED_NODE_ID
+    assert nbr_nids[0][1][0] == PADDED_NODE_ID
+    assert nbr_times.shape == (1, 2, 1)
+    assert nbr_times[0][0][0] == _PADDED_TIME_ID
```
shenyangHuang, line 270: should this be a constant as well `_PADDED_TIME_ID`
though we implicitly use zero.

shenyangHuang: might make it consistent across test by importing the same constant

PR #214, opened 2025-09-04
Vectorized Discretization

Jacob-Chmura:
### Summary / Description

The purpose of this pr is to vectorize our discretization implementation. Goal was to maximize performance gains with minimal implementation change, while also making it easy to add other reduction ops in the future (e.g. _mean_, _last_, _random_).

**Related Issues:** #90 

#### Latency

| Dataset | Main | perf/discretize | Speedup |
|:---:|:---:|:---:|:---:|
| tgbl-wiki | 0.84s | 0.02s | 29.41x |
| tgbn-genre | 95.29s | 1.06s | 89.90x |
| tgbl-coin | 123.39s | 1.28s | 96.27x |

#### Correctness
Discretization unit tests pass. Here are test results on all methods which use `.discretize()`

| Task / Model | Main | perf/discretize |
|---|---|---|
| Graph Prop Persistant Forecast | 0.6378 AUROC | 0.6378 AUROC |
| Graph Prop TGCN | 0.9087 AUROC | 0.9089 AUROC |
| Link Prop GC-LSTM (10 epochs) | 0.2283 MRR | 0.2283 MRR |
| Link Prop GCN (10 epochs) | 0.4068 MRR | 0.4041 MRR |

### DIFF
--- a/tgm/data.py
+++ b/tgm/data.py
@@ -295,27 +295,25 @@ class DGData:
             if ids.ndim == 1:
                 id_key = ids
             else:
-                # Collision free radix-style encoding of edge id [src, dst]
-                # assuming edge ids >= 0, and no overflow
-                base = ids.max().item() + 1
+                # Radix-style encoding of edge [src, dst].
+                # Collision-free assuming ids >= 0 and no overflow
                 src, dst = ids[:, 0], ids[:, 1]
+                base = ids.max().item() + 1
                 id_key = src * base + dst
 
             # Radix-style encoding of [bucket, flat_id]
             base = id_key.max().item() + 1
             final_key = event_buckets * base + id_key
 
-            # Stable sort to get adjacent duplicates while preserving ties
-            # with respect to the original event ordering prior to time conversion
+            # Stable sort to get adjacent duplicates while preserving original event order
             sort_key, sort_idx = torch.sort(final_key, stable=True)
 
             # Mark first occurrence in each [bucket, flat_id] group
             is_first = torch.ones_like(sort_key, dtype=torch.bool)
             is_first[1:] = sort_key[1:] != sort_key[:-1]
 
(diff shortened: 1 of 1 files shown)


tgm/data.py
```diff
@@ -291,19 +291,32 @@ def discretize(
...
-                if node_or_edge not in seen:
-                    seen.add(node_or_edge)
-                    keep.append(i)
-            return torch.tensor(keep, dtype=torch.long)
+
+            if ids.ndim == 1:
+                id_key = ids
+            else:
+                # Radix-style encoding of edge [src, dst].
+                # Collision-free assuming ids >= 0 and no overflow
```
Jacob-Chmura, line 299: Overflow could be problematic, namely if pair of node ids exceed `int32_t` range (~4billion)

tgm/data.py
```diff
@@ -291,19 +291,32 @@ def discretize(
...
+                src, dst = ids[:, 0], ids[:, 1]
+                base = ids.max().item() + 1
+                id_key = src * base + dst
+
+            # Radix-style encoding of [bucket, flat_id]
+            base = id_key.max().item() + 1
+            final_key = event_buckets * base + id_key
+
+            # Stable sort to get adjacent duplicates while preserving original event order
```
Jacob-Chmura, line 308: Stable sorting is only technically required for reduce ops that require preserving order (e.g. first/last). For mean reduction, this won't be necessary.

tgm/data.py
shenyangHuang: I see, this will break if we have a graph with more than 4 billion unique set of nodes

shenyangHuang: maybe let's leave a comment here so it is easier for us to see in the future
`# limit 4 billion nodes`

PR #225, opened 2025-09-10
Add Hook Tutorial

### LINKED ISSUE #106: TGM Hook Tutorial

Jacob-Chmura:
Close #106

### DIFF
--- a/docs/tutorials/hook_tutorial.md
+++ b/docs/tutorials/hook_tutorial.md
@@ -1,20 +1,20 @@
 # Hook Management in TGM
 
-Temporal graph learning pipelines often require dynamic transformations on graph batches—like sampling neighbors, generating negative edges, or moving data to GPU. TGM defines `DGHook`s to provide a flexible, composable way to perform these transformations automatically during batch iteration.
+Temporal graph learning pipelines often require dynamic transformations on graph batches—like sampling neighbors, generating negative edges, or moving data to GPU. TGM defines `DGHook`s to provide a flexible, composable way to perform these transformations automatically during batch iteration. Think of `DGHook`s as all the necessary data processing and operations before you feed the current batch into the TG ML model.
 
 ______________________________________________________________________
 
 ## 1. Hooks: The Basics
 
-A `DGHook` is a callable object that receives a `DGBatch` (a batch of graph events), and a `DGraph` (a temporal view over the entire graph) and returns a transformed `DGBatch`, with additional properties.
+A `DGHook` is a callable object that takes a `DGBatch` (a batch of graph events) and a `DGraph` (a temporal view over the entire graph) as inputs and returns a transformed `DGBatch`, with additional properties.
 
 See [`tgm.graph.DGBatch`](../api/batch.md) for a full reference of the base `DGBatch` yielded by our `DGDataLoader`.
 
 Hooks declare the following information
 
 - `requires: Set[str]`: Names of attributes that the hook needs to exist on the batch
-- `produces: Set[str]`: Names of attributes that the hook adds to the batch
-- `has_state: bool`: A flag to denote whether the hook stores state internally
+- `produces: Set[str]`: Names of attributes from the batch that the hook requires 
+- `has_state: bool`: A flag to denote whether the hook stores state internally (i.e. some memory or attribute that may change upon subsequent invocations of the hook). An example of a stateful hook is a `RecencyNeighborSampler` which keeps track of node interactions over subsequent `__call__`s.
 
 > Note:
 > \- `StatelessHook`: only transforms the batch, no internal state (`has_state = False`)
@@ -22,12 +22,12 @@ Hooks declare the following information
 
 ### Built-in Hooks
 
-TGM ships with several commonly used hooks. The table below summarizes them:
+TGM implements several commonly used hooks. The table below summarizes them:
 
 | Hook Name                    | Type      | `requires` | `produces`                           | Description                                                                                          |
 | ---------------------------- | --------- | ---------- | ------------------------------------ | ---------------------------------------------------------------------------------------------------- |
 | `NegativeEdgeSamplerHook`    | Stateless | None       | `neg`, `neg_time`                    | Generates random negatives for link prediction                                                       |
-| `TGBNegativeEdgeSamplerHook` | Stateless | None       | `neg`, `neg_time`, `neg_batch_list`  | Loads pre-computed negative edges for TGB datasets                                                   |
+| `TGBNegativeEdgeSamplerHook` | Stateless | None       | `neg`, `neg_time`, `neg_batch_list`  | Loads pre-computed negative edges for [TGB](https://tgb.complexdatalab.com/) datasets                                                   |
 | `NeighborSamplerHook`        | Stateless | None       | `nbr_nids`, `nbr_times`, `nbr_feats` | Uniform sampler neighbor for a given number of hops                                                  |
 | `RecencyNeighborSamplerHook` | Stateful  | None       | `nbr_nids`, `nbr_times`, `nbr_feats` | Recency neighbor sampler for a given number of hops                                                  |
 | `PinMemoryHook`              | Stateless | None       | None                                 | Pins all `torch.Tensor` in `DGBatch` for fast CPU-GPU transfer                                       |
@@ -35,7 +35,7 @@ TGM ships with several commonly used hooks. The table below summarizes them:
 
 ### Custom Hooks
 
-If you are developing a new model or new sampling strategy, chances are, you need to define your own hook. The first step is to think about whether you need internal state. If not, you can subclass `tgm.hooks.StatelessHook`.
+Along with the hooks provided by `TGM` team, users are welcome to write custom hooks to perform any operations on `DGBatch` as desired. For instance, if you are developing a new model or new sampling strategy, chances are, all you need to do is define a custom hook. The first step is to think about whether you need internal state. If not, you can subclass `tgm.hooks.StatelessHook`.
 
 For example, the following shows a simple implementation of a negative sampler hook, which add random negative nodes in the range `[10, 100)`, and a corresponding *negative time* which matches the ground truth batch time:
 
@@ -53,11 +53,11 @@ class MyNegativeHook(StatelessHook):
         return batch
 ```
 
-**Important**: Each hooks adds attributes to the batch. Hooks that run after it may dependent on these attributes (with `requires`). More on that later.
+> **Important**: Each hooks adds attributes to the batch. Hooks that run after it may depend on these attributes (defined in `requires`). More on that later.
 
 ## 2. HookManager: Orchestrator of Hooks
 
-`HookManager` manages which hooks are applied to a batch, and in what order. You can think of it like a key-value store where:
+Typically, a full training and evaluation pipeline will require multiple hooks, perhaps some of which execute conditionally on your workload (e.g. validation vs. test). The `HookManager` manages which hooks are applied to a batch, and in what order. You can think of it like a key-value store where:
 
 - *Keys*: e.g. `'train'`, `'val'`, `'test'`
 - *Values*: List of hooks associated with each key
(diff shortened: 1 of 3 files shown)


docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
+# Hook Management in TGM
+
+Temporal graph learning pipelines often require dynamic transformations on graph batches—like sampling neighbors, generating negative edges, or moving data to GPU. TGM defines `DGHook`s to provide a flexible, composable way to perform these transformations automatically during batch iteration.
+
+______________________________________________________________________
+
+## 1. Hooks: The Basics
+
+A `DGHook` is a callable object that receives a `DGBatch` (a batch of graph events), and a `DGraph` (a temporal view over the entire graph) and returns a transformed `DGBatch`, with additional properties.
```
ntgbaoo:
```suggestion
A `DGHook` is a callable object that take a `DGBatch` (a batch of graph events) and a `DGraph` (a temporal view over the entire graph) as inputs and returns a transformed `DGBatch`, with additional properties.
```

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+
+See [`tgm.graph.DGBatch`](../api/batch.md) for a full reference of the base `DGBatch` yielded by our `DGDataLoader`.
+
+Hooks declare the following information
+
+- `requires: Set[str]`: Names of attributes that the hook needs to exist on the batch
```
ntgbaoo, line 15:
```suggestion
- `requires: Set[str]`: List of attributes from the batch that the hook requires
```

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+- `produces: Set[str]`: Names of attributes that the hook adds to the batch
+- `has_state: bool`: A flag to denote whether the hook stores state internally
```
ntgbaoo: If I am a completely new user to `TGM` and read this manual. I may ask what does `state` mean? We may want clearly define it before using the terminology or define it somewhere else below and reference to it.

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+
+> Note:
+> \- `StatelessHook`: only transforms the batch, no internal state (`has_state = False`)
+> \- `StatefulHook`: maintains internal state, (`has_state = True`)
+
+### Built-in Hooks
+
+TGM ships with several commonly used hooks. The table below summarizes them:
```
ntgbaoo:
```suggestion
TGM offers implementations of several commonly used hooks. The table below summarizes them:
```

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+
+| Hook Name                    | Type      | `requires` | `produces`                           | Description                                                                                          |
+| ---------------------------- | --------- | ---------- | ------------------------------------ | ---------------------------------------------------------------------------------------------------- |
+| `NegativeEdgeSamplerHook`    | Stateless | None       | `neg`, `neg_time`                    | Generates random negatives for link prediction                                                       |
+| `TGBNegativeEdgeSamplerHook` | Stateless | None       | `neg`, `neg_time`, `neg_batch_list`  | Loads pre-computed negative edges for TGB datasets                                                   |
```
ntgbaoo:
```suggestion
| `TGBNegativeEdgeSamplerHook` | Stateless | None       | `neg`, `neg_time`, `neg_batch_list`  | Loads pre-computed negative edges for [TGB](https://tgb.complexdatalab.com/) datasets                                                   |
```

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+| `NeighborSamplerHook`        | Stateless | None       | `nbr_nids`, `nbr_times`, `nbr_feats` | Uniform sampler neighbor for a given number of hops                                                  |
+| `RecencyNeighborSamplerHook` | Stateful  | None       | `nbr_nids`, `nbr_times`, `nbr_feats` | Recency neighbor sampler for a given number of hops                                                  |
+| `PinMemoryHook`              | Stateless | None       | None                                 | Pins all `torch.Tensor` in `DGBatch` for fast CPU-GPU transfer                                       |
+| `DeduplicationHook`          | Stateful  | None       | `unique_nids`, `global_to_local`     | Computes unique node ids in `DGBatch` and a mapping from global (graph) to local (batch) coordinates |
+
+### Custom Hooks
+
+If you are developing a new model or new sampling strategy, chances are, you need to define your own hook. The first step is to think about whether you need internal state. If not, you can subclass `tgm.hooks.StatelessHook`.
```
ntgbaoo:
```suggestion
Along with `Hooks` provided by `TGM` team, users are welcome to write custom hooks to perform any operations on `DGBatch` as desired. For instance, if you are developing a new model or new sampling strategy, chances are, all you need to do is define a custom hook. The first step is to think about whether you need internal state. If not, you can subclass `tgm.hooks.StatelessHook`.
```

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+    produces = {'my_neg', 'my_neg_time'}
+    requires = set()
+
+    def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
+        batch.my_neg = torch.randint(10, 100, (len(batch.dst),))
+        batch.my_neg_time = batch.time.clone()
+        return batch
+```
+
+**Important**: Each hooks adds attributes to the batch. Hooks that run after it may dependent on these attributes (with `requires`). More on that later.
```
ntgbaoo:
```suggestion
**Important**: Each hooks adds attributes to the batch. Hooks that run after it may depend on these attributes (defined in `requires`). More on that later.
```

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
+# Hook Management in TGM
+
+Temporal graph learning pipelines often require dynamic transformations on graph batches—like sampling neighbors, generating negative edges, or moving data to GPU. TGM defines `DGHook`s to provide a flexible, composable way to perform these transformations automatically during batch iteration.
```
shenyangHuang: [to add] Think of `DGHook` as all the necessary data processing and operations before you feed the current batch into the TG ML model.

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+See [`tgm.graph.DGBatch`](../api/batch.md) for a full reference of the base `DGBatch` yielded by our `DGDataLoader`.
+
+Hooks declare the following information
+
+- `requires: Set[str]`: Names of attributes that the hook needs to exist on the batch
+- `produces: Set[str]`: Names of attributes that the hook adds to the batch
+- `has_state: bool`: A flag to denote whether the hook stores state internally
+
+> Note:
+> \- `StatelessHook`: only transforms the batch, no internal state (`has_state = False`)
```
shenyangHuang, line 20: we should maybe have the Stateless and Stateful hooks as its own heading section and explain the difference first

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
+# Hook Management in TGM
```
shenyangHuang, line 1: would be great to have a diagram as reference here, could take the one from the paper so people know where hooks are positioned at

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+        batch.my_neg = torch.randint(10, 100, (len(batch.dst),))
+        batch.my_neg_time = batch.time.clone()
+        return batch
+```
+
+**Important**: Each hooks adds attributes to the batch. Hooks that run after it may dependent on these attributes (with `requires`). More on that later.
+
+## 2. HookManager: Orchestrator of Hooks
+
+`HookManager` manages which hooks are applied to a batch, and in what order. You can think of it like a key-value store where:
```
shenyangHuang: Note that we often need more than 1 hook and some of them will be Stateful, some of them stateless. Therefore, the job of a `HookManager` is to manage ....

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+hm.register('train', NegativeEdgeSamplerHook(low=0, high=dg.num_nodes))
+
+# Test: Use the dummy class we defined above
+hm.register('test', MyNegativeHook())
+
+train_loader = DGDataLoader(train_dg, hook_manager=hm)
+test_loader = DGDataLoader(test_dg, hook_manager=hm)
+```
+
+**Important**: When creating custom hooks, you need to make sure you follow the correct hook API. See [`tgm.hooks`](../api/hooks/hooks.md) for more information. A `BadHookProtocolError` will be thrown if you accidently tried registering a hook with the wrong API.
```
shenyangHuang: we also suggest you write your own unit test for your hook to make sure the input / output is as expected. You can find example unit tests of our built-in hooks in `test_hooks`

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+
+You can also selectively reset hooks for a particular key.
+
+```python
+hm.reset_state('train')
+```
+
+## 4. Shared Hooks
+
+Sometimes it happens that you truly want the same hook instance (as in, the same python object) to be shared among different keys. A good example is the `tgm.hooks.RecencyNeighborSampler` which buffers internal state for events for each node in the graph. We want this hook to be shared between 'train' and 'validation' splits, because we want the neighbours accumulated during training to *warm-start* the recent neighbour in the validation split.
```
shenyangHuang: In temporal graph learning, it is common that information you received in the past needed to be used for future prediction. For example, the stored neighbours in `RecencyNeighborSampler` hook is required to be carried to the validation phase to ensure that the models can access information from the training set. Therefore, this rises the need for sharing the state of a hook in train, val and test stages.

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+
+# Register our dummy hook across both the train and test split
+hm.register_shared(MyNegativeHook())
+```
+
+> *Note*: Using shared hooks is typically only useful if the hook has state, that needs to be shared across activation keys.
+
+## 5. Hook Resolution
+
+As you may have guessed, hooks add attributes that may depend on other hooks. Formally, the set of `requires` and `produces` attributes defined on `DGBatch` by the list of hooks defines a directed-acylic-graph (*DAG*) for every key in the hook manager. When we activate a key, the hook manager performs a topological sort of the hook list and finds a topological ordering to execute during data loading. This is only done once and cached, until (if) you decide to add more hooks for that key.
```
shenyangHuang, line 169: one of to do for us is to allow the user to specify the order of operations they want for the hook, maybe one of the default is to set the order as the insertion order and we check for correctness?

docs/tutorials/hook_tutorial.md
```diff
@@ -0,0 +1,245 @@
...
+
+```
+
+______________________________________________________________________
+
+## 6. Recipes
+
+Under construction.
+
+## Summary
```
shenyangHuang, line 241: To do, I can add after ICLR deadline: tutorial for writing your own neighbor sampler like how to find common neighbor etc. (more advanced custom hook tutorial)

shenyangHuang: or a tutorial on a property prediction hook

shenyangHuang (dismissed):

docs/tutorials/hook_tutorial.md
Jacob-Chmura: Yeah my plan is to add a bunch of diagrams from our paper once we have them, throughout the tutorials, and our docs

docs/tutorials/hook_tutorial.md
Jacob-Chmura: Good idea

docs/tutorials/hook_tutorial.md
Jacob-Chmura: Sounds good, I mostly mean state in the traditional programming sense; anything without state is thread safe, can be serialized etc

docs/tutorials/hook_tutorial.md
Jacob-Chmura: Yeah I just added a bit of an exaplanation above, trust that the idea of an object holding state should hopefully not be new to people. If we happen to add more constraints or ideas behind the distinction between the two (e.g. if all your hooks are stateless you can easily shard the data), then we can expand on this section here.

docs/tutorials/hook_tutorial.md
Jacob-Chmura: great idea

docs/tutorials/hook_tutorial.md
Jacob-Chmura: Yeah keeping user defined order is potentially a todo item. I think we need to see more hooks to understand the right approach here.

docs/tutorials/hook_tutorial.md
Jacob-Chmura: Yeah, sounds good. Wer also have a documentation page for "creating your own model in tgm". Perhaps that would be the right place for this.

Issue #203, opened 2025-09-01
Consider keeping dynamic node features and edge features dense in backend


Jacob-Chmura:
**Description**
We never actually use sparse tensors. We immediately convert them to dense when materializing. 

**Motivation**
Our current workflow is slow. Keeping it as dense slices is potentially dangerous if user accidently materializes a huge slice. But, given the typical workflow is via the dataloader, this is probably fine.

**Additional context**
Need to see how much gains we see in speed. It's a much simpler solution if it works.

PR #226, opened 2025-09-10
POC for Dense Feature Store in Backend

Jacob-Chmura:
Aafter profiling, it turns out this is not the primary bottleneck it seems, for `nodeprop` examples at least:

E.g.
Screenshot from 2025-09-10 19-46-55

23/25seconds are spent in tgb evaluator.

However, I did encounter a `IntegerOverflowError` when trying to run the `sparse-coo` backend on `tgbl-review`, which probably had trouble tracking indices of a sparse tensor having size `(T x V x V x d_edge)`

Relevant issue: #203

### DIFF
--- a/tgm/graph.py
+++ b/tgm/graph.py
@@ -76,7 +76,7 @@ class DGraph:
             batch.node_times, batch.node_ids = self.dynamic_node_feats._indices()
             batch.dynamic_node_feats = self.dynamic_node_feats._values()
         if materialize_features and self.edge_feats is not None:
-            batch.edge_feats = self.edge_feats._values()
+            batch.edge_feats = self.edge_feats
         return batch
 
     def slice_events(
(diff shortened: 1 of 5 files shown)


tgm/graph.py
```diff
@@ -76,7 +76,7 @@ def materialize(self, materialize_features: bool = True) -> DGBatch:
             batch.node_times, batch.node_ids = self.dynamic_node_feats._indices()
             batch.dynamic_node_feats = self.dynamic_node_feats._values()
```
Jacob-Chmura, line 77: Currently, I've made our edge features dense (as we never use them sparsely anyway). But dynamic node features are a bit different, because the `indices` are indeed used to extract `batch.node_times, batch.node_ids`, and then we have the actual values `batch.dynamic_node_feats`. One option is to leave this as is for now. Another is to create seperate API points to get dynamic node feats indices and values? Need to discuss. We could just have dynamic node feats returned sparse until post-iclr since dynamic node features are not there in tgb anyway

shenyangHuang: I think keep as is is okay, we can create a separate issue to fix it post ICLR. prioritize finishing experiments for the paper for now.

shenyangHuang (approved):

PR #245, opened 2025-09-18
adding tgn node prop pred example

shenyangHuang:
### Summary / Description

adding node prop pred example 

For CTDG methods on node prop pred, to match the same performance with TGB we shouldn't discretize, because it means less frequent updates potentially. 

Might need to adjust this for DyGformer as well

Validation ndcg=0.3936 for TGN on `tgbn-trade`

### DIFF
--- a/examples/nodeproppred/tgn.py
+++ b/examples/nodeproppred/tgn.py
@@ -42,12 +42,6 @@ parser.add_argument(
     default=[10],
     help='num sampled nbrs at each hop',
 )
-parser.add_argument(
-    '--time-gran',
-    type=str,
-    default='Y',
-    help='raw time granularity for dataset',
-)
 
 
 class NodePredictor(torch.nn.Module):
--- a/examples/nodeproppred/tgn.py
+++ b/examples/nodeproppred/tgn.py
@@ -43,6 +43,12 @@ parser.add_argument(
     default=[10],
     help='num sampled nbrs at each hop',
 )
+parser.add_argument(
+    '--time-gran',
+    type=str,
+    default=None,
+    help='raw time granularity for dataset',
+)
 
 
 class NodePredictor(torch.nn.Module):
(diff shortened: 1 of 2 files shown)


tgm/hooks/hooks.py
```diff
@@ -307,7 +318,8 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
             batch.nbr_times.append(nbr_times)  # type: ignore
             batch.nbr_feats.append(nbr_feats)  # type: ignore
 
-        self._update(batch)
+        if batch.src.numel():
```
Jacob-Chmura, line 321: Nice, creating an issue to add a test case for node-only event batches

tgm/hooks/hooks.py
```diff
@@ -246,6 +247,7 @@ def __init__(
         num_nbrs: List[int],
         edge_feats_dim: int,
         directed: bool = False,
+        seed_nodes_key: str = None,  # type: ignore
```
Jacob-Chmura, line 250: Creating an issue to manage this across both uniforma and recency sampler

tgm/hooks/hooks.py
```diff
@@ -285,14 +288,22 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+                        times.append(batch.neg_time)
+                    seed_nodes = torch.cat(seed)
+                    seed_times = torch.cat(times)
+                else:
+                    seed_nodes = getattr(batch, self._seed_nodes_key)
+                    if seed_nodes is None:
+                        return batch
+                    else:
+                        seed_nodes = seed_nodes.to(device)
+                        seed_times = batch.node_ids.to(device)  # type: ignore #! to adjust later, it should point to node event time
```
Jacob-Chmura: Can't we just so `seed_times = batch.node_times`?

tgm/hooks/hooks.py
```diff
@@ -285,14 +288,22 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+                    seed = [batch.src, batch.dst]
+                    times = [batch.time.repeat(2)]  # Real link times
+                    if hasattr(batch, 'neg'):
+                        batch.neg = batch.neg.to(device)
+                        seed.append(batch.neg)
+                        times.append(batch.neg_time)
+                    seed_nodes = torch.cat(seed)
+                    seed_times = torch.cat(times)
+                else:
+                    seed_nodes = getattr(batch, self._seed_nodes_key)
```
Jacob-Chmura, line 301: This should probably be a list when we actually do it. I may want seed nodes to be something like `batch.node_ids, batch.dst, batch.foo`

And technically, this should make each of these part of the `requires()` so that for example, if you try to do node sampling (`batch.node_ids`) and then you run on a graph without dynamic node events, you get a reasonable error saying this hook cannot be resolved on a graph that has empty dynamic node features

examples/nodeproppred/tgn.py
```diff
@@ -0,0 +1,466 @@
...
+    msg_dim=test_dg.edge_feats_dim,
+    time_enc=memory.time_enc,
+).to(args.device)
+decoder = NodePredictor(in_dim=args.embed_dim, out_dim=num_classes).to(args.device)
+opt = torch.optim.Adam(
+    set(memory.parameters()) | set(encoder.parameters()) | set(decoder.parameters()),
+    lr=args.lr,
+)
+
+with Usage(gpu=True):
```
Jacob-Chmura: We should remove this before merging.

Jacob-Chmura: Also, I created an issue, that our exampels should manager measurement and profiling better #251

tgm/hooks/hooks.py
shenyangHuang: yes, this is actually a bug I think

shenyangHuang: fixed

examples/nodeproppred/tgn.py
shenyangHuang: removed

tgm/hooks/hooks.py
shenyangHuang: I agree, but this should be a systematic PR

Jacob-Chmura (approved):

PR #259, opened 2025-09-21
Adding DOS plotting function and tested 

shenyangHuang:
### Summary / Description

Adding optional plt plotting function for dos, 
tried on tgbn-trade and tgbn-genre with their respective snapshot time gran


Jacob-Chmura (commented):
Nice, won't merge as I'm not sure we want in master, and it requires extra dependency group etc. But good to keep here for now.

PR #261, opened 2025-09-24
Patch Negative Sampler Hook

### LINKED ISSUE #255: Store only TGB Negative sampler without loading PygLinkPropPredDataset
First of all, we don't want suer to have to do load_val_ns and other weird setup. Also, currently, this requires loading in the dataset twice into memory, which doubles our peak memory usage, and is really slow for large dataset.

Instead, we should actually re-write the internal of TGB within our hook: https://github.com/shenyangHuang/TGB/blob/95a898033aa0689bde6b3010639f7ec932099eaa/tgb/linkproppred/dataset.py#L161C13-L165C14

This way user can just do

`hm.register(TGBNegativeEdgeSampler("tgbl-wiki"))` and be done with it.

Jacob-Chmura:
Close #255

### DIFF
--- a/tgm/hooks/hooks.py
+++ b/tgm/hooks/hooks.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 from dataclasses import is_dataclass
+from pathlib import Path
 from typing import Any, List, Set, Tuple
 
 import torch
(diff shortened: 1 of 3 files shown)


tgm/hooks/hooks.py
```diff
@@ -114,22 +115,38 @@ class TGBNegativeEdgeSamplerHook(StatelessHook):
...
+        neg_sampler.load_eval_set(fname=str(val_ns_fname), split_mode='val')
+        neg_sampler.load_eval_set(fname=str(test_ns_fname), split_mode='test')
+
         self.neg_sampler = neg_sampler
         self.split_mode = split_mode
 
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
         # this might complain if the edge is not found in the negative sampler, which could happen if the user is not using the correct version of dataset
-        neg_batch_list = self.neg_sampler.query_batch(  # type: ignore
+        neg_batch_list = self.neg_sampler.query_batch(
```
shenyangHuang: if you can't find the neg batch for the edge, this will show an error message from TGB, I think it is fine for now, but maybe we can a TGBerror or sth and say please verify version number of your dataset, for better error message

PR #269, opened 2025-09-26
DType Management

Jacob-Chmura:
### Summary / Description

The purpose of this PR is to consolidate and standardize our dtype management throughout the codebase.

I am going to use the following rules:

1. All features will be `torch.float32`
2. All integral ids will be `torch.int32` (~2B max)
3. All timestamps will be `torch.int64`

#### Composite Keys
We use radix/lexigraphical sort by node id and time twice
- In our DGData discretize
- In our recency sampler scatter

In both cases, the composite key is of the form: `node_ids * max_time + times` which can overflow `torch.int32`. Pytorch implicitly promotes to 64-bit integers here. We have tests for it.

**Related Issues:** #224

### DIFF
--- a/test/unit/test_data.py
+++ b/test/unit/test_data.py
@@ -176,6 +176,26 @@ def test_init_dg_data_bad_args_invalid_node_id():
         )
 
 
+def test_init_dg_data_node_id_with_downcast_warning():
+    assert False
+
+
+def test_init_dg_data_timestamp_with_downcast_warning():
+    assert False
+
+
+def test_init_dg_data_invalid_node_id_with_id_overflow():
+    assert False
+
+
+def test_init_dg_data_invalid_time_overflow():
+    assert False
+
+
+def test_init_dg_data_number_of_events_overflow():
+    assert False
+
+
 def test_init_dg_data_bad_args_empty_graph():
     # Empty graph not supported
     with pytest.raises(EmptyGraphError):
@@ -911,6 +931,10 @@ def test_discretize_no_op():
     data = DGData.from_raw(edge_timestamps, edge_index, time_delta='s')
     coarse_data = data.discretize('s')
     assert id(data) != id(coarse_data)  # No Shared memory
+
+
+def test_discretize_with_huge_ids_no_overflow():
+    assert False
 
 
 def test_discretize_bad_args():
--- a/test/unit/test_data.py
+++ b/test/unit/test_data.py
@@ -176,6 +176,46 @@ def test_init_dg_data_bad_args_invalid_node_id():
         )
 
 
+def test_init_dg_data_with_nan_feats():
+    edge_index = torch.IntTensor([[0, 0]])
+    edge_timestamps = torch.LongTensor([1])
+    node_ids = torch.IntTensor([0])
+    node_timestamps = torch.LongTensor([2])
+
+    edge_feats = torch.rand((1, 1)).float()
+    edge_feats[0][0] = torch.nan
+
+    with pytest.raises(ValueError):
+        DGData.from_raw(
+            edge_timestamps,
+            edge_index,
+            node_ids=node_ids,
+            node_timestamps=node_timestamps,
+            edge_feats=edge_feats,
+        )
+
+    node_feats = torch.rand((1, 1)).float()
+    node_feats[0][0] = torch.nan
+
+    with pytest.raises(ValueError):
+        DGData.from_raw(
+            edge_timestamps,
+            edge_index,
+            node_ids=node_ids,
+            node_timestamps=node_timestamps,
+            dynamic_node_feats=node_feats,
+        )
+
+    with pytest.raises(ValueError):
+        DGData.from_raw(
+            edge_timestamps,
+            edge_index,
+            node_ids=node_ids,
+            node_timestamps=node_timestamps,
+            static_node_feats=node_feats,
+        )
+
+
 def test_init_dg_data_with_downcast_warning():
     edge_index = torch.IntTensor([[0, 1]])
     edge_timestamps = torch.LongTensor([1])
--- a/test/unit/test_data.py
+++ b/test/unit/test_data.py
@@ -176,22 +176,42 @@ def test_init_dg_data_bad_args_invalid_node_id():
         )
 
 
+@pytest.mark.skip('TODO')
 def test_init_dg_data_node_id_with_downcast_warning():
     assert False
 
 
+@pytest.mark.skip('TODO')
 def test_init_dg_data_timestamp_with_downcast_warning():
     assert False
 
 
+@pytest.mark.skip('TODO')
+def test_init_dg_data_edge_feats_with_downcast_warning():
+    assert False
+
+
+@pytest.mark.skip('TODO')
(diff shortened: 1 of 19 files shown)


shenyangHuang:
Sounds good, however quick question.
Would it be possible that some users might need to use Double or other data types for features? (I think this is totally fine, but maybe we can have it as an argument for a given DGraph?) default float 32

test/unit/test_data.py
```diff
@@ -637,12 +724,16 @@ def test_from_pandas_with_node_features():
...
+    edge_df[['src', 'dst', 't']] = edge_df[['src', 'dst', 't']].astype('int32')
+
     node_dict = {
         'node': [7, 8],
         't': [3, 6],
         'node_features': [torch.rand(5).tolist(), torch.rand(5).tolist()],
     }
     node_df = pd.DataFrame(node_dict)
+    node_df[['node', 't']] = node_df[['node', 't']].astype('int32')
+    # node_df['node_features'] = node_df['node_features'].astype('float32')
```
Jacob-Chmura:
```suggestion
```

tgm/nn/recurrent/gclstm.py
Jacob-Chmura, line 1: Torch geometric requires our indices to be `int64`. Same in `tgcn.py`

tgm/hooks/hooks.py
```diff
@@ -487,7 +487,7 @@ def _update(self, batch: DGBatch) -> None:
 
         # Increment write_pos per node
         num_writes = torch.ones_like(sorted_nodes, device=self._device)
-        self._write_pos.scatter_add_(0, sorted_nodes, num_writes)
+        self._write_pos.scatter_add_(0, sorted_nodes.long(), num_writes)
```
Jacob-Chmura, line 490: Scatter ops require longs

tgm/data.py
```diff
@@ -78,6 +79,42 @@ def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
...
+        _assert_is_tensor(self.timestamps, 'timestamps')
+        _assert_tensor_is_integral(self.timestamps, 'timestamps')
+        if not torch.all(self.timestamps >= 0):
+            raise ValueError('timestamps must all be non-negative')
+        if not torch.all(self.timestamps < max_int32_capacity):
+            raise ValueError(
+                f'timestamps exceed the int32 limit ({max_int32_capacity}). '
+                'TGM does not yet support graphs this large.'
+            )
+        self.timestamps = _maybe_downcast_tensor(self.timestamps, 'timestamps')
```
Jacob-Chmura: This is still fine within our paradigm, but this would probably be the next bottleneck... current unix epoch is over 1B...

tgm/data.py
```diff
@@ -592,13 +650,13 @@ def _check_pandas_import(min_version_number: str | None = None) -> None:
         _check_pandas_import()
 
         # Read in edge data
-        edge_index = torch.from_numpy(
-            edge_df[[edge_src_col, edge_dst_col]].to_numpy()
-        ).long()
-        edge_timestamps = torch.from_numpy(edge_df[edge_time_col].to_numpy()).long()
+        edge_index = torch.from_numpy(edge_df[[edge_src_col, edge_dst_col]].to_numpy())
```
Jacob-Chmura, line 658: For custom datasets (pandas and csv), we avoid numpy-torch conversion which will always upcast to float64. We keep memory buffer the same so that user gets the appropraite warning (if in fact their data will be downcasted by tgm)

tgm/data.py
```diff
@@ -674,13 +735,21 @@ def from_tgb(cls, name: str, **kwargs: Any) -> DGData:
             raise ValueError(f'Unknown dataset: {name}')
 
         data = dataset.full_data
-        src, dst = data['sources'], data['destinations']
-        edge_index = torch.from_numpy(np.stack([src, dst], axis=1)).long()
-        timestamps = torch.from_numpy(data['timestamps']).long()
+
+        # IDs and timestamps are downcast to int32, and features to float32,
```
Jacob-Chmura: TGB stores everything in numpy which auto-casts to 64 bits. We downcast here implicitly since we know its safe, thereby avoiding the user from getting a bunch of warnings on pre-defined data.

tgm/data.py
```diff
@@ -78,6 +79,45 @@ def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
             if x.dtype not in int_types:
                 raise TypeError(f'{name} must have integer dtype but got: {x.dtype}')
 
+        def _maybe_cast_float_tensor(x: Tensor, name: str) -> Tensor:
```
shenyangHuang, line 84: just checking what happens when edge feature is Nan

tgm/data.py
```diff
@@ -78,6 +79,45 @@ def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
...
+        if not torch.all(self.timestamps < max_int32_capacity):
+            raise ValueError(
+                f'timestamps exceed the int32 limit ({max_int32_capacity}). '
+                'TGM does not yet support graphs this large.'
+            )
+        if self.timestamps.dtype != torch.int64:  # This can only be an upcast
+            self.timestamps = self.timestamps.to(torch.int64)
+
+        # Ensure our data does not overflow int32 capacity
+        if len(self.timestamps) > max_int32_capacity:
```
shenyangHuang, line 117: technically with unsigned int64, it can go up more timestamps than this right? (not sure we want unsigned int here though)

tgm/data.py
```diff
@@ -152,6 +210,9 @@ def _assert_tensor_is_integral(x: Tensor, name: str) -> None:
                         'dynamic_node_feats must have shape [num_node_events, D_node_dynamic], '
                         f'got {num_node_events} node events and shape {self.dynamic_node_feats.shape}'
                     )
+                self.dynamic_node_feats = _maybe_cast_float_tensor(
```
shenyangHuang, line 215: might be NaN here

tgm/data.py
```diff
@@ -288,7 +348,8 @@ def discretize(
         # Note: If we simply return a new storage this will 2x our peak memory since the GC
         # won't be able to clean up the current graph storage while `self` is alive.
         time_factor = self.time_delta.convert(time_delta)  # type: ignore
-        buckets = (self.timestamps.float() * time_factor).floor().long()
+        # Doing time conversion in 64-bit to avoid numerical issues with float cast
+        buckets = (self.timestamps.to(torch.float64) * time_factor).floor().int()
```
shenyangHuang, line 354: default floor is fine, potentially later on we can allow user to define some custom behavior

tgm/hooks/hooks.py
shenyangHuang: should be okay considering the batches are not large (or contain the entire graph)

tgm/data.py
Jacob-Chmura: Yeah, this is not a limit on the timestamps values (which are int64), but a limit on the _number_ of timestamps (i.e. the number of total events in the graphs), which is currenlty 2B

tgm/data.py
Jacob-Chmura: It passes through. I'll add a seperate check and unit test to invalidate nan values

PR #272, opened 2025-09-27
TPNet Static Node Feat

### LINKED ISSUE #234: Static Node Feat not on right device
**Describe the bug**
When loading the static node feat from `tgbl-flight`, the tensor is not on the right device. 

**To Reproduce**
run TPNet for `tgbl-flight`. 

**Additional context**
Need to ensure that in DGraph the static node feat matrix is on the same device as DGraph

Jacob-Chmura:
Close #234

### DIFF
--- a/examples/linkproppred/tpnet.py
+++ b/examples/linkproppred/tpnet.py
@@ -253,25 +253,22 @@ seed_everything(args.seed)
 
 evaluator = Evaluator(name=args.dataset)
 
-data = DGData.from_tgb(args.dataset)
-dgraph = DGraph(data)
-num_nodes = dgraph.num_nodes
-edge_feats_dim = dgraph.edge_feats_dim
-
-train_data, val_data, test_data = data.split()
+train_data, val_data, test_data = DGData.from_tgb(args.dataset).split()
 train_dg = DGraph(train_data, device=args.device)
 val_dg = DGraph(val_data, device=args.device)
 test_dg = DGraph(test_data, device=args.device)
 
-if dgraph.static_node_feats is not None:
+if train_dg.static_node_feats is not None:
     static_node_feat = train_dg.static_node_feats
 else:
-    static_node_feat = torch.randn((num_nodes, args.node_dim), device=args.device)
+    static_node_feat = torch.randn(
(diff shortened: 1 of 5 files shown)


examples/linkproppred/tpnet.py
```diff
@@ -255,21 +255,19 @@ def eval(
...
-    static_node_feat = torch.randn((num_nodes, args.node_dim), device=args.device)
-
 train_data, val_data, test_data = data.split()
-
 train_dg = DGraph(train_data, device=args.device)
 val_dg = DGraph(val_data, device=args.device)
 test_dg = DGraph(test_data, device=args.device)
 
+if dgraph.static_node_feats is not None:
+    static_node_feat = train_dg.static_node_feats
```
Jacob-Chmura, line 262: Read the static node feats from train graph as all our other examples (since its on the right device)

tgm/data.py
```diff
@@ -750,7 +750,7 @@ def from_tgb(cls, name: str, **kwargs: Any) -> DGData:
             ],
             dim=1,
         )
-        timestamps = torch.from_numpy(data['timestamps'])
+        timestamps = torch.from_numpy(data['timestamps']).to(torch.int64)
```
Jacob-Chmura, line 753: `tgbn-genre`  holds timestamps as floats it seems @shenyangHuang so I added this explicitly

tgm/nn/model/tpnet.py
```diff
@@ -165,14 +165,18 @@ def update(
...
             self.random_projections[i].scatter_add_(
-                dim=0, index=dst[:, None].expand(-1, self.dim), src=dst_update_messages
+                dim=0,
+                index=dst[:, None].expand(-1, self.dim).long(),
+                src=dst_update_messages,
             )
 
         # set current timestamp to the biggest timestamp in this batch
-        self.now_time.data = torch.tensor(next_time, device=self.device)
+        self.now_time.data = next_time.clone().detach()
```
Jacob-Chmura, line 179: This was previously giving a warning. We may not need a copy construction here at all, but for now this silences the warning

Issue #249, opened 2025-09-19
Add test cases for node-only batches for recency sampler


Jacob-Chmura:

shenyangHuang:
This is related to #242

PR #273, opened 2025-09-27
adding node only batch unit test for hooks

shenyangHuang:
### Summary / Description

Adding node event only batches to hook unit test 

**Related Issues:** #249 

#### Type of Change

- [x] Bug fix

Found undefined behavior for recency sampler when edge feature not defined

### DIFF
--- a/test/unit/test_hooks/test_recency_nbr_hook.py
+++ b/test/unit/test_hooks/test_recency_nbr_hook.py
@@ -703,3 +703,55 @@ def test_tgb_non_time_respecting_negative_neighbor_sampling_test(
     assert nbr_nids[0][4][0] == 2
     assert nbr_nids[1][4][0] == PADDED_NODE_ID
     assert nbr_nids[0][5][0] == 2
+
+
+@pytest.fixture
+def node_only_data():
+    edge_index = torch.LongTensor([[1, 2], [2, 3], [3, 4]])
+    edge_timestamps = torch.LongTensor([1, 2, 3])
+    edge_feats = torch.LongTensor(
+        [[1], [2], [3]]
+    )  # edge feat is simply summing the node IDs at two end points
+    dynamic_node_feats = torch.rand(2, 5)
+    node_timestamps = torch.LongTensor([4, 5])
+    node_ids = torch.LongTensor([5, 6])
+    return DGData.from_raw(
+        edge_timestamps,
+        edge_index,
+        edge_feats=edge_feats,
+        dynamic_node_feats=dynamic_node_feats,
+        node_timestamps=node_timestamps,
+        node_ids=node_ids,
+    )
+
+
+def test_node_only_batch_recency_nbr_sampler(node_only_data):
+    dg = DGraph(node_only_data)
+    hm = HookManager(keys=['unit'])
+    n_nbrs = [1]  # 1 neighbor for each node
+    recency_hook = RecencyNeighborHook(
+        num_nbrs=n_nbrs,
+        num_nodes=dg.num_nodes,
+        edge_feats_dim=dg.edge_feats_dim,
+        directed=True,
+    )
+    hm = HookManager(keys=['unit'])
+    hm.register('unit', recency_hook)
+    loader = DGDataLoader(dg, batch_size=3, hook_manager=hm)
+    with hm.activate('unit'):
+        batch_iter = iter(loader)
+        batch_1 = next(batch_iter)
+        assert isinstance(batch_1, DGBatch)
+        nids, nbr_nids, nbr_times, nbr_feats = _nbrs_2_np(batch_1)
+        assert nids.shape == (1, 6)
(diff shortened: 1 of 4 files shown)


test/unit/test_hooks/test_deduplication_hook.py
Jacob-Chmura, line 1: A test here with an additional batch containing both node and edge events would be nice

test/unit/test_hooks/test_neighbor_sampler_hook.py
```diff
@@ -272,3 +272,76 @@ def test_init_basic_sampled_graph_directed_1_hop(basic_sample_graph):
...
+        assert nbr_times.shape == (1, 6, 1)
+        assert nbr_feats.shape == (1, 6, 1, 1)
+
+        batch_2 = next(batch_iter)
+        assert isinstance(batch_2, DGBatch)
+        nids, nbr_nids, nbr_times, nbr_feats = _nbrs_2_np(batch_2)
+        assert len(nids) == 0  # empty list
+        assert len(nbr_nids) == 0  # empty list
+        assert len(nbr_times) == 0  # empty list
+        assert len(nbr_feats) == 0  # empty list
```
Jacob-Chmura: Should this be an empty list? I thought the convention we wanted was that `len(nbr_nids)` is always number of hops? With empty tensors at each hop?

test/unit/test_hooks/test_recency_nbr_hook.py
```diff
@@ -703,3 +703,87 @@ def test_tgb_non_time_respecting_negative_neighbor_sampling_test(
...
+        assert nbr_times.shape == (1, 6, 1)
+        assert nbr_feats.shape == (1, 6, 1, 1)
+
+        batch_2 = next(batch_iter)
+        assert isinstance(batch_2, DGBatch)
+        nids, nbr_nids, nbr_times, nbr_feats = _nbrs_2_np(batch_2)
+        assert len(nids) == 0  # empty list
+        assert len(nbr_nids) == 0  # empty list
+        assert len(nbr_times) == 0  # empty list
+        assert len(nbr_feats) == 0  # empty list
```
Jacob-Chmura: Same thing here about empty list vs. list of empty tensors

tgm/hooks/hooks.py
```diff
@@ -58,6 +58,10 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
         if hasattr(batch, 'nbr_nids'):
             for hop in range(len(batch.nbr_nids)):
                 nids.append(batch.nbr_nids[hop].to(batch.src.device))
+        if hasattr(batch, 'node_ids'):
+            nids.append(
+                batch.node_ids.to(batch.src.device)
+            ) if batch.node_ids is not None else None
```
Jacob-Chmura:
```suggestion
        nids.append(
            batch.node_ids.to(batch.src.device)
        ) if batch.node_ids is not None else None
```

since `batch.node_ids` always exists as an attribute

tgm/hooks/hooks.py
```diff
@@ -145,37 +153,47 @@ def __init__(self, dataset_name: str, split_mode: str) -> None:
...
+            gen = torch.Generator(device=dg.device)
+            gen.manual_seed(0)
+            batch.neg_time = torch.randint(  # type: ignore
+                int(batch.time.min().item()),
+                int(batch.time.max().item()) + 1,
+                (batch.neg.size(0),),  # type: ignore
+                device=dg.device,
+                generator=gen,
+            )
+            return batch
```
Jacob-Chmura:
```suggestion
        try:
            neg_batch_list = self.neg_sampler.query_batch(
                batch.src, batch.dst, batch.time, split_mode=self.split_mode
            )
        except ValueError as e:
            raise ValueError(
                f'Negative sampling failed for split_mode={self.split_mode}. Try updating your TGB package: `pip install --upgrade py-tgb`'
            ) from e

        batch.neg_batch_list = [  # type: ignore
            torch.tensor(neg_batch, dtype=torch.int32, device=dg.device)
            for neg_batch in neg_batch_list
        ]
        batch.neg = torch.unique(torch.cat(batch.neg_batch_list))  # type: ignore

        # This is a heuristic. For our fake (negative) link times,
        # we pick random time stamps within [batch.start_time, batch.end_time].
        # Using random times on the whole graph will likely produce information
        # leakage, making the prediction easier than it should be.

        # Use generator to local constrain rng for reproducibility
        gen = torch.Generator(device=dg.device)
        gen.manual_seed(0)
        batch.neg_time = torch.randint(  # type: ignore
            int(batch.time.min().item()),
            int(batch.time.max().item()) + 1,
            (batch.neg.size(0),),  # type: ignore
            device=dg.device,
            generator=gen,
        )
        return batch
```

Jacob-Chmura (changes requested):

PR #275, opened 2025-09-28
DGraph Cache on CPU

### LINKED ISSUE #266: Profile Large Datasets and Validate DGraph Caching Hit Rate

Jacob-Chmura:
### Summary / Description

This PR changes the caching behavior of DGraph tensors to always cache on CPU instead of self.device. The public API wraps these caches and moves them to the requested device on access.

Note that Python’s garbage collector cleans up GPU memory (cudafree) almost immediately, so this optimization primarily benefits large batches. Previously, peak memory usage per batch was roughly `2 × batch_size`; now it is `1 × batch_size`. In practice, DGraph objects remain short-lived, so the impact is limited for smaller batches.

Close #266

### DIFF
--- a/tgm/graph.py
+++ b/tgm/graph.py
@@ -193,42 +193,57 @@ class DGraph:
         return self._storage.get_nodes(self._slice)
 
     @cached_property
+    def _edges_cpu(self) -> Tuple[Tensor, Tensor, Tensor]:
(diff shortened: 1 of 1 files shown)


shenyangHuang:
How does this affect the actual run time? can we try some compute time numbers?

PR #276, opened 2025-09-28
Fixed init assumption for RandomProjectionModule

benjaminnNgo:
### Summary / Description

An issue was found when experimenting with `TPNet` for recent submission: current logic in `__init__` in `RandomProjectionModule` fails when trying to set `use_matrix` to false due to incorrect logic.

### DIFF
--- a/tgm/nn/model/tpnet.py
+++ b/tgm/nn/model/tpnet.py
@@ -225,9 +225,7 @@ class RandomProjectionModule(nn.Module):
 
         Returns:
         """
-        assert (
-            len(random_projections) == 2
-        ), (
+        assert len(random_projections) == 2, (
             'Expect a tuple of (now_time,random_projections)'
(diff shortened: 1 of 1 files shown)


tgm/nn/model/tpnet.py
```diff
@@ -41,18 +41,17 @@ def __init__(
...
         super(RandomProjectionModule, self).__init__()
-        assert (not num_edges and not dim_factor) or not enforce_dim or use_matrix
-
-        if enforce_dim is not None:
-            self.dim = enforce_dim
-        elif use_matrix:
-            self.dim = num_nodes
-        elif num_edges is not None and dim_factor is not None:
-            self.dim = min(int(math.log(num_edges * 2)) * dim_factor, num_nodes)
+        if not use_matrix:
```
Jacob-Chmura, line 46: Do we want to make this default `use_matrix = False`?

Issue #246, opened 2025-09-18
Nodeproppred hooks to handle seen nodes for DyGFormer and TPNet


benjaminnNgo:
**Description**
Make hook to maintain seen nodes so far [as follows](https://github.com/tgm-team/tgm/blob/main/examples/nodeproppred/dygformer.py#L172) and [update it](https://github.com/tgm-team/tgm/blob/main/examples/nodeproppred/dygformer.py#L185C13-L188C25)

**Motivation**
To enable nodeproppred for models such as DyGFormer and TPNet, we need to handle seen nodes so far [as follows](https://github.com/tgm-team/tgm/blob/main/examples/nodeproppred/dygformer.py#L172) and [update it](https://github.com/tgm-team/tgm/blob/main/examples/nodeproppred/dygformer.py#L185C13-L188C25) manually for every example

PR #277, opened 2025-09-29
Added Hooks to keep track of which nodes seen before

benjaminnNgo:
### Summary / Description

For event-based models like `DyGFormer` and `TPNet`, there is a case that nodes from node events have not been seen before in any edge events. Hence, we can't perform the backward pass in training. We need to keep track of seen nodes in a hook and only perform the backward pass on seen nodes.

**Related Issues:** #246

#### Type of Change

- [x] Bug fix
- [x] New feature
- [ ] Breaking Change
- [x] Refactoring
- [ ] Documentation update

#### Test Evidence
**Dataset:** `tgbn-genre`
**Loss and Validation over epoch**
The version on this branch shows much more reasonable loss and loss trend over epoch compared to the version in main

DyGFormer_nodeproppred

**Test NDCG:** `0.3036`

- [x] Unit tests
- [ ] Integration tests
- [ ] Performance tests

I switched the order of operations so that I want to update the seen nodes mask  according to the current batch before inference `seen_nodes`. I am re-running `DyGFormer` example to see how that affects the performance.


test/unit/test_hooks/test_seen_nodes_track_hook.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture
+def dg():
+    edge_index = torch.IntTensor([[2, 3], [2, 5]])
+    edge_timestamps = torch.LongTensor([1, 5])
+    node_timestamps = torch.LongTensor([2, 3, 5, 5, 5])
+    node_ids = torch.IntTensor([4, 2, 5, 1, 2])
+
+    # Can't actually get node events without dynamic node feats
```
Jacob-Chmura, line 17: Good point, might be worth fixing in the near future

examples/nodeproppred/dygformer.py
```diff
@@ -179,27 +183,17 @@ def train(
...
-
-            if len(keep_mask) == 0:
-                # First time all nodes are new, skip backward
-                seen_nodes.update(batch_nodes)
+            if len(batch.seen_nodes) == 0:
                 continue
 
-            train_idx = torch.tensor(keep_mask, device=z.device)
-            z_node = z[train_idx]
+            z_node = z[batch.seen_nodes]
```
Jacob-Chmura, line 185: So, to be clear, `|z|` is the number of nodes in the entire graph (not in the batch)?

tgm/hooks/hooks.py
```diff
@@ -498,6 +498,59 @@ def _move_queues_to_device_if_needed(self, device: torch.device) -> None:
...
+        ValueError: If the num_nodes list is negative.
+    """
+
+    requires = {'unique_nids'}
+    produces = {'seen_nodes', 'batch_nodes_mask'}
+
+    def __init__(
+        self,
+        num_nodes: int,
+    ) -> None:
```
Jacob-Chmura:
```suggestion
    def __init__(self, num_nodes: int) -> None:
```

tgm/hooks/hooks.py
```diff
@@ -498,6 +498,59 @@ def _move_queues_to_device_if_needed(self, device: torch.device) -> None:
...
+            raise ValueError('num_nodes must be non-negative')
+        self._seen_mask = torch.zeros(num_nodes, dtype=torch.bool)
+        self._device = torch.device('cpu')
+
+    def reset_state(self) -> None:
+        self._seen_mask.fill_(False)
+
+    def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
+        device = dg.device
+        self._move_to_device_if_needed(device)  # No-op after first batch
```
Jacob-Chmura:
```suggestion
        self._move_to_device_if_needed(dg.device)  # No-op after first batch
```

tgm/hooks/hooks.py
```diff
@@ -498,6 +498,59 @@ def _move_queues_to_device_if_needed(self, device: torch.device) -> None:
...
+
+        if batch.node_ids is not None:
+            batch_nodes = batch.node_ids
+        else:
+            batch_nodes = torch.empty(
+                0,
+                device=self._device,
+                dtype=torch.int,
+            )
```
Jacob-Chmura:
```suggestion
            batch_nodes = torch.empty(0, device=self._device, dtype=torch.int)
```

tgm/hooks/hooks.py
```diff
@@ -498,6 +498,59 @@ def _move_queues_to_device_if_needed(self, device: torch.device) -> None:
             self._write_pos = self._write_pos.to(device)
 
 
+class SeenNodesTrackHook(StatefulHook):
+    """This hook return all nodes appearing in node events of current batch that have seen in the past edge events.
```
shenyangHuang: this is very specific and not clear from the name of the hook. 
The hook name suggest it keeps track of all nodes that have been seen so far, however it only contains nodes that are seen in the node events of the current batch (a very small subset of all seen nodes). We should use a more specific hook name

tgm/hooks/hooks.py
```diff
@@ -498,6 +498,59 @@ def _move_queues_to_device_if_needed(self, device: torch.device) -> None:
...
+        if batch.node_ids is not None:
+            batch_nodes = batch.node_ids
+        else:
+            batch_nodes = torch.empty(
+                0,
+                device=self._device,
+                dtype=torch.int,
+            )
+
+        edge_event_nodes = batch.unique_nids  # type: ignore[attr-defined]
```
Jacob-Chmura: Note, this includes node events as per the most recent prs, so I think the current unit test should break after merging upstream to main. We may need to manually reconstruct unique edge event nodes here again :/

tgm/hooks/hooks.py
```diff
@@ -498,6 +498,59 @@ def _move_queues_to_device_if_needed(self, device: torch.device) -> None:
...
+
+        self._seen_mask[edge_event_nodes] = True
+        previous_seen = self._seen_mask[batch_nodes]
```
Jacob-Chmura: Shouldn't previous seen be computed _before_ updating the seen mask?

Jacob-Chmura (changes requested):

tgm/hooks/hooks.py
Jacob-Chmura: Still not sure abou this @benjaminnNgo

benjaminnNgo:
Test message

PR #279, opened 2025-09-29
Remove Edge Feat Dim from Recency Sampler API

### LINKED ISSUE #278: Remove Edge Feat Dim from Recency API

Jacob-Chmura:
Close #278

### DIFF
--- a/tgm/hooks/hooks.py
+++ b/tgm/hooks/hooks.py
@@ -288,6 +288,7 @@ class RecencyNeighborHook(StatefulHook):
         self._write_pos = torch.zeros(num_nodes, dtype=torch.int32)
 
         # Wait until first __call__ to infer the edge_feats_dim on the underlying graph
+        self._need_to_initialize_nbr_feats = True
         self._edge_feats_dim = None
         self._nbr_feats = None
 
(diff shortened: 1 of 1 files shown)


tgm/hooks/hooks.py
```diff
@@ -276,31 +274,37 @@ def __init__(
...
         self._nbr_ids.fill_(PADDED_NODE_ID)
         self._nbr_times.zero_()
-        self._nbr_feats.zero_()
         self._write_pos.zero_()
 
+        if self._nbr_feats is not None:
+            self._nbr_feats.zero_()
+
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
+        self._initialize_nbr_feats_if_needed(dg)
```
shenyangHuang, line 308: do we need to run this check each time? maybe run it and then record it in a boolean?

tgm/hooks/hooks.py
```diff
@@ -490,13 +496,21 @@ def _update(self, batch: DGBatch) -> None:
...
+        assert self._nbr_feats is not None  # For mypy
         if device != self._device:
             self._device = device
             self._nbr_ids = self._nbr_ids.to(device)
             self._nbr_times = self._nbr_times.to(device)
             self._nbr_feats = self._nbr_feats.to(device)
             self._write_pos = self._write_pos.to(device)
 
+    def _initialize_nbr_feats_if_needed(self, dg: DGraph) -> None:
+        if self._edge_feats_dim is None and dg.edge_feats_dim is not None:
```
shenyangHuang: so when `dg.edge_feats_dim` is None, then `nbr_feats` is None?

tgm/hooks/hooks.py
Jacob-Chmura: We need some sort of a check each time, whether the check involves a boolean flag, or the check of `self._edge_feats_dim is None`

tgm/hooks/hooks.py
Jacob-Chmura: Yes, that's what I did here. I can directly constuct the empty tensor in this PR as you've done in #249 if you want

shenyangHuang (approved):

tgm/hooks/hooks.py
shenyangHuang: lets merge with the other PR

PR #280, opened 2025-09-29
GraphMixer Fix

### LINKED ISSUE #257: Graphmixer overhaul
Performance is questionable. Need to double check implementation when moving modules inside tgm.

Jacob-Chmura:
Close #257 

Previously we were computed TIME_GAP nbr info _for each node_). All we really need is to track the last TIME_GAP events on _the entire graph_. Luckily the storage can answer this query efficiently for us, so no need to manage state. The previous implementation was essentially using _more_ information than we should have been, giving mrr ~0.65 (way higher than tgb leaderboard).

Here is 1 epoch `tgbl-wiki` numbers on A100L

| Branch                | Train Latency (s) | Peak GPU (GB) | Val MRR | Test MRR |
|-----------------------|-------------:|--------------:|--------:|---------:|
| master                | 59.16          |  2.55         |  N/A      | N/A        |
| dev/graphmixer_fix    | 16.14          |  0.40           |  0.1039       |  -        |

### DIFF
--- a/examples/linkproppred/graphmixer.py
+++ b/examples/linkproppred/graphmixer.py
@@ -119,7 +119,7 @@ class GraphMixerEncoder(nn.Module):
             if batch.time_gap_nbrs[i]:
                 time_gap_feat[i] = node_feat[batch.time_gap_nbrs[i]].mean(dim=0)
 
-        z_node = time_gap_feat + node_feat[seed_node]
+        z_node = time_gap_feat + node_feat[torch.cat([batch.src, batch.dst, batch.neg])]
         z = self.output_layer(torch.cat([z_link, z_node], dim=1))
         return z
 


examples/linkproppred/graphmixer.py
```diff
@@ -196,75 +196,38 @@ def eval(
...
         device = dg.device
-        self._move_queues_to_device_if_needed(device)  # No-op after first batch
+        seed_nodes = torch.cat([batch.src, batch.dst, batch.neg.to(device)])
 
-        batch.neg = batch.neg.to(device)
+        # Construct a the time_gap slice
+        time_gap_slice = replace(dg._slice)
+        time_gap_slice.start_idx = max(dg._slice.end_idx - self._time_gap, 0)
+        time_gap_slice.end_time = int(batch.time.min()) - 1
+        time_gap_src, time_gap_dst, _ = dg._storage.get_edges(time_gap_slice)
```
Jacob-Chmura, line 221: Main idea:

1. Instead of building huge sparse tensor of size (seed_nodes, time_gap), just use dictionary of seed_nodes -> list of nbrs. This has much less memory overhead
2. Instead of tracking history internally, we can just mangle the slice and query the backend for the edge history in the interval [t - TIME_GAP, t)

tgm/nn/decoder/linkproppred.py
Jacob-Chmura, line 1: CRITICAL bug which silently inflated the performance of _all_ our methods due to shape misalignment

tgm/nn/decoder/linkproppred.py
```diff
@@ -61,4 +61,4 @@ def forward(self, z_src: torch.Tensor, z_dst: torch.Tensor) -> torch.Tensor:
             z_src (torch.Tensor): embedding of src node
             z_dst (torch.Tensor): embedding of dst node
         """
-        return self.model(self.merge_op(z_src, z_dst))
+        return self.model(self.merge_op(z_src, z_dst)).view(-1)
```
shenyangHuang, line 64: yeah this was causing the GCN bug previously as well (or sth similar to this)

examples/linkproppred/graphmixer.py
shenyangHuang: so, it is similar to uniform neighbor sampler

Jacob-Chmura: yes, uniform sampler with a max lower bound on the "lookback" whereas uniform hook can in the worst case samplle to t=0

Issue #251, opened 2025-09-19
Properly handle measurement during examples


Jacob-Chmura:
Once our models are moved out of the examples, we should consider doing a better job of capturing  / measurement output. Similar to our ICLR prs, where we have optional flags liike `--capture-gpu`. We should have more of these.

And, we should make it straightforward to report all sorts of latencies and perf numbers, and save to output file. We need this anyway for proper integrarion tests to capture the outpu and have it parsed on github #47

PR #281, opened 2025-09-29
TGM Standardized Logging

### LINKED ISSUE #260: Standardize library logging

Jacob-Chmura:
Close #260

### DIFF
--- a/tgm/util/logging.py
+++ b/tgm/util/logging.py
@@ -1,0 +1,67 @@
+import functools
+import logging
+import time
+from pathlib import Path
+from typing import Any, Callable, List
+
+
+def enable_logging(
+    *,
+    console_log_level: int = logging.INFO,
+    file_log_level: int = logging.DEBUG,
+    log_file_path: str | Path | None = None,
+) -> None:
+    """Enable library-wide logging to stdout and optionally to a file.
+
+    Args:
+        console_log_level (int): Logging level for console stream handler (default = logging.INFO).
+        file_log_level (int): Logging level for file handler if configured (default = logging.DEBUG).
+        log_file_path (Optional[str | Path]): Optional path to a log file.
+    """
+    console_formatter = logging.Formatter(
+        '[%(asctime)s] %(name)s | %(levelname)s | %(message)s',
+        datefmt='%Y-%m-%d %H:%M:%S',
+    )
+    console_handler = logging.StreamHandler()
+    console_handler.setLevel(console_log_level)
+    console_handler.setFormatter(console_formatter)
+
+    handlers: List[logging.Handler] = [console_handler]
+    if log_file_path is not None:
+        file_formatter = logging.Formatter(
+            '[%(asctime)s] %(name)s - %(levelname)s '
+            '[%(processName)s %(threadName)s %(name)s.%(funcName)s:%(lineno)d] '
+            '%(message)s',
+            datefmt='%Y-%m-%d %H:%M:%S',
+        )
+        file_handler = logging.FileHandler(filename=log_file_path, mode='a')
+        file_handler.setLevel(file_log_level)
+        file_handler.setFormatter(file_formatter)
+        handlers.append(file_handler)
+
+    logger = logging.getLogger('tgm')
+    logger.setLevel(min(console_log_level, file_log_level))
+    for handler in handlers:
+        logger.addHandler(handler)
+    logger.propagate = False  # Don't spam user's root logger
+
+
+def _get_logger(name: str) -> logging.Logger:
+    logger = logging.getLogger(name)
+    if not logger.handlers:
+        logger.addHandler(logging.NullHandler())
+    return logger
+
+
+def _log_latency(func: Callable) -> Any:
+    logger = _get_logger(func.__module__)
+
+    @functools.wraps(func)
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        start_time = time.perf_counter()
+        result = func(*args, **kwargs)
+        latency = time.perf_counter() - start_time
+        logger.info(f"Function '{func.__name__}' executed in {latency:.4f} seconds")
+        return result
+
+    return wrapper
--- a/docs/api/util.md
+++ b/docs/api/util.md
@@ -1,1 +1,3 @@
 ::: tgm.util.seed
+::: tgm.util.logging
+::: tgm.util.perf


examples/graphproppred/persistant_forecast.py
```diff
@@ -45,6 +48,13 @@
     default='W',
     help='time granularity to operate on for snapshots',
 )
+parser.add_argument(
```
Jacob-Chmura, line 51: Will actually use this in our integration tests in a seperate pr while addressing #47

examples/graphproppred/persistant_forecast.py
```diff
@@ -109,6 +119,7 @@ def __call__(self, batch: DGBatch) -> torch.tensor:
         return pred
 
 
+@log_latency
```
Jacob-Chmura, line 122: In all the examples, I log latency for train and eval. Once we move our models into tgm core, we should default enable model profiling under `logging.DEBUG` as well

tgm/_storage/backends/array_backend.py
```diff
@@ -203,11 +214,14 @@ def _binary_search(self, slice: DGSliceTracker) -> Tuple[int, int]:
         if slice.start_time not in self._lb_cache:
             t = ts[0] if slice.start_time is None else slice.start_time
             self._lb_cache[slice.start_time] = int(torch.searchsorted(ts, t))
+            logger.debug(f'Cache miss: start_time={slice.start_time}')
```
Jacob-Chmura, line 219: Consider making a TRACE level and switching this to TRACE logger

tgm/data.py
```diff
@@ -310,6 +324,7 @@ def split(self, strategy: SplitStrategy | None = None) -> Tuple[DGData, ...]:
 
         return strategy.apply(self)
 
+    @log_latency(level=logging.DEBUG)
```
Jacob-Chmura, line 327: All internal latency profiles must be DEBUG level

tgm/graph.py
```diff
@@ -165,34 +185,34 @@ def end_time(self) -> Optional[int]:
             self._slice.end_time = self._storage.get_end_time(self._slice)
         return self._slice.end_time
 
-    @cached_property
+    @_cached_property_log_cache_activity
```
Jacob-Chmura: Created our own decorator to track cache usage at DEBUG level. Consider making it trace, as it's quite spammy, even for DEBUG. Also, it reveals the obvious fact that typical recipes never get  cache hit.

tgm/hooks/hook_manager.py
```diff
@@ -134,7 +143,11 @@ def execute_active_hooks(self, dg: DGraph, batch: DGBatch) -> DGBatch:
             self.resolve_hooks(key)
 
         for hook in self._key_to_hooks[key]:
+            start_time = time.perf_counter()
             batch = hook(dg, batch)
+            latency = time.perf_counter() - start_time
+            logger.debug('%s hook executed in %.4fs', hook.__class__.__name__, latency)
```
Jacob-Chmura, line 149: Explicit hook profiling here instead of with decorator so user-defined hooks get auto-profiled.

tgm/hooks/hooks.py
```diff
@@ -136,10 +144,14 @@ def __init__(self, dataset_name: str, split_mode: str) -> None:
         else:
             version_suffix = ''
 
-        val_ns_fname = root / f'{dataset_name}_val_ns{version_suffix}.pkl'
-        test_ns_fname = root / f'{dataset_name}_test_ns{version_suffix}.pkl'
-        neg_sampler.load_eval_set(fname=str(val_ns_fname), split_mode='val')
-        neg_sampler.load_eval_set(fname=str(test_ns_fname), split_mode='test')
```
Jacob-Chmura, line 149: Made a diff here. I think it's safe to assume that when `self.split_mode = 'val'`, we only need to load val samples ( and analogous for test). This will save us some IO

tgm/loader.py
```diff
@@ -152,6 +162,12 @@ def __call__(self, slice_start: List[int]) -> DGBatch:
         dg = self._slice_op(slice_start[0], slice_end)
         batch = dg.materialize()
         if self._hook_manager is not None:
+            logger.debug(
```
Jacob-Chmura, line 165: Not the most useful log here, but still gives us some signal

tgm/util/logging.py
```diff
@@ -0,0 +1,108 @@
...
+    logger = logging.getLogger(name)
+    if not logger.handlers:
+        logger.addHandler(logging.NullHandler())
+    return logger
+
+
+logger = _get_logger(__name__)
+
+
+class _cached_property_log_cache_activity(functools.cached_property):
```
Jacob-Chmura: This should potentially go to trace level, but it's a useful class, generally speaking

examples/analytics/dos.py
```diff
@@ -1,4 +1,6 @@
 import argparse
+import logging
```
shenyangHuang, line 2: a little strange to me that we need both `logging` and `tgm.util.logging` but I am not too familiar with logging correctly.

examples/graphproppred/persistant_forecast.py
shenyangHuang: clean and cool

examples/graphproppred/tgcn.py
```diff
@@ -185,6 +195,7 @@ def train(
     return total_loss, h_0, metrics.compute()
 
 
+@log_latency
```
shenyangHuang, line 198: to add later maybe, can we have sth like `@log_profiler`

shenyangHuang: where it will automatically run the profiler and log the breakdowns

tgm/_storage/backends/array_backend.py
```diff
@@ -23,12 +26,14 @@ def __init__(self, data: DGData) -> None:
     def get_start_time(self, slice: DGSliceTracker) -> Optional[int]:
         lb_idx, ub_idx = self._binary_search(slice)
         if lb_idx >= ub_idx:
+            logger.debug('No events in slice: %s', slice)
```
shenyangHuang, line 29: is it possible to log too many things? maybe dataloader will just log a total of x slices has no events?

shenyangHuang: I think it is fine as is, just potentially some TG will have a lot of empty slices (also depend on your time gran)

examples/analytics/dos.py
Jacob-Chmura: Think of these examples as application code. We need `logging` to setup our application-side logger, which emites to `tgm/{__file__}` (e.g. `tgm/dos`). On the other hand, `tgm.util.logging` is the tgm internal logging. In a proper application, you would likely have your own logging module setup rather than using a barebones logger as I did here.

examples/graphproppred/tgcn.py
Jacob-Chmura: Yes, good point. I'll add this when doing #251

tgm/_storage/backends/array_backend.py
Jacob-Chmura: Figuring out the right level to log at is its own task when it comes to proper production software. It's definitely possible to have noisy logs. This one in particular is not too bad I don't think, but moving forward, we should think carefully around what we are logging.

As I mentioned in my earlier review, for this purpose I may consider adding a `TRACE` log level (not natively there in python) which is dev-focused (e.g. for cache invalidation)

tgm/graph.py
```diff
@@ -165,34 +185,34 @@ def end_time(self) -> Optional[int]:
...
         return len(src)
 
-    @cached_property
+    @_cached_property_log_cache_activity
     def num_timestamps(self) -> int:
         """The total number of unique timestamps encountered over the dynamic graph."""
         return self._storage.get_num_timestamps(self._slice)
 
-    @cached_property
+    @_cached_property_log_cache_activity
```
Jacob-Chmura:
```suggestion
    @_cached_property_log
```

tgm/loader.py
shenyangHuang: lets add the logs in for now, later on once we get more usage feedback, we can add or remove logs, it is hard to know usefulness exactly at this point.

tgm/util/logging.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import Any, Callable, List
+
+
+def enable_logging(
+    *,
+    console_log_level: int = logging.INFO,
+    file_log_level: int = logging.DEBUG,
+    log_file_path: str | Path | None = None,
+) -> None:
+    """Enable library-wide logging to stdout and optionally to a file.
```
shenyangHuang, line 14: empty log_file_path means logging is disabled.

tgm/util/logging.py
```diff
@@ -0,0 +1,108 @@
...
+    handlers: List[logging.Handler] = [console_handler]
+    if log_file_path is not None:
+        file_formatter = logging.Formatter(
+            '[%(asctime)s.%(msecs)03d] %(name)s - %(levelname)s '
+            '[%(processName)s %(threadName)s %(name)s.%(funcName)s:%(lineno)d] '
+            '%(message)s',
+            datefmt='%Y-%m-%d %H:%M:%S',
+        )
+        file_handler = logging.FileHandler(filename=log_file_path, mode='a')
+        file_handler.setLevel(file_log_level)
```
shenyangHuang, line 41: not familiar with this, what if the user enter a logging level that is a typo like "DEBG"

shenyangHuang (approved):

tgm/util/logging.py
Jacob-Chmura: Log levels are just enums that are really just integers. To be clear "DEBUG" is not a valid option either. The input is `int` as the type hint suggest. If a user tried to do `logging.DEBG` this would fail outside of our code, saying there is no such enum.

PR #282, opened 2025-09-30
Move all model definition to tgm.nn.model

benjaminnNgo:
### Summary / Description

Move all model definitions to `tgm.nn.model`. Currently, our hierarchy is as follows:
```
tgm/nn/
├── model
│   ├── dygformer.py
│   ├── edgebank.py
│   ├── gclstm.py
│   ├── graphmixer.py
│   ├── tgcn.py
│   └── tpnet.py
├── attention.py
└── time_encoding.py
```

**Related Issues:** #162 , #267

#### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking Change
- [X] Refactoring
- [ ] Documentation update

#### Test Evidence

- [X] Unit tests
- [ ] Integration tests [run](https://github.com/tgm-team/tgm/actions/runs/18114998370)
- [ ] Performance tests

Where should we place `attention.py` and `time_encoding.py`. Is everything `model`?

### DIFF
--- a/tgm/nn/encoder/tpnet.py
+++ b/tgm/nn/encoder/tpnet.py
@@ -41,7 +41,7 @@ class RandomProjectionModule(nn.Module):
         dim_factor: int | None = None,
         device: str = 'cpu',
     ) -> None:
-        super(RandomProjectionModule, self).__init__()
+        super().__init__()
         if not use_matrix:
             if enforce_dim is not None:
                 self.dim = enforce_dim
@@ -250,11 +250,10 @@ class TPNet(nn.Module):
         node_feat_dim (int): Dimension of static/dynamic node features (`d_N`).
         edge_feat_dim (int): Dimension of edge features (`d_E`).
         time_feat_dim (int): Dimension of time encodings (`d_T`).
-        channel_embedding_dim (int): Dimension of each channel embedding.
         output_dim (int): Dimension of output embedding.
-        dropout (float): Drop out rate.
-        num_layers (int): Number of transformer layers.
         num_neighbors (int): Number of recent temporal neighbors consider
+        num_layers (int): Number of transformer layers.
+        dropout (float): Drop out rate.
         random_projections (nn.Module): Random projection module that maintains a series temporal walk matrices
         device (str) : cpu or cuda
         time_encoder (PyTorch Module) : Time encoder module.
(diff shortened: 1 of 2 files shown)


docs/api/nn.md
Jacob-Chmura, line 1: Nice, we'll want to also add a section for decoders

tgm/nn/decoder/graphproppred.py
```diff
@@ -0,0 +1,26 @@
...
+import torch.nn as nn
+
+
+def mean_pooling(z: torch.Tensor) -> torch.Tensor:
+    r"""Default graph pooling: Mean pooling."""
+    # @TODO: we can define this in different module and have a base class for this
+    return torch.mean(z, dim=0).squeeze()
+
+
+class GraphPredictor(torch.nn.Module):
```
Jacob-Chmura, line 21: We'll want to add docstrings and tests for the decoders

tgm/nn/__init__.py
```diff
@@ -1,8 +1,14 @@
 from .attention import TemporalAttention
-from .memory import EdgeBankPredictor
 from .time_encoding import Time2Vec
-from .recurrent import GCLSTM
-from .model import DyGFormer, TPNet, RandomProjectionModule
+from .model import (
```
Jacob-Chmura: Maybe we call this `encoders`? 

That way we have `encoders`, `decoders`

tgm/nn/encoder/tpnet.py
```diff
@@ -7,6 +7,7 @@
 from tgm.constants import PADDED_NODE_ID
 
 from ..time_encoding import Time2Vec
+from .graphmixer import MLPMixer
```
Jacob-Chmura: Since MPLMixer is shared, we should move it out of here, and into `tgm/nn` along with time encoding and attention I think

Jacob-Chmura: I would create directory `modules` and move mplmixer, time encoder, and attention there.

Jacob-Chmura (changes requested):
Looks almost there, just some minor things. And we should add unit tests for our model coverage in this PR I think, unless its blocking something else (which I don't think it is)

tgm/nn/encoder/tpnet.py
Jacob-Chmura: So we have `nn/encoders, nn/decoders, nn/modules`. Of course, all is exported up to `nn` namespace so user can just get it from `tgm.nn`

PR #283, opened 2025-09-30
Generic Seed Nodes Sampler API

### LINKED ISSUE #250: Add seed node keys abstraction to both samplers

Jacob-Chmura:
#### Overview
The purpose of this PR is to enable customization of the seed node and query times to start your neighbor sampling; i.e. the query on the first hop. Subsequent hops bootstrap from this (assuming it's not empty).

This involves paramaterizing two fields:
1. `seed_nodes_keys`: The node ids to start sampling from
2. `seed_times_keys`: The query times to sample from

#### Notes
- by default, we will use the link prediction seed set (`seed_nodes_keys` = ['src', 'dst', 'neg'], `seed_times_keys` = ['time', 'time', 'neg_time']
- TGN node prop is the only current use case that switches it up, using the dyamic node events as it's sampling seeds
- did some clean up of the recent additions to the nbr hooks which were messy
- unified more of the api across samplers so that in the future we can merge their API completely (or create base nbr sampling hook interface)

#### Corner cases
The user could technically specify atrbitrary attrs on the batch. These are the rules I followed
- If a batch is missing the attributes requested, we emit a _1-time UserWarning_ to signal to the user, something could be off. Keep in mind, it could very well be that an attribute does not exist on a given batch (but this is unlikely, it would probably be set to `None` in that case). We _could_ consider raising an exception here.
- If the batch attribute is there, but is `None` (e.g. dynamic node events exist, but not on the current batch), then we emit a _1-time_ UserWarning.
- If the batch attribute is not an `Optional[Tensor]` or is a tensor, but is not 1-dimensional, we raise a ValueError. We can't reasonable recover from these
- If the batch attribute is a 1d tensor, but has value out of range, we raise a ValueError as we can't recover. Bad seed times are any negative timestamps. Bad seed node ids are any ids that are either negative, or > than number of nodes on the graph. (Technically, we could bypass this by just considering this an empty batch instead of throwing a runtimeerorr....)

If after these filters we end up with no seed nodes, we return empty tensors as per the convention established in #150. 

#### Out of scope
- If we really want to support generic ids / times, we should also handle upcasting/downcasting dtypes, device transfer, and potentially more to ensure our sampler doesn't raise an uncaught exception. 
- Uniform sampler does not direclty have knowledge of number of nodes in the graph. I'm considering dropping `num_nodes` from recnecy API, and have both samplers dynamically infer it on the first batch (similar to edge feats dim #278 ). To be discussed since you may want to in fact run recency on a subset of nodes (in which case, it should be a default argument in _both_ samplers).

**To be discussed**
- Consider raising an exception on missing attribute. Users should follow our convention and use None if it's not present on the batch.
- Whether we want to make the `num_nodes` api change for beta.

Close #250

### DIFF
--- a/examples/nodeproppred/tgn.py
+++ b/examples/nodeproppred/tgn.py
@@ -409,7 +409,8 @@ num_classes = train_dg.dynamic_node_feats_dim
 nbr_hook = RecencyNeighborHook(
     num_nbrs=args.n_nbrs,
     num_nodes=test_dg.num_nodes,  # Assuming node ids at test set > train/val set
-    seed_nodes_key='node_ids',
+    seed_nodes_keys=['node_ids'],
+    seed_times_keys=['node_times'],
 )
 
 hm = HookManager(keys=['train', 'val', 'test'])
--- a/test/unit/test_hooks/test_recency_nbr_hook.py
+++ b/test/unit/test_hooks/test_recency_nbr_hook.py
@@ -163,39 +163,87 @@ def test_bad_neighbor_sampler_init_bad_seed_keys():
         )
 
 
-@pytest.mark.skip('TODO')
-def test_sample_with_node_events_seeds():
-    pass
+def test_sample_with_node_events_seeds(node_only_data):
+    dg = DGraph(node_only_data)
+    hook = RecencyNeighborHook(
+        num_nbrs=[1],
+        num_nodes=dg.num_nodes,
+        seed_nodes_keys=['node_ids'],
+        seed_times_keys=['node_times'],
+    )
+    batch = dg.materialize()
+    batch = hook(dg, batch)
+    assert len(batch.nids) == 1
+    assert len(batch.times) == 1
+    torch.testing.assert_close(batch.nids[0], batch.node_ids)
+    torch.testing.assert_close(batch.times[0], batch.node_times)
 
 
-@pytest.mark.skip('TODO')
-def test_sample_with_None_seeds():
-    pass
+def test_sample_with_none_seeds(basic_sample_graph):
+    dg = DGraph(basic_sample_graph)
+    hook = RecencyNeighborHook(
+        num_nbrs=[1], num_nodes=2, seed_nodes_keys=['foo'], seed_times_keys=['bar']
+    )
+    batch = dg.materialize()
+
+    with pytest.warns(UserWarning):
+        batch = hook(dg, batch)
 
 
-@pytest.mark.skip('TODO')
-def test_sample_with_non_existent_seeds():
-    pass
+def test_bad_sample_with_non_tensor_non_None_seeds(basic_sample_graph):
+    dg = DGraph(basic_sample_graph)
+    hook = RecencyNeighborHook(
+        num_nbrs=[1], num_nodes=2, seed_nodes_keys=['foo'], seed_times_keys=['bar']
+    )
+    batch = dg.materialize()
+    batch.foo = 'should_be_1d_tensor'
+    batch.bar = 'should_be_1d_tensor'
+
+    with pytest.raises(ValueError):
+        batch = hook(dg, batch)
 
 
-@pytest.mark.skip('TODO')
-def test_bad_sample_with_non_tensor_non_None_seeds():
-    pass
+def test_bad_sample_with_non_1d_tensor_seeds(basic_sample_graph):
+    dg = DGraph(basic_sample_graph)
+    hook = RecencyNeighborHook(
+        num_nbrs=[1], num_nodes=2, seed_nodes_keys=['foo'], seed_times_keys=['bar']
+    )
+    batch = dg.materialize()
+    batch.foo = torch.rand(2, 3)  # should be 1-d
+    batch.bar = torch.rand(2, 3)  # should be 1-d
+
+    with pytest.raises(ValueError):
+        batch = hook(dg, batch)
 
 
-@pytest.mark.skip('TODO')
-def test_bad_sample_with_non_1d_tensor_seeds():
-    pass
+def test_bad_sample_with_seeds_id_out_of_range(basic_sample_graph):
+    dg = DGraph(basic_sample_graph)
+    hook = RecencyNeighborHook(
+        num_nbrs=[1], num_nodes=2, seed_nodes_keys=['foo'], seed_times_keys=['bar']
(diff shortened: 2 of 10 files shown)


examples/nodeproppred/tgn.py
```diff
@@ -409,7 +409,8 @@ def eval(
 nbr_hook = RecencyNeighborHook(
     num_nbrs=args.n_nbrs,
     num_nodes=test_dg.num_nodes,  # Assuming node ids at test set > train/val set
-    seed_nodes_key='node_ids',
+    seed_nodes_keys=['node_ids'],
+    seed_times_keys=['node_times'],
```
shenyangHuang, line 423: good idea to add this, though we can also infer it (as the system is hard coded for now)

tgm/hooks/hooks.py
```diff
@@ -200,55 +203,78 @@ class NeighborSamplerHook(StatelessHook):
...
+    ) -> None:
         if not len(num_nbrs):
             raise ValueError('num_nbrs must be non-empty')
         if not all([isinstance(x, int) and (x > 0) for x in num_nbrs]):
             raise ValueError('Each value in num_nbrs must be a positive integer')
         self._num_nbrs = num_nbrs
         self._directed = directed
 
+        if seed_nodes_keys is None and seed_times_keys is None:
+            self._seed_nodes_keys = ['src', 'dst', 'neg']
```
shenyangHuang: why not as default argument in the method

tgm/hooks/hooks.py
```diff
@@ -271,6 +297,45 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+    def _get_seed_tensors(self, batch: DGBatch) -> Tuple[torch.Tensor, torch.Tensor]:
+        device = batch.src.device
+        seeds, times = [], []
+
+        for node_attr, time_attr in zip(self._seed_nodes_keys, self._seed_times_keys):
+            missing = [
+                attr for attr in (node_attr, time_attr) if not hasattr(batch, attr)
+            ]
+            if missing:
+                warnings.warn(
```
shenyangHuang: this is going to warn for every single batch?

shenyangHuang: might as well just raise error at this point

shenyangHuang (commented):
default keys maybe?

examples/nodeproppred/tgn.py
Jacob-Chmura: Technically, yes, but it seems much more error prone, worse to implement and harder to change in the future

tgm/hooks/hooks.py
Jacob-Chmura: Never ever use lists as default argument in python, they are mutable. By the same logic [don't use any mutable object as default argument](https://web.archive.org/web/20200221224620/http://effbot.org/zone/default-values.htm)

tgm/hooks/hooks.py
Jacob-Chmura: See the warnings filter I added on top of the module. It only gets emitted once. We want to be very conservative with runtime errrors inside the main loop since it could really annoy people: imagine an uncaught exception 2 hours into a training loop

tgm/hooks/hooks.py
```diff
@@ -271,6 +298,64 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+            seed = getattr(batch, node_attr)
+            time = getattr(batch, time_attr)
+
+            for name, tensor in [(node_attr, seed), (time_attr, time)]:
+                # We recover from tensor = None, since the current batch could just
+                # be missing certain attributes (e.g. dynamic node events), but for
+                # non-Tensor and non-None attrs we explicitly raise
+                if tensor is None:
+                    warnings.warn(
+                        f'Seed attribute {name} is None on this batch, skipping',
```
Jacob-Chmura: Consider raising exception here

tgm/hooks/hooks.py
```diff
@@ -271,6 +298,64 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
...
+                    )
+                    break
+                if not isinstance(tensor, torch.Tensor):
+                    raise ValueError(f'{name} must be a Tensor, got {type(tensor)}')
+                if tensor.ndim != 1:
+                    raise ValueError(f'{name} must be 1-D, got shape {tensor.shape}')
+
+                # Bounds checks
+                # TODO: Infer self._num_nodes from underlying graph
+                self._num_nodes = float('inf')
```
Jacob-Chmura, line 354: Left this here to make it clear how to adapt in case of API unification between samplers.

tgm/hooks/hooks.py
```diff
@@ -409,6 +523,63 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
             self._update(batch)
         return batch
 
+    def _get_seed_tensors(self, batch: DGBatch) -> Tuple[torch.Tensor, torch.Tensor]:
+        device = batch.src.device
+        seeds, times = [], []
+
+        # Implicitly add negatives to seed nodes and query times if they exist
+        seed_nodes_keys, seed_times_keys = self._seed_nodes_keys, self._seed_times_keys
+        if hasattr(batch, 'neg'):
```
shenyangHuang: why do we still need to manually append neg here? shouldn't we now let the user to specific if they want neg or not? they can tell from what neg sampler produces, same with neg_time

shenyangHuang: I think it is more intuitive for users to add it themselves, rather than we magically also include neg as seed_nodes

test/unit/test_hooks/test_recency_nbr_hook.py
```diff
@@ -149,6 +149,115 @@ def test_bad_neighbor_sampler_init():
...
+
+    with pytest.raises(ValueError):
+        batch = hook(dg, batch)
+
+    batch.foo = torch.IntTensor([3])  # cannot be > num_nodes
+    with pytest.raises(ValueError):
+        batch = hook(dg, batch)
+
+
+def test_bad_sample_with_seeds_time_out_of_range(basic_sample_graph):
```
shenyangHuang, line 252: we need a test case with recency and neg sampler

PR #287, opened 2025-10-01
Add structured logging

### LINKED ISSUE #251: Properly handle measurement during examples

Jacob-Chmura:
Close #251


examples/graphproppred/persistant_forecast.py
```diff
@@ -171,8 +169,10 @@ def eval(
 
 val_labels = generate_binary_trend_labels(val_loader, snapshot_measurement=edge_count)
 val_results = eval(val_loader, val_labels, model, val_metrics)
-logger.info(' '.join(f'{k}={v:.4f}' for k, v in val_results.items()))
+for metric_name, metric_value in val_results.items():
+    log_metric(metric_name, metric_value)
```
Jacob-Chmura: In the future (or now) we can create a wrapper called `log_metrics_dict` which does exactly this.

examples/linkproppred/dygformer.py
```diff
@@ -313,13 +312,14 @@ def eval(
...
+    log_metric(f'Validation {METRIC_TGB_LINKPROPPRED}', val_mrr, epoch=epoch)
+
     # Clear memory state between epochs, except last epoch
     if epoch < args.epochs:
         hm.reset_state()
 
 with hm.activate(test_key):
     test_mrr = eval(evaluator, test_loader, model, static_node_feat)
-    logger.info(f'Test MRR:{METRIC_TGB_LINKPROPPRED}={test_mrr:.4f}')
+log_metric(f'Test {METRIC_TGB_LINKPROPPRED}', test_mrr, epoch=args.epochs)
```
Jacob-Chmura, line 325: Here is the convention. Assuming all future examples can do this, integration tests should be good to parse nicely

examples/nodeproppred/tgn.py
```diff
@@ -455,17 +454,17 @@ def eval(
...
-    logger.info(
-        f'Epoch={epoch:02d} Loss={loss:.4f} Train {METRIC_TGB_NODEPROPPRED}={train_metric:.4f}'
-    )
 
     with hm.activate('val'):
         val_metric = eval(val_loader, memory, encoder, decoder, evaluator)
-    logger.info(f'Validation {METRIC_TGB_NODEPROPPRED}={val_metric:.4f}')
+
+    log_metric('Loss', loss, epoch=epoch)
+    log_metric(f'Train {METRIC_TGB_NODEPROPPRED}', train_metric, epoch=epoch)
```
Jacob-Chmura, line 462: We should unify all examples. Here we also run train ndcg. But everywhere else, we only report train loss.

tgm/util/logging.py
Jacob-Chmura, line 1: Should deprecate the profilers in `util/perf` now. User won't need it. The only thing to keep there would be the `cprofiler` since that's something quite different from this.

examples/graphproppred/persistant_forecast.py
```diff
@@ -171,8 +168,8 @@ def eval(
 
 val_labels = generate_binary_trend_labels(val_loader, snapshot_measurement=edge_count)
 val_results = eval(val_loader, val_labels, model, val_metrics)
-logger.info(' '.join(f'{k}={v:.4f}' for k, v in val_results.items()))
+log_metrics_dict(val_results)
```
shenyangHuang, line 171: nice and clean!

examples/graphproppred/persistant_forecast.py
```diff
@@ -171,8 +168,8 @@ def eval(
...
 
 test_labels = generate_binary_trend_labels(test_loader, snapshot_measurement=edge_count)
 test_results = eval(test_loader, test_labels, model, test_metrics)
-logger.info(' '.join(f'{k}={v:.4f}' for k, v in test_results.items()))
+log_metrics_dict(test_results)
```
shenyangHuang, line 175: does the metrics_dict know the above is val and next one is test, do we need
`log_metrics_dict("test", test_results)`, what if a model skips validation step for some reason?

examples/linkproppred/dygformer.py
shenyangHuang: yeah this looks good, why is this syntax different from the other scripts?

examples/graphproppred/persistant_forecast.py
shenyangHuang: shouldn't this say log_metrics as well like the other examples

examples/nodeproppred/tgn.py
shenyangHuang: I think its okay. this is only because the evaluation of link prop pred (mrr ranking) is drastically different than its training (binary classification)

shenyangHuang: we could drop it but it is also beneficial to track it

shenyangHuang (approved):

PR #291, opened 2025-10-02
Test Coverage

### LINKED ISSUE #290: Get our test coverage up

Jacob-Chmura:
Note: waiting on model migrations to add testing to nn

Close #290

### DIFF
--- a/tgm/util/logging.py
+++ b/tgm/util/logging.py
@@ -220,7 +220,7 @@ def log_metric(
 
     Args:
         metric_name (str): Name of the metric to log.
-        metric_value (Any): Value fo the metric to log.
+        metric_value (Any): Value of the metric to log.
         epoch (Optional[int]): Optional epoch number.
         level (int): Logging level for human-readable log (default INFO)
         extra (Dict[str, Any]): Optional dictionary of extra metadata to include in JSON.


tgm/graph.py
Jacob-Chmura, line 1: There was an infinite recursion in case batch attr was string since strings are technically conside `Iterable`

tgm/data.py
```diff
@@ -663,29 +647,13 @@ def from_pandas(
             ImportError: If the Pandas package is not resolved in the current python environment.
         """
 
-        def _check_pandas_import(min_version_number: str | None = None) -> None:
```
Jacob-Chmura, line 666: We weren't using the version check, so simplifying this for now

tgm/data.py
```diff
@@ -154,11 +154,6 @@ def _maybe_cast_integral_tensor(x: Tensor, name: str) -> Tensor:
         # Validate edge event idx
         _assert_is_tensor(self.edge_event_idx, 'edge_event_idx')
         _assert_tensor_is_integral(self.edge_event_idx, 'edge_event_idx')
-        if self.edge_event_idx.ndim != 1 or self.edge_event_idx.shape[0] != num_edges:
```
Jacob-Chmura, line 157: These are all internally constructed so we should only be validating (and raising) incorrect user input.

tgm/hooks/hook_manager.py
```diff
@@ -59,14 +59,11 @@ def _stringify_hook(h: DGHook) -> str:
             lines.append(_stringify_hook(h))
 
         lines.append(f'  Active key: {self._active_key}')
-        if self._key_to_hooks:
```
Jacob-Chmura, line 62: This else clause is impossible since we now force keys to be defined on construction

tgm/hooks/hooks.py
```diff
@@ -20,11 +20,12 @@ class PinMemoryHook(StatelessHook):
     """Pin all tensors in the DGBatch to page-locked memory for faster async CPU-GPU transfers."""
 
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
-        pin_if_needed = (
```
Jacob-Chmura, line 23: user should have cuda setup if running pin memory, however, this safeguards us and makes the code portable
- e.g. you have code running on machine with gpu, then you try running it on machine with only cpu

Jacob-Chmura: Then you can still use pin memory hook (though it doesnt do anything)

tgm/split.py
```diff
@@ -49,8 +46,6 @@ def _masked_copy(
         if data.node_ids is not None:
             if node_mask is None:
                 node_mask = torch.ones(data.node_ids.shape[0], dtype=torch.bool)
-            if node_mask.dtype != torch.bool:
```
Jacob-Chmura, line 52: Again, these are internally constructed, we already validated the user input so this is impossible.

tgm/util/logging.py
```diff
@@ -1,12 +1,18 @@
...
 import json
 import logging
+import os
 import time
 from pathlib import Path
 from typing import Any, Callable, Dict, List
 
 import torch
 
+_TGM_LOGGING_ENABLED: bool = os.getenv('TGM_LOGGING_ENABLED', '0').lower() in (
```
Jacob-Chmura, line 11: This will be nice since we should be able to just use this in CI

tools/profiling.py
Jacob-Chmura, line 1: Out of core and into dev tools

test/unit/test_data.py
```diff
@@ -303,6 +303,30 @@ def test_init_dg_data_bad_args_empty_graph():
...
+            edge_index,
+            node_ids=node_ids,
+            node_timestamps=node_timestamps,
+        )
+
+
+def test_init_dg_Data_bad_static_node_feats():
+    edge_index = torch.IntTensor([[2, 3], [10, 20]])
+    edge_timestamps = torch.LongTensor([0, 1])
+    static_node_feats = torch.rand(0)  # should be 2d
```
shenyangHuang, line 323: `torch.empty(num_nodes, 0)` would be fine?

tgm/hooks/hooks.py
shenyangHuang: sounds good

tgm/util/logging.py
shenyangHuang: indeed.

shenyangHuang (approved):

test/unit/test_data.py
Jacob-Chmura: yes technically, in which case `statc_node_feat_dim` would just return 0 everywhere (not None)

PR #311, opened 2025-10-06
Nice Log Interpolator

### LINKED ISSUE #310: Human Format Log Out
add commas for integers < 1e6, use k/m/b suffix for larger numbers, handles float('inf') and Nan

Jacob-Chmura:
Close #310

### DIFF
--- a/test/unit/test_util/test_logging.py
+++ b/test/unit/test_util/test_logging.py
@@ -228,6 +228,17 @@ def test_log_metrics_dict_console_and_file(capsys):
         assert json_lines[1] == {'metric': 'bar', 'value': 234, 'epoch': 1}
 
 
+def test_human_format():
+    assert tgm_logging.human_format(None) == 'None'
+    assert tgm_logging.human_format(123) == '123'
+    assert tgm_logging.human_format(123.45) == '123.45'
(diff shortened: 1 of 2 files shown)


tgm/data/split.py
```diff
@@ -7,7 +7,7 @@
 import torch
 from torch import Tensor
 
-from tgm.util.logging import _get_logger
+from tgm.util.logging import _get_logger, human_format
```
Jacob-Chmura: Technically, we can just inject a new LogRecord formatter into our root tgm logger. The only problem with this is that we have to enforce _all_ numeric values are logged as `str`. So for now, I'm keeping simple and explicitly calling `human_format` where needed (our library is small enough that this is fine for now)

shenyangHuang: yeah sounds good, also human_format feels a bit skynet to me (its fine).

shenyangHuang: if we use `human_format` a lot, why not make the name shorter and easier to remember, like `hf`? or actually just say `nice_format` instead of `human_format`.

shenyangHuang (approved):
nice, a prettier name

PR #318, opened 2025-10-07
Patch Workflow Input

Jacob-Chmura:


.github/workflows/integration.yml
```diff
@@ -4,9 +4,15 @@ on: # yamllint disable-line rule:truthy
   schedule:
     - cron: "30 6 * * 1"  # every Monday at 6:30 AM UTC
   workflow_dispatch:
-    pr_number:
-      description: 'PR number to comment metrics on'
-      required: false
+    inputs:
+      pr_number:
```
ntgbaoo, line 8: Just for my understanding, do we use this `pr_number` anywhere?

examples/linkproppred/tpnet.py
```diff
@@ -25,6 +27,12 @@
 parser.add_argument('--dataset', type=str, default='tgbl-wiki', help='Dataset name')
 parser.add_argument('--bsize', type=int, default=200, help='batch size')
 parser.add_argument('--device', type=str, default='cpu', help='torch device')
+parser.add_argument(
+    '--max-eval-batches-per-epoch',
+    type=int,
+    default=None,
+    help='if specified, restricts the number of batches used in validation. Primarily meant to speed up integration tests',
+)
```
ntgbaoo: Not sure if it is ok to expose this to users because users should never use this. But I couldn't think about any better ways. May be we can define some kind of system flag? and retrieve it by using `os.getenv('TGM_INTEGRATION_TEST_MODE', 'FALSE')`? When running integration test, we should set this to True.

In our core, we can check for this flag, if it is True, then just return the first `n` batches. Not sure if it makes sense. Just think out loud.

examples/linkproppred/tpnet.py
```diff
@@ -207,10 +216,12 @@ def eval(
     loader: DGDataLoader,
     model: nn.Module,
     static_node_feat: torch.Tensor,
+    max_eval_batches_per_epoch: int | None,
```
ntgbaoo: Same concern here. Please see previous comments. This is something users should be aware and need to worry about when using `TGM`.

.github/workflows/integration.yml
Jacob-Chmura: It's not a required input. If specified it determines which pull request the github bot comments on.

examples/linkproppred/tpnet.py
Jacob-Chmura: Yeah I agree. Ultimately I opted for the cli argument for simplicity. Not sure env var is much better but we can change this.

examples/linkproppred/tpnet.py
Jacob-Chmura: Addressed

examples/linkproppred/tpnet.py
```diff
@@ -244,6 +250,12 @@ def eval(
             }
             perf_list.append(evaluator.eval(input_dict)[METRIC_TGB_LINKPROPPRED])
 
+        if max_eval_batches_per_epoch and batch_num >= int(max_eval_batches_per_epoch):
+            logger.warning(
+                f'Exiting evaluation prematurely since max_eval_batches_per_epoch ({max_eval_batches_per_epoch}) was reached. Reported performance may not reflect ground truth on the entire dataset.'
+            )
+            break
```
ntgbaoo, line 257: In an ideal case, we should move this logic to the core library somewhere. But we can keep it as it is for now.

Issue #241, opened 2025-09-18
TPNet memory optimzation


benjaminnNgo:
**Description**
TPNet is currently highly memory-demanding, which makes it impractical for large-scale experiments, especially on `tgl-subreddit`

PR #321, opened 2025-10-11
TPNet lightweight version

benjaminnNgo:
### Summary / Description

Adding a flag to enable lightweight `RandomProjectionModule` for `TPNet`

**Related Issues:** #241

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [x] [Integration tests](https://github.com/tgm-team/tgm/actions/runs/18422349902/job/52498272050)
- [ ] Performance tests

### DIFF
--- a/examples/linkproppred/tpnet.py
+++ b/examples/linkproppred/tpnet.py
@@ -62,7 +62,7 @@ parser.add_argument(
     '--use-matrix',
     default=True,
     action=argparse.BooleanOptionalAction,
-    help='if no-use-matrix, won't explicitly maintain the temporal walk matrices',
+    help='if no-use-matrix, will not explicitly maintain the temporal walk matrices',
 )
 parser.add_argument(
     '--concat-src-dst',


examples/linkproppred/tpnet.py
```diff
@@ -297,9 +309,11 @@ def eval(
     num_layer=args.rp_num_layers,
     time_decay_weight=args.rp_time_decay_weight,
     beginning_time=train_dg.start_time,
+    use_matrix=not args.not_use_matrix,
     enforce_dim=args.enforce_dim,
     num_edges=train_dg.num_edges,
     dim_factor=args.rp_dim_factor,
+    concat_src_dst=not args.not_concat_src_dst,
```
Jacob-Chmura: The double negation seems a bit unecessary to me, can we just have the arguments be `args.use_matrix` and `args.concat_src_dst`?

test/integration/test_tpnet.py
Jacob-Chmura, line 1: TPNet is by far the slowest integration test that we have, which is why I set TGM_CI_MAX_EVAL_BATCHES_PER_EPOCH=5

How much faster are the alternative tests? I would imagine we need to limit to 5 eval batches for all of them. Also, I'm not sure if we need all the tests here, or we can get away with just say 2 of them

examples/linkproppred/tpnet.py
ntgbaoo: Agree.

test/integration/test_tpnet.py
ntgbaoo: Yeah. I agree. I moved parts that I want to test from integration tests to unit tests and kept only one integration test for TPNet.

examples/linkproppred/tpnet.py
```diff
@@ -58,6 +59,18 @@
...
+    default=True,
+    type=strtobool,
+    help='if True, explicitly maintain the temporal walk matrices',
+)
+parser.add_argument(
+    '--concat-src-dst',
+    default=True,
+    type=strtobool,
+    help='if False, Random projection avoids concat src and dst in computation',
+)
```
Jacob-Chmura, line 72: Any reason we don't just make this a flag directly?

```python
parser.add_argument('--use-matrix', action=argparse.BooleanOptionalAction)
```

examples/linkproppred/tpnet.py
```diff
@@ -58,6 +58,18 @@
     default=10,
     help='the dim factor of random feature w.r.t. the node num',
 )
+parser.add_argument(
+    '--use-matrix',
+    default=True,
+    action=argparse.BooleanOptionalAction,
+    help='if no-use-matrix, wont explicitly maintain the temporal walk matrices',
```
Jacob-Chmura:
```suggestion
    help='if no-use-matrix, won't explicitly maintain the temporal walk matrices',
```

PR #323, opened 2025-11-02
Add mask for nbr

benjaminnNgo:
### Summary / Description

This is to facilitate the communication between hooks, because right now we let the user define the seed node themselves, the orders of neighbour information are now dynamically determined, and there is no way for downstream hooks to know the order.

**Related Issues:** #324

#### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking Change
- [x] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [ ] Integration tests
- [ ] Performance tests

### DIFF
--- a/examples/nodeproppred/dygformer.py
+++ b/examples/nodeproppred/dygformer.py
@@ -145,7 +145,6 @@ class DyGFormer_NodePrediction(nn.Module):
         src_nbr_idx = batch.seed_node_nbr_mask['src']
         dst_nbr_idx = batch.seed_node_nbr_mask['dst']
         edge_idx = torch.stack((src, dst), dim=0)
-        src.shape[0]
 
         src_dst_nbr_idx = torch.cat([src_nbr_idx, dst_nbr_idx])
         z_src, z_dst = self.encoder(
--- a/tgm/hooks/neighbors.py
+++ b/tgm/hooks/neighbors.py
@@ -90,6 +90,7 @@ class NeighborSamplerHook(StatelessHook):
             logger.debug('No seed_nodes found, appending empty hop information')
             for _ in self.num_nbrs:
                 _append_empty_hop()
+            batch.seed_node_nbr_mask = seed_node_nbr_mask  # type: ignore
             return batch
 
         for hop, num_nbrs in enumerate(self.num_nbrs):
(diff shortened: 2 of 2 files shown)


examples/nodeproppred/dygformer.py
```diff
@@ -142,16 +142,19 @@ def forward(
         nbr_nids = batch.nbr_nids[0]
         nbr_times = batch.nbr_times[0]
         nbr_feats = batch.nbr_feats[0]
+        src_nbr_idx = batch.seed_node_nbr_mask['src']
+        dst_nbr_idx = batch.seed_node_nbr_mask['dst']
         edge_idx = torch.stack((src, dst), dim=0)
-        batch_size = src.shape[0]
+        src.shape[0]
```
Jacob-Chmura:
```suggestion
```

tgm/hooks/neighbors.py
```diff
@@ -119,12 +119,17 @@ def _append_empty_hop() -> None:
             batch.nbr_times.append(nbr_times)  # type: ignore
             batch.nbr_feats.append(nbr_feats)  # type: ignore
 
+        batch.seed_node_nbr_mask = seed_node_nbr_mask  # type: ignore
```
Jacob-Chmura, line 123: If  `not seed_nodes.numel()` triggers (empty batch), then this attribute won't exist on the batch. I think it would be slightly safer to add the empty `seed_node_mask` dict in this case.

So we can just move this up

tgm/hooks/neighbors.py
```diff
@@ -166,6 +171,11 @@ def _get_seed_tensors(self, batch: DGBatch) -> Tuple[torch.Tensor, torch.Tensor]
                             f'got values in range [{tensor.min().item()}, {tensor.max().item()}]'
                         )
                     seeds.append(seed.to(device))
+                    num_seed_nodes = tensor.shape[0]
+                    seed_node_mask[name] = torch.arange(
+                        offset, offset + num_seed_nodes
+                    ).to(device)
```
Jacob-Chmura: Nit: 
```suggestion
                    seed_node_mask[name] = torch.arange(offset, offset + num_seed_nodes, device=device)
```

Jacob-Chmura (changes requested):
I like this, just some minor comments before approve and merge

Jacob-Chmura (approved):

PR #325, opened 2025-11-08
Added ability to evict out of window interaction in EdgeBank

benjaminnNgo:
### Summary / Description

Evict edges that are out of window in EdgeBank

**Related Issues:** #312

#### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking Change
- [x] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [x] Integration tests
- [ ] Performance tests

Since we need to clean up memory for everytime `update` is called. To avoid iterate over the whole memory, a min heap `_time_edge` is added to keep track of interaction and timestamp in order, enabling iterate over the edges that are out of window only.

### DIFF
--- a/test/unit/test_nn/test_edgebank.py
+++ b/test/unit/test_nn/test_edgebank.py
@@ -130,7 +130,7 @@ def test_out_of_order_update_fixed():
     assert bank.window_end == 4
     assert bank.window_ratio == WINDOW_RATIO
 
-    expected_edge_order = [(1.0, 1.0), (4.0, 5.0), (2.0, 3.0)]
+    expected_edge_order = [(4.0, 5.0), (1.0, 1.0), (2.0, 3.0)]
     expected_ts = [3, 3, 4]
     assert bank._head is not None and bank._tail is not None
     curr = bank._head
@@ -140,8 +140,16 @@ def test_out_of_order_update_fixed():
         assert curr.ts == expected_ts[count]
         count += 1
         curr = curr.right
-
     assert count == 3
+
+    curr = bank._tail
+    count = 3
+    while curr is not None:
+        count -= 1
+        assert curr.edge == expected_edge_order[count]
+        assert curr.ts == expected_ts[count]
+        curr = curr.left
+    assert count == 0
 
 
 def test_out_of_order_construct_unlimited():
--- a/tgm/nn/modules/edgebank.py
+++ b/tgm/nn/modules/edgebank.py
@@ -2,6 +2,10 @@ from typing import Any, Dict, Literal, Optional, Tuple
 
 import torch
 
+from tgm.util.logging import _get_logger
+
+logger = _get_logger(__name__)
+
 
 class _Event:
     def __init__(
@@ -84,6 +88,10 @@ class EdgeBankPredictor:
         self._head: Optional[_Event] = None
         self._tail: Optional[_Event] = None
 
+        logger.warning(
+            'EdgeBank will be slow if events are added/updated out of order.'
+        )
+
         self.update(src, dst, ts)
 
     def _clean_up(self) -> None:
@@ -129,17 +137,25 @@ class EdgeBankPredictor:
                 elif self._head is not None and self._tail is not None:
                     new_event = _Event((src_, dst_), ts_, left=None, right=None)
                     curr: _Event | None = self._tail
-                    curr = self._tail
 
+                    # This while loop should never run assuming events added in time-ascending order.
+                    # When events added out of order, time complexity would be O(n)
                     while curr is not None and ts_ < curr.ts:
                         curr = curr.left
+
                     if curr == None:
                         new_event.right = self._head
+                        if self._head is not None:
+                            self._head.left = new_event
                         self._head = new_event
                     else:
                         new_event.left = curr
-                        self._tail.right = new_event
-                        self._tail = new_event
+                        new_event.right = curr.right  # type: ignore[union-attr]
+                        if curr.right is not None:  # type: ignore[union-attr]
+                            curr.right.left = new_event  # type: ignore[union-attr]
+                        curr.right = new_event  # type: ignore[union-attr]
+                        if curr == self._tail:
+                            self._tail = new_event
 
     def __call__(
         self, query_src: torch.Tensor, query_dst: torch.Tensor


test/unit/test_nn/test_edgebank.py
```diff
@@ -48,6 +48,9 @@ def test_fixed_time_window(pos_prob):
     bank.update(torch.Tensor([7]), torch.Tensor([8]), torch.Tensor([7]))
     assert bank(torch.Tensor([7]), torch.Tensor([8])) == torch.Tensor([pos_prob])
     assert bank(torch.Tensor([4]), torch.Tensor([5])) == torch.Tensor([0])
+    assert (
```
shenyangHuang, line 51: Since now we are testing this correctly, I would add a stronger test where the new batch updated into the memory is so far in the future that all past interactions are evicted from memory. 

say all past interactions are 
`ts = torch.Tensor([1, 2, 3, 4, 5, 6])` and the new update is 

`bank.update(torch.Tensor([7]), torch.Tensor([8]), torch.Tensor([100000000]))`

I would call this test sth like `test_complete_eviction_fixed_time_window`

tgm/nn/modules/edgebank.py
```diff
@@ -65,8 +66,17 @@ def __init__(
...
 
         self.memory: Dict[Tuple[int, int], int] = {}
+        self._time_edge: List[Tuple[int, Tuple[int, int]]] = []  # min heap
+
         self.update(src, dst, ts)
 
+    def _clean_up(self) -> None:
+        """Clean up edges that are out of window in memory."""
+        while self._time_edge and self._time_edge[0][0] < self._window_start:
+            ts, edge = heapq.heappop(self._time_edge)
```
shenyangHuang: This can get expensive. O(log n) for each pop and push

tgm/nn/modules/edgebank.py
```diff
@@ -82,9 +92,20 @@ def update(self, src: torch.Tensor, dst: torch.Tensor, ts: torch.Tensor) -> None
...
+            not self._fixed_memory
+            and len(self._time_edge) > 0
+            and self._time_edge[0][0] < self._window_start
+        ):
+            self._clean_up()
+
         for src_, dst_, ts_ in zip(src, dst, ts):
             src_, dst_, ts_ = src_.item(), dst_.item(), ts_.item()
-            self.memory[(src_, dst_)] = ts_
+            if ts_ >= self._window_start:
```
shenyangHuang, line 133: I think there might be a more efficient way to do this especially we know that the edges from the dataloader will be sorted in most cases. 

Instead of the min heap, I suggest we keep a chronologically sorted array of the edges (say `arr`). Each time an update happens, we will perform a binary search to determine where the time boundary should be in this array (i.e. the index). Then we simple drop all the elements that are outside of the boundary (by making a new array). This guarantees that we only ever need to do one binary search for each update phase. In terms of dropping the edges, we simply do `arr[new_start:end]`)

shenyangHuang (dismissed):
An alternative approach to update the edges assuming most updates are sorted chronologically

tgm/nn/modules/edgebank.py
Jacob-Chmura: Agreed. And I suggest we re-use `self.memory` instead of a separate edge array to avoid keeping another edge index in memory when not strictly necessary.

Jacob-Chmura (changes requested):

tgm/nn/modules/edgebank.py
ntgbaoo: The idea of maintaining an extra array or a linked list is to efficiently update memory. Without it, we need to iterate over every single event in the memory to find which events are out of the window. Do you have any better idea?

tgm/nn/modules/edgebank.py
```diff
@@ -65,8 +80,22 @@ def __init__(
...
 
+    def _clean_up(self) -> None:
+        """Clean up edges that are out of window in memory."""
+        while self._head and self._head.ts < self._window_start:
+            curr_event = self._head
+            if self.memory[curr_event.edge] == curr_event.ts:
+                self.memory.pop(curr_event.edge)
+            self._head = curr_event.right
+            if self._head == None:
+                self._tail = None
```
ntgbaoo, line 105: Without the help of a linked list of sorted events, we will need to iterate over every single event in the memory to clean up.

tgm/nn/modules/edgebank.py
```diff
@@ -82,9 +111,35 @@ def update(self, src: torch.Tensor, dst: torch.Tensor, ts: torch.Tensor) -> None
...
+
+                    while curr is not None and ts_ < curr.ts:
+                        curr = curr.left
+                    if curr == None:
+                        new_event.right = self._head
+                        self._head = new_event
+                    else:
+                        new_event.left = curr
+                        self._tail.right = new_event
+                        self._tail = new_event
```
ntgbaoo: The operation of adding new events to the linked list will be O(1) in the case that events updated to memory follow chronological order, which is true for all datasets we have right now. The inner while loop is just for handling the case when the users update the event out of chronological order.

Jacob-Chmura (commented):
To be clear, I don't think we need to strictly optimize edgebank in the current PR, as the scope of this change should be priamrily functional. So long as the new implementation has similar latency as our previous, I think it's fine for now.

In terms of making improvements beyond this, my thought was that we use a ring-buffer based approach (similar to our recency sampler). Note, this requires that fixed-memory mode edgebank also has a max buffer capacity (which should be fine). 

The ops we need to support are:
- (a) Batch inclusion check: check whether list of events are in our buffer
- (b) Batch eviction: evict all events with timestamp < current_time + WINDOW

The idea would be to have a single contiguous tensor with a head/tail pointer to emulate a fixed length queue. Batch insertion basically does something like `buffer[tail:tail+batch_size] = new_events`. We also need to define a hash function that allows us to _fingerprint_ the edges. This enables us to efficiently query for edge inclusion by doing bit operations on the query edges. To handle evictions, we can similarly use vector ops (find valid event mask based on query time, then advance tail pointer with wraparound to). 

For this PR, let's just profile the edgebank implementaion on a few datasets, and once we confirm that it's similar performance, we can handle this optimization seperately (if we think its important to do so)

tgm/nn/modules/edgebank.py
```diff
@@ -82,9 +111,35 @@ def update(self, src: torch.Tensor, dst: torch.Tensor, ts: torch.Tensor) -> None
...
             src_, dst_, ts_ = src_.item(), dst_.item(), ts_.item()
-            self.memory[(src_, dst_)] = ts_
+            if ts_ >= self._window_start:
+                self.memory[(src_, dst_)] = ts_
+                if self._head == self._tail == None:
+                    self._head = self._tail = _Event((src_, dst_), ts_, None, None)
+                elif self._head is not None and self._tail is not None:
+                    new_event = _Event((src_, dst_), ts_, left=None, right=None)
+                    curr: _Event | None = self._tail
+                    curr = self._tail
```
Jacob-Chmura:
```suggestion
```

tgm/nn/modules/edgebank.py
Jacob-Chmura: Are we missing some pointer ops here? E.g. shouldn't we do something like:

```python
new_event.right = curr.right
if curr.right is not None:
    curr.right.left = new_event
curr.right = new_event
if curr == self._tail:
    self._tail = new_event
```

Jacob-Chmura: Understood. I am tempted to emit some sort of warning to the user, telling them that edgebank will be slow if you are pushign out of order events. But this should be fine. Given we sort the edgelist ourselves, this is better than using a BST and paying `logn` despite being sorted.

tgm/nn/modules/edgebank.py
```diff
@@ -82,9 +111,35 @@ def update(self, src: torch.Tensor, dst: torch.Tensor, ts: torch.Tensor) -> None
...
+
+                    while curr is not None and ts_ < curr.ts:
```
Jacob-Chmura, line 143: Can we add a comment here, explaining that this should never run assuming you push events in time-ascending order. And that worst case insertion to middle for linked list will be O(n)

tgm/nn/modules/edgebank.py
Jacob-Chmura: I mean, this is striclty in the case where we insert in the middle of the list due to out of order insertion.

ntgbaoo: Yeah you are right, I missed testing `_tail` and didn't catch this bug. Added a unit test for it [here](https://github.com/tgm-team/tgm/blob/312-EvictEventsOutOfWindow/test/unit/test_nn/test_edgebank.py#L145C4-L152C22)

PR #326, opened 2025-11-10
TNCN implementation

benjaminnNgo:
### Summary / Description

1. Added a version of `TNCN` in `torch` (our)
2. Moved `TGN` to `tgm.nn`

**Related Issues:** #18 

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [x] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [x] Integration tests
- [x] Performance tests

- I am not sure how to prepare a unit test for `torch_sparse` version of `TNCN`, as we don't have `torch_sparse` as a dependency on TGM
- I converted most of the operations on `torch_sparse` from the original implementation to operations on `torch`. I think this may hurt performance in terms of runtime, but the performance shouldn't be affected. 
- Not sure why the [warning message](https://github.com/tgm-team/tgm/blob/18-TNCNImplementation/tgm/nn/decoder/ncnpred.py#L16) doesn't show up in the terminal. Do you have any idea why @Jacob-Chmura

### DIFF
--- a/examples/linkproppred/tncn.py
+++ b/examples/linkproppred/tncn.py
@@ -37,6 +37,7 @@ parser.add_argument(
     '--k',
     type=int,
     default=2,
+    choices=[2, 4, 8],
     help='k-th hop common neighbour (CN) embedding extraction (select from 2/4/8)',
 )
 parser.add_argument('--lr', type=str, default=0.0001, help='learning rate')


tgm/nn/decoder/ncnpred.py
```diff
@@ -0,0 +1,609 @@
...
+
+from tgm.util.logging import _get_logger
+
+logger = _get_logger(__name__)
+
+try:
+    import torch_sparse  # noqa: F401
+
+    HAS_TORCH_SPARSE = True
+    logger.warning('Torch sparse is available. Use original implementation.')
```
Jacob-Chmura: You might not see this on the terminal because you have to first call `tgm.util.logging.enable_logging()` in order to setup the loggers. Since this is triggered on import, it's unlikely to have the logging setup.

For this reason, I've been emitting `warnings.warn` on top of the loggers (which may or may not be setup): https://github.com/tgm-team/tgm/blob/772457a20b74bc25174416feceab573d0e4f5026/tgm/data/dg_data.py#L89-L95

Jacob-Chmura:
```suggestion
```
I'd omit this warning

tgm/nn/decoder/ncnpred.py
```diff
@@ -0,0 +1,609 @@
...
+def _fill(sp: torch.Tensor, value: int | float) -> torch.Tensor:
+    return torch.sparse_coo_tensor(
+        sp.indices(),
+        torch.full((sp.values().numel(),), value),
+        sp.size(),
+        device=sp.device,
+    ).coalesce()
+
+
+if HAS_TORCH_SPARSE:
```
Jacob-Chmura: What if we just branch the `HAS_TORCH_SPARSE` control flow in the `get_cn_emb` method, rather than the entire class? I.e., the forward and init seem compatible. Perhaps even forwarding to a seperate method (e.g. `get_cn_emb_sparse` so that we minimize areas in our code where sparse is required.

tgm/nn/decoder/ncnpred.py
```diff
@@ -0,0 +1,609 @@
...
+                        xcn_1_0,
+                        xcn_1_1,
+                        xcn_1_2,
+                        xcn_2_1,
+                        xcn_2_2,
+                        special_xcn_2_2,
+                    ],
+                    dim=-1,
+                )
+                print(cn_emb)
```
Jacob-Chmura:
```suggestion
```

test/unit/test_nn/test_ncn.py
```diff
@@ -0,0 +1,27 @@
...
+    B, D, Z = 2, 5, 1
+    ncn = NCNPredictor(in_channels=D, hidden_dim=Z, out_channels=Z, k=k)
+    node_feat = torch.rand((B, D))
+    edge_index = torch.randint(0, B, size=(B, 2))
+    h = ncn(node_feat, edge_index, edge_index)
+    assert h.shape[0] == (B)
+    assert not torch.isnan(h).any()
+
+
+@pytest.mark.skipif(not HAS_TORCH_SPARSE, reason='torch_sparse required')
```
Jacob-Chmura: One path forward to testing this would be

(a) Dispatch `get_cn_emb` on torch sparse rather than entire class (as mention in my comment on the file)
(b) Make `sparse` an argument. The default should use the _try import_ logic to determine which to dispatch to. If a user explicitly asks for sparse and it's not available _then_ we triggger a warning
(c) add torch sparse as `dev` dependency (for tests)
(d) `pytest.mark.parametrize('sparse', [True, False])`

examples/linkproppred/tgn.py
```diff
@@ -22,7 +15,12 @@
 )
 from tgm.data import DGData, DGDataLoader
 from tgm.hooks import RecencyNeighborHook, RecipeRegistry
-from tgm.nn import LinkPredictor, Time2Vec
+from tgm.nn import LinkPredictor, TGNMemory
+from tgm.nn.encoder.tgn import (
```
shenyangHuang, line 19: nice clean up

tgm/nn/decoder/ncnpred.py
```diff
@@ -0,0 +1,609 @@
...
+
+try:
+    import torch_sparse  # noqa: F401
+
+    HAS_TORCH_SPARSE = True
+    logger.warning('Torch sparse is available. Use original implementation.')
+except ImportError:
+    HAS_TORCH_SPARSE = False
+    logger.warning(
+        'Torch Sparse library not found. Falling back to the NCN implementation using PyTorch, which may impact time performance. Install torch_sparse to use the optimized version.'
```
shenyangHuang: do we know how big the efficiency difference actually is?

tgm/nn/decoder/ncnpred.py
shenyangHuang: I agree that might be more concise

tgm/nn/decoder/ncnpred.py
```diff
@@ -0,0 +1,609 @@
...
+    from torch_sparse.matmul import spmm_add
+
+    class NCNPredictor(torch.nn.Module):
+        r"""An implementation of Temporal Neural Common Neighbor (TNCN).
+
+        Args:
+            in_channels (int): Number of input features.
+            out_channels (int): Number of output features.
+            hidden_dim (int): Size of each hidden embedding.
+            k (int): ...
```
shenyangHuang: what is the k? why does it only support [2,4,8]?

shenyangHuang: add documentation here

examples/linkproppred/tncn.py
```diff
@@ -0,0 +1,268 @@
...
+            ]
+        )
+
+        nbr_times = batch.nbr_times[0].flatten()[nbr_mask]
+        nbr_feats = batch.nbr_feats[0].flatten(0, -2).float()[nbr_mask]
+
+        z, last_update = memory(batch.unique_nids)
+        z = encoder(z, last_update, nbr_edge_index, nbr_times, nbr_feats)
+
+        inv_src = batch.global_to_local(batch.src)
```
shenyangHuang, line 116: later on, we can come up with the way to make these cleaner

test/unit/test_nn/test_ncn.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+import torch
+
+from tgm.nn import NCNPredictor
+from tgm.nn.decoder.ncnpred import HAS_TORCH_SPARSE
+
+
+@pytest.mark.parametrize('k', [2, 4, 8])
```
shenyangHuang, line 7: maybe also test bad argument when k is wrong or sth

tgm/nn/decoder/ncnpred.py
```diff
@@ -0,0 +1,609 @@
...
+                    sparse_sizes=(id_num, id_num),
+                )
+                .fill_value_(1.0)
+                .coalesce()
+                .to(x.device)
+            )
+            if self.k == 4:
+                adj0 = SparseTensor.eye(id_num, device=x.device)
+
+                i_0_v, i_1_v, j_0_v, j_1_v = (
```
shenyangHuang: nit, potentially consider better variable name

tgm/nn/decoder/ncnpred.py
```diff
@@ -0,0 +1,609 @@
...
+                cn_time_decay (bool): indicate whether applying decay on time,
+                time_info (Optional[Tuple[torch.Tensor, torch.Tensor]]): A tuple of last update and current time of each edge
+            """
+            xi = x[tar_ei[0]]
+            xj = x[tar_ei[1]]
+
+            xij = torch.mul(xi, xj).reshape(-1, x.size(1))
+
+            cn_emb = self.get_cn_emb(x, edge_index, tar_ei, cn_time_decay, time_info)
+            xs = torch.cat([xij, cn_emb], dim=-1)
```
shenyangHuang: note that here `cn_emb` exist outside of `x` which means technically all other models can make use of `cn_emb`.

shenyangHuang (changes requested):
thanks for the PR, left some comments

tgm/nn/decoder/ncnpred.py
ntgbaoo: The difference is negligible. Torch version takes 25.7648s (21.43it/s) for training and torch-sparse version takes 24.8126s (22.25it/s)

tgm/nn/decoder/ncnpred.py
ntgbaoo: So should we return cn_emb? I guess we may confuse users about which embedding should be used.  I guess if someone really want to explore, they can call `self.get_cn_emb`

examples/linkproppred/tncn.py
ntgbaoo: Maybe we add `inv_src ` and `inv_dst` to the batch when performing `Dedup` hook?

tgm/nn/decoder/ncnpred.py
Jacob-Chmura: To be honest, if this is the case on other datasets (can we check?) than I rather we drop the sparse version for now (avoid dealing with dependency issues https://github.com/tgm-team/tgm/pull/359). 

Then we can create an epic and release candidate for dedicated sparse support where we
- manager wheels for the right driver as in pyg
- build out the tooling to dispatch code and CI based on whether sparse support is enabled
- start writting sparse implementation at the most useful spots (e.g. neighbor sampling, message passing, NCN)

ntgbaoo: I am running both versions on tgbl-coin now.  But they both have the same number of iteration/s, so I guess that will likely be the case

examples/linkproppred/tncn.py
```diff
@@ -25,14 +25,20 @@
...
 )
 parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
 parser.add_argument('--dataset', type=str, default='tgbl-wiki', help='Dataset name')
 parser.add_argument('--bsize', type=int, default=200, help='batch size')
 parser.add_argument('--device', type=str, default='cpu', help='torch device')
 parser.add_argument('--epochs', type=int, default=30, help='number of epochs')
+parser.add_argument(
+    '--k',
+    type=int,
+    default=2,
```
Jacob-Chmura, line 39:
```suggestion
    default=2,
    choices=[2, 4, 8],
```

PR #338, opened 2025-11-19
CTAN

### LINKED ISSUE #349: CTAN Implementation

Jacob-Chmura:
Close #349

### DIFF
--- a/tgm/nn/encoder/ctan.py
+++ b/tgm/nn/encoder/ctan.py
@@ -16,7 +16,7 @@ class CTAN(torch.nn.Module):
         memory_dim (int): Dimension of memory embeddings.
         time_dim (int): Dimension of time encodings.
         node_dim (int): Dimension of static/dynamic node features.
-        num_iters(int): Number of AntiSymmetricConv layers.
+        num_iters (int): Number of AntiSymmetricConv layers.
         mean_delta_t (float): Mean delta time between edge events (used to normalize time signal).
         std_delta_t (float): Std delta time between edge events (used to normalize time signal).
         epsilon (float): Discretization step size for AntiSymmetricConv.


examples/linkproppred/ctan.py
```diff
@@ -0,0 +1,365 @@
...
+        super().__init__()
+        self.mean_delta_t = mean_delta_t
+        self.std_delta_t = std_delta_t
+        self.time_enc = TimeEncoder(time_dim)
+        self.enc_x = nn.Linear(memory_dim + node_dim, memory_dim)
+
+        phi = TransformerConv(
+            memory_dim, memory_dim, edge_dim=edge_dim + time_dim, root_weight=False
+        )
+        self.aconv = AntiSymmetricConv(memory_dim, phi, num_iters=num_iters)
```
Jacob-Chmura: The authors expose `epsilon` and `gamma` as hyperparameters whereas I'm using defaults provided by pyg

examples/linkproppred/ctan.py
```diff
@@ -0,0 +1,365 @@
...
+        out = msg.new_zeros((dim_size, msg.size(-1)))
+
+        if index.numel() > 0:
+            scores = torch.full((dim_size, t.size(0)), float('-inf'), device=t.device)
+            scores[index, torch.arange(t.size(0), device=t.device)] = t.float()
+            argmax = scores.argmax(dim=1)
+            valid = scores.max(dim=1).values > float('-inf')
+            out[valid] = msg[argmax[valid]]
+
+        return out
```
ntgbaoo: I know that we got this from TGN, and we know TGN works. To me, this logic makes sense. But our code differs from CTAN this module. May be worth replacing this with torch_scatter as the last thing to try.

examples/linkproppred/ctan.py
```diff
@@ -0,0 +1,365 @@
...
+    memory_dim=args.memory_dim,
+    time_dim=args.time_dim,
+    node_dim=static_node_feats.shape[1],
+    num_iters=args.n_layers,
+    mean_delta_t=mean_delta_t,
+    std_delta_t=std_delta_t,
+).to(args.device)
+decoder = LinkPredictor(node_dim=args.memory_dim, hidden_dim=args.memory_dim).to(
+    args.device
+)
```
ntgbaoo: Jacob, their link predictor does element-wise sum between `src` and `dst` embeddings. I believe we do `concat`. I think it won't make any difference. But it may be good to try to replicate as close as we can 
https://github.com/gravins/non-dissipative-propagation-CTDGs/blob/main/models/predictors.py#L16C1-L19C33

examples/linkproppred/ctan.py
```diff
@@ -0,0 +1,390 @@
...
+        )
+
+    def forward(self, x, last_update, edge_index, t, msg):
+        rel_t = (last_update[edge_index[0]] - t).abs()
+        rel_t = ((rel_t - self.mean_delta_t) / self.std_delta_t).to(x.dtype)
+        enc_x = self.enc_x(x)
+        edge_attr = torch.cat([msg, self.time_enc(rel_t)], dim=-1)
+        z = self.aconv(enc_x, edge_index, edge_attr=edge_attr)
+        z = torch.tanh(z)
+        return z
```
Jacob-Chmura: This is a direct port of [the original](https://github.com/gravins/non-dissipative-propagation-CTDGs/tree/main/models) as far as I can tell @benjaminnNgo 

As written, I'm seeing about 0.53 validation mrr (better than what we were getting before at 0.3), but still less than the reported 0.65 tgb number. I'm not doing early stopping etc.

ntgbaoo: So you took the same `CTAN` as the authors' implementations (not changing `torch_scatter`)? This may suggest that we did something wrong when we refactored `CTAN` before. But it is good that we can get up to 0.53 now. I think we can try 2 things:
1. You said that you did the grid search to search for the best configuration from the author's source code. Let's take the best config and train with the author's source code and see how many epochs (assume `n`) `CTAN` need to get the best validation results. Then go back to our implementation, and run for 1.5 * `n` epochs, we expect to see the best validation results achieves within 1.5 * `n` epochs should be close to `0.65`. (Why 1.5 * `n`, just trying to consider randomness makes the optimization converge slightly different)
2. If the first step happens in the way we expect, then refactor from this version (changing from torch_scatter to torch operations etc.) and make sure the performance doesn't change

Jacob-Chmura: ~0.56 Test MRR @ 100 epochs with _optimal_ hyperparams.

tgm/nn/decoder/linkproppred.py
```diff
@@ -9,8 +11,21 @@ def cat_merge(z_src: torch.Tensor, z_dst: torch.Tensor) -> torch.Tensor:
     return torch.cat([z_src, z_dst], dim=1)
 
 
-MERGE_OP = {
+class LearnableSumMerge(nn.Module):
```
Jacob-Chmura, line 14: This is distinct from doing something like `z_src + z_dst` as our merge op. Note, doing this simple aggregation could not reproducive the `CTAN` numbers we expect from the original implementation (which has the `LearnableSumMerge` decoder

tgm/nn/encoder/ctan.py
Jacob-Chmura, line 1: Down the line this should probably be unified with `TGNMemory`

tgm/data/split.py
Jacob-Chmura, line 1: Had some log formatting warnings that went through to master

tgm/nn/encoder/ctan.py
```diff
@@ -0,0 +1,106 @@
...
+
+import torch
+import torch.nn as nn
+from torch_geometric.nn import AntiSymmetricConv, TransformerConv
+from torch_geometric.nn.inits import ones, zeros
+from torch_geometric.nn.models.tgn import TimeEncoder
+from torch_geometric.utils import scatter
+
+
+class CTAN(torch.nn.Module):
```
ntgbaoo, line 11: Can we add pydoc for model the model definition like we had with other models:
https://github.com/tgm-team/tgm/blob/main/tgm/nn/encoder/gclstm.py#L12C1-L42C8

tgm/nn/encoder/ctan.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        last_update: torch.Tensor,
+        edge_index: torch.Tensor,
+        t: torch.Tensor,
+        msg: torch.Tensor,
+    ) -> torch.Tensor:
```
ntgbaoo, line 60: Good to have pydoc for this method too

examples/linkproppred/ctan.py
```diff
@@ -0,0 +1,291 @@
...
+
+    for batch in tqdm(loader):
+        nbr_nodes = batch.nbr_nids[0].flatten()
+        nbr_mask = nbr_nodes != PADDED_NODE_ID
+
+        all_nids = torch.cat(
+            [batch.edge_src, batch.edge_dst, batch.neg, nbr_nodes[nbr_mask]]
+        )
+        batch.unique_nids = torch.unique(all_nids, sorted=True)  # type: ignore
+        batch.global_to_local = lambda x: torch.searchsorted(batch.unique_nids, x)  # type: ignore
```
ntgbaoo: I think we no longer need this
```suggestion
```

examples/linkproppred/ctan.py
```diff
@@ -0,0 +1,291 @@
...
+        opt.zero_grad()
+
+        nbr_nodes = batch.nbr_nids[0].flatten()
+        nbr_mask = nbr_nodes != PADDED_NODE_ID
+
+        all_nids = torch.cat(
+            [batch.edge_src, batch.edge_dst, batch.neg, nbr_nodes[nbr_mask]]
+        )
+        batch.unique_nids = torch.unique(all_nids, sorted=True)  # type: ignore
+        batch.global_to_local = lambda x: torch.searchsorted(batch.unique_nids, x)  # type: ignore
```
ntgbaoo: I think we no longer need this
```suggestion
```

examples/linkproppred/ctan.py
```diff
@@ -0,0 +1,291 @@
...
+        src, dst, t = src.item(), dst.item(), t.item()
+
+        dt_src = t - last_timestamp.get(src, train_dg.start_time)
+        dt_dst = t - last_timestamp.get(dst, train_dg.start_time)
+        delta_times.extend([dt_src, dt_dst])
+
+        last_timestamp[src] = t
+        last_timestamp[dst] = t
+
+    return np.mean(delta_times), np.std(delta_times)
```
ntgbaoo, line 220: nit: I prefer moving this up before:

```python
seed_everything(args.seed)
```

examples/linkproppred/ctan.py
Jacob-Chmura: Good catch

PR #340, opened 2025-11-20
TGB SEQ io

### LINKED ISSUE #341: TGB SEQ IO integration

Jacob-Chmura:
Close #341

### DIFF
--- a/tgm/data/dg_data.py
+++ b/tgm/data/dg_data.py
@@ -13,7 +13,7 @@ import torch
 from torch import Tensor
 
 from tgm.constants import PADDED_NODE_ID
-from tgm.core import TGB_TIME_DELTAS, TimeDeltaDG
+from tgm.core import TGB_SEQ_TIME_DELTAS, TGB_TIME_DELTAS, TimeDeltaDG
 from tgm.data.split import SplitStrategy, TemporalRatioSplit, TGBSplit
 from tgm.exceptions import (
     EmptyGraphError,
@@ -888,6 +888,7 @@ class DGData:
             split_bounds[split_name] = (int(times.min()), int(times.max()))
 
         data = cls.from_raw(
+            time_delta=TGB_SEQ_TIME_DELTAS[name],
             edge_timestamps=timestamps,
             edge_index=edge_index,
             edge_feats=edge_feats,
--- a/test/unit/test_core/test_timedelta.py
+++ b/test/unit/test_core/test_timedelta.py
@@ -1,7 +1,7 @@
 import pytest
 
 from tgm import TimeDeltaDG
-from tgm.core.timedelta import TGB_TIME_DELTAS
+from tgm.core.timedelta import TGB_SEQ_TIME_DELTAS, TGB_TIME_DELTAS
 from tgm.exceptions import EventOrderedConversionError
 
 
@@ -189,3 +189,17 @@ def test_tgb_native_time_deltas():
         'tgbn-token': TimeDeltaDG('s'),
     }
     assert TGB_TIME_DELTAS == exp_dict
+
+
+def test_tgb_seq_native_time_deltas():
+    exp_dict = {
+        'ML-20M': TimeDeltaDG('s'),
+        'Taobao': TimeDeltaDG('s'),
+        'Yelp': TimeDeltaDG('s'),
+        'GoogleLocal': TimeDeltaDG('s'),
+        'Flickr': TimeDeltaDG('s'),
+        'Youtube': TimeDeltaDG('s'),
+        'Patent': TimeDeltaDG('s'),
+        'WikiLink': TimeDeltaDG('s'),
+    }
+    assert TGB_SEQ_TIME_DELTAS == exp_dict
--- a/test/unit/test_data/test_data.py
+++ b/test/unit/test_data/test_data.py
@@ -1164,6 +1164,9 @@ def test_from_bad_tgb_seq_name():
 
 @pytest.mark.parametrize('with_node_feats', [True, False])
 @pytest.mark.parametrize('with_edge_feats', [True, False])
+@patch.dict(
+    'tgm.core.timedelta.TGB_SEQ_TIME_DELTAS', {'tgb-seq-mock': TimeDeltaDG('s')}
+)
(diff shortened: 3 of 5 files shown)


examples/tgb_seq/edgebank.py
```diff
@@ -0,0 +1,158 @@
...
+            y_pred_neg.append(y_pred[1:])
+
+        model.update(batch.src, batch.dst, batch.time)
+
+    y_pred_pos, y_pred_neg = torch.stack(y_pred_pos), torch.stack(y_pred_neg)
+    mrr_list = evaluator.eval(y_pred_pos, y_pred_neg)
+    return float(np.mean(mrr_list))
+
+
+class TGBSEQ_NegativeEdgeSamplerHook(StatelessHook):
```
Jacob-Chmura, line 66: Not clear if we want to move to core `hooks` at the moment.

tgm/data/dg_data.py
```diff
@@ -828,6 +828,73 @@ def from_tgb(cls, name: str, **kwargs: Any) -> DGData:
...
+        split_bounds = {}
+        for split_name, mask in {
+            'train': data.train_mask,
+            'val': data.val_mask,
+            'test': data.test_mask,
+        }.items():
+            times = data.node_interact_times[mask]
+            split_bounds[split_name] = (int(times.min()), int(times.max()))
+
+        data = cls.from_raw(
```
Jacob-Chmura, line 890: Note: this currently only provides `Ordered` time granularity support. Would have to investigate the native time deltas on each seq datasets

examples/tgb_seq/edgebank.py
Jacob-Chmura, line 1: I intentionally removed TGAT/TGN examples as they directly duplicate our link pred logic. The added maintenance did not seem worth the benefits. Keeping an edgebank example seems sufficient for now.

Jacob-Chmura: Note, I preferred having a seperate example directory for TGB SEQ since other we would have to
- figure out a uniform way to encode dataset names for tgb seq vs. tgb
- include some extra logic for e.g. using the right negative sampler

This approach keeps each example lightweight.

test/integration/test_edgebank.py
Jacob-Chmura, line 1: Note: I did not create a script to pre-download the `tgb-seq` dataset in CI. Unlike TGB, which stores datasets in the `.venv`, our custom data path means the download does not happen on each run.

examples/tgb_seq/edgebank.py
```diff
@@ -0,0 +1,158 @@
...
+            query_dst = torch.cat([batch.dst[idx].unsqueeze(0), batch.neg[idx]])
+            y_pred = model(query_src, query_dst)
+            y_pred_pos.append(y_pred[0])
+            y_pred_neg.append(y_pred[1:])
+
+        model.update(batch.src, batch.dst, batch.time)
+
+    y_pred_pos, y_pred_neg = torch.stack(y_pred_pos), torch.stack(y_pred_neg)
+    mrr_list = evaluator.eval(y_pred_pos, y_pred_neg)
+    return float(np.mean(mrr_list))
```
ntgbaoo: Can we expect this `eval` function to be the same as the ones in the [existing edgebank for `TGB`](https://github.com/tgm-team/tgm/blob/main/examples/linkproppred/edgebank.py#L41C1-L61C37)?

The main difference that I can tell that `evaluator.eval()` is called once rather than called for every batch. Why is there a difference?

examples/tgb_seq/edgebank.py
ntgbaoo: Not now. But potentially yes, since we claim we support `TGB-seq`

examples/tgb_seq/edgebank.py
ntgbaoo: Yeah I agree to keep it like this for now. If we want to adapt existing examples to make them work for both `TGB` and `TGB-seq`, we may need to add 2 recipes that do the setup for each setting. But I think it is not worth for now.

tgm/data/dg_data.py
ntgbaoo: I think this is pretty important. Maybe we need to hold on to this PR and check with the authors?

examples/tgb_seq/edgebank.py
Jacob-Chmura: Good catch, this is a mistake. The authors follow the existing approach in their [mrr specific eval function](https://github.com/TGB-Seq/TGB-Seq/blob/master/examples/evaluate_models_utils_mrr.py). Will update

tgm/data/dg_data.py
Jacob-Chmura: I checked their [file preprocessing](https://github.com/TGB-Seq/TGB-Seq/blob/master/tgb_seq/datasets/preprocess.py) in a bit more detail. I also loaded and checked each dataset. Each uses Unix epoch and so second-wise granularity is appropriate.

Note, while they encode everything in unix epoch, the actual resolution need not be. For instance on `Patent` they do: 

```python
t_obj = datetime.strptime(str(t), '%Y%m%d')
unix_timestamp = int(dt_obj.timestamp())
data.append([int(u), int(i), int(unix_timestamp)])
```

Notice they convert to daily datetime, and then grab the unix time from that day (implicitly at midnight i.e. time = 00:00:000).

In any case, using native unit `s` seems the right approach. I'll update

examples/tgb_seq/edgebank.py
```diff
@@ -0,0 +1,156 @@
...
+            batch.neg = self.negs[self.neg_idx : self.neg_idx + batch_size]  # type: ignore
+            self.neg_idx += batch_size
+        else:
+            size = (self.num_negs, batch_size)
+            batch.neg = torch.randint(  # type: ignore
+                self.low, self.high, size, dtype=torch.int32, device=dg.device
+            )
+
+        # TODO: decide whether to keep this (similar to random negative sampler), or sample
+        # negative time stamps (similar to TGB negative sampler)
```
ntgbaoo, line 105: I think it makes sense to have `TGB_seq` share the existing [`NegativeEdgeSamplerHook`](https://github.com/tgm-team/tgm/blob/main/tgm/hooks/negatives.py) to generate negative edges for training.

PR #347, opened 2025-12-02
Graph Download Compatible with bash <4

### LINKED ISSUE #329: Download Script for Graph Prop Task Breaks in Older Bash version
**Describe the bug**
- graph property prediction examples all require a data download from the custom link, we should not require this
- when running `./scripts/download_graph_prop_datasets.sh examples/graphproppred` on macbook run into the following error

```
./scripts/download_graph_prop_datasets.sh: line 8: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
```
That error means the shell doesn’t support associative arrays (declare -A), which are only available in Bash (version 4+), not in sh or older Bash versions.

**To Reproduce**
running `./scripts/download_graph_prop_datasets.sh examples/graphproppred` on bash version older than 4+

**Expected behavior**
Data download correctly
- ideally graph prop task doesn't require a custom download

**Environment:**

- OS: Mac 
- TGM Version: beta 
- CPU

Jacob-Chmura:
Close #329

### DIFF
--- a/scripts/download_graph_prop_datasets.sh
+++ b/scripts/download_graph_prop_datasets.sh
@@ -5,8 +5,18 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 PROJECT_ROOT="$SCRIPT_DIR/.."
 cd "$PROJECT_ROOT"
 
-declare -A DATASETS_URLS
-DATASETS_URLS[tokens_data]="https://raw.githubusercontent.com/Jacob-Chmura/mint-datasets/master/fetch.sh"
+
+DATASETS=("tokens_data")
+
+get_dataset_url() {
+    case "$1" in
+        tokens_data)
+            echo "https://raw.githubusercontent.com/Jacob-Chmura/mint-datasets/master/fetch.sh"
+            ;;
+        *)
+            ;;  # return empty for unknown dataset
+    esac
+}
 
 print_usage() {
     echo "Usage: $0 DATA_ROOT"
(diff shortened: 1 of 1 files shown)


scripts/download_graph_prop_datasets.sh
```diff
@@ -36,17 +46,49 @@ parse_args() {
...
+        if [ -e "$path" ] || [ -L "$path" ]; then
+            realpath "$path"
+            return
+        fi
+    fi
+    # fallback: absolute path using pwd
+    # https://apple.stackexchange.com/questions/450035/is-the-unix-realpath-command-distributed-with-macos-ventura
+    echo "$PWD/$path"
+}
+
```
Jacob-Chmura, line 74: Note, `realpath` is not available until macOS 13. @shenyangHuang which version are you on? Can you test this to ensure the fallback works reasonably well? 

The reason we need this is that if I issue: `./scripts/download_graph_prop_datasets.sh .` I end up _attempting_ to move the downloaded data into the current working directory, and then `rm -rf` the original download folder. But, if these two match (i.e. when the `DATA_DIR` resolves to the current working directory) this will just delete the directory, leaving you without data. 

The check bellow attempts to resolve the src/dst paths, and skips the `mv && rm -rf` on the original download directory if they match. 

To test, please issue:

```sh
./scripts/download_graph_prop_datasets.sh . # download to current path, data should be there
./scripts/download_graph_prop_datasets.sh . # download to current path, should be skipped
./scripts/download_graph_prop_datasets.sh ./ # download to current path, should be skipped
rm -rf tokens_data # delete downloaded data
./scripts/download_graph_prop_datasets.sh ./ # download to current path, data should be there
rm -rf tokens_data # clean up
```

ntgbaoo: @Jacob-Chmura , I tried it out on macOS 13, Bash 3.2. Got the error:
```bash
bash scripts/download_graph_prop_datasets.sh ./
Storing datasets at: ./
Downloading dataset: tokens_data from https://raw.githubusercontent.com/Jacob-Chmura/mint-datasets/master/fetch.sh
bash: line 51: mapfile: command not found
ERROR: Failed to fetch TOKENS from https://raw.githubusercontent.com/Jacob-Chmura/mint-datasets/master/TOKENS
(base) ngtbaoo@ROBOPOPO tgm % curl: (23) Failed writing body
```

Seems like mapfile is not available on Bash 3.2

Jacob-Chmura: Ah, this is coming upstream from my hosted repo script: https://github.com/Jacob-Chmura/mint-datasets/blob/master/fetch.sh

Will need to patch and cut a new build

Jacob-Chmura: @benjaminnNgo Can you try again now? When you run it you should see `v0.2` asset download [here](https://github.com/Jacob-Chmura/mint-datasets/blob/master/VERSION). I replaced `mapfile` in [this commit](https://github.com/Jacob-Chmura/mint-datasets/pull/4/files)

ntgbaoo: @Jacob-Chmura, now we got the issue from `wget`

```bash
(base) ngtbaoo@ROBOPOPO tgm % bash scripts/download_graph_prop_datasets.sh ./
Storing datasets at: ./
Downloading dataset: tokens_data from https://raw.githubusercontent.com/Jacob-Chmura/mint-datasets/master/fetch.sh
[*] Fetching all MiNT token datasets v0.2 ...
[*] Downloading token test-token from https://github.com/Jacob-Chmura/mint-datasets/releases/download/v0.2...
bash: line 76: wget: command not found
ERROR: Failed to download test-token from https://github.com/Jacob-Chmura/mint-datasets/releases/download/v0.2 to ./tokens_data/test-token.zip
```

Jacob-Chmura: @benjaminnNgo Can you try now, I [replaced `wget` with `curl`](https://github.com/Jacob-Chmura/mint-datasets/commit/926c6795bed6b86f200da7395ca1c70d1de6746d)

ntgbaoo: @Jacob-Chmura , it works. Thank you!

PR #350, opened 2025-12-03
Mark Raw Event Hook Dependencies

### LINKED ISSUE #248: Create proper abstraction for node/event requirements on each hook

Jacob-Chmura:
Close #248 

Note: this is not strictly enforcable at hook registration time by the manager because asserting that each batch has (e.g. node events or edge events) can only be inferred during data loading (e.g. depends on batch size).

### DIFF
--- a/tgm/hooks/negatives.py
+++ b/tgm/hooks/negatives.py
@@ -160,7 +160,7 @@ class TGBTHGNegativeEdgeSamplerHook(StatelessHook):
         ValueError: If neg_sampler is not provided.
     """
 
-    requires: Set[str] = set()
+    requires = {'src', 'dst', 'time', 'edge_type'}
     produces = {'neg', 'neg_batch_list', 'neg_time'}
 
     def __init__(
--- a/test/unit/test_hooks/test_batch_analytics_hook.py
+++ b/test/unit/test_hooks/test_batch_analytics_hook.py
@@ -33,7 +33,7 @@ def dg():
 
 def test_hook_dependancies():
     hook = BatchAnalyticsHook()
-    assert hook.requires == set()
+    assert hook.requires == {'src', 'dst', 'time', 'node_times', 'node_ids'}
     assert hook.produces == {
         'num_edge_events',
         'num_node_events',
--- a/test/unit/test_hooks/test_node_analytics_hook.py
+++ b/test/unit/test_hooks/test_node_analytics_hook.py
@@ -243,7 +243,7 @@ def test_node_analytics_hook_produces_and_requires():
     tracked_nodes = torch.tensor([0])
     hook = NodeAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=5)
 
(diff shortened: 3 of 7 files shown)


tgm/hooks/neighbors.py
ntgbaoo, line 1: It is a bit tricky here to define what attributes are required for these neighbour sampler hooks. Since users can define what seed nodes and seed times are, `requires` is dynamically defined in this case

Jacob-Chmura: Indeed, it's a bit of a _runtime_declaration_ and there is not much we can assert until iteration time.

However, `batch.src, batch.dst, batch.time` are all required since we access them while updating our buffers:

https://github.com/tgm-team/tgm/blob/d5d177872e94f8a4dffb4a8e3ba8bc6b8136c23d/tgm/hooks/neighbors.py#L489-L491

Jacob-Chmura: We could in theory append the requires attributes to the class variable but that would mess things up in case there are multiple instances of the hook :/

Alternatively, we can make `requries/produces` an instance variable. Thoughts? @benjaminnNgo

ntgbaoo: For the first point: yeah, I don't think we should do that.
Makes `requires/produces` as instance variables makes sense. Just trying to think if it affects anything down the line.

Jacob-Chmura: Not directly related, but this makes me also think we should improve our error handling here:

https://github.com/tgm-team/tgm/blob/d5d177872e94f8a4dffb4a8e3ba8bc6b8136c23d/tgm/hooks/hook_manager.py#L142-L146

In particular, any `AttributeError` that is thrown on `DGBatch` should be wrapped in a `UnresolvableHookDependenciesError`. We should meet as a team and think through these hook ergonomics in more detail

Jacob-Chmura: > Just trying to think if it affects anything down the line.

The think that comes to mind is that we want to still enforce that these these requirements are "static" for the duration of the object... I think...  Any dependencies that change during graph iteration cannot reasonably be guarded against (node event / edge event missing included).

tgm/hooks/negatives.py
ntgbaoo, line 1: @Jacob-Chmura, can you pull udpate from `main` and add `requires` for [TGBTHGNegativeEdgeSamplerHook](https://github.com/tgm-team/tgm/blob/main/tgm/hooks/negatives.py#L148). The requires would be `src`, `dst`, `time` and `edge_type`

benjaminnNgo (approved):
Looks good to me. But before merging, you may need to pull updates from main to get a new hook for `thgl` negative sampler 

We also have these new hooks: `tgm/hooks/node_analytics.py`; `tgm/hooks/batch_analytics.py`. But I think they don't really require anything.

Approved to unblock.

PR #351, opened 2025-12-06
Added ROLAND

benjaminnNgo:
### Summary / Description

Added ROLAND for linkproppred

**Related Issues:** #344 

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [x] Documentation update

#### Test Evidence

Describe how this PR has been tested.

- [x] Unit tests
- [x] Integration tests
- [ ] Performance tests

#### Questions / Discussion Points
N/A

PR #354, opened 2025-12-21
Poptrack

### LINKED ISSUE #353: PopTrack Implementation
**Description**
Implement and add [PopTrack](https://openreview.net/pdf?id=9kLDrE5rsW) to the TGM example suite.

**Motivation**
Broader motivation to enable studying heuristic-based models using TGM.

ekmpa:
### Summary / Description

Implementation of PopTrack and integration into the TGM example suite. 

**Related Issues:**
Close #353 

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [x] Integration tests
- [ ] Performance tests

#### Questions

- TBD: whether it is better to leave out the decay parameter search, compromise a few % of accuracy but making the model running consistent with other examples. Will determine when I add the example into the integration test harness.

### DIFF
--- a/examples/linkproppred/poptrack.py
+++ b/examples/linkproppred/poptrack.py
@@ -70,7 +70,7 @@ def eval(
                 'eval_metric': [METRIC_TGB_LINKPROPPRED],
             }
             perf_list.append(evaluator.eval(input_dict)[METRIC_TGB_LINKPROPPRED])
-        model.update(batch.src, batch.dst, batch.time, decay)
+        model.update(batch.src, batch.dst, batch.time)
 
     return float(np.mean(perf_list))
 
@@ -82,7 +82,7 @@ train_data, val_data, test_data = DGData.from_tgb(args.dataset).split()
 train_dg = DGraph(train_data)
 val_dg = DGraph(val_data)
 test_dg = DGraph(test_data)
-num_nodes = train_dg.num_nodes + val_dg.num_nodes + test_dg.num_nodes
+num_nodes = test_dg.num_nodes
 
 train_data = train_dg.materialize(materialize_features=False)
 
--- a/tgm/nn/modules/poptrack.py
+++ b/tgm/nn/modules/poptrack.py
@@ -20,9 +20,8 @@ class PopTrackPredictor:
 
         This predictor implements the PopTrack baseline for dynamic link prediction,
         introduced in `https://openreview.net/pdf?id=9kLDrE5rsW`.
-        It caches the k most popular nodes, and predicts the probability of a link
-        reoccurring based on whether the queried edge leads to one of these k-most
-        popular nodes.
+        It predicts the probability of a link reoccurring based on
+        the popularity score of the queried edge's destination.
 
         Args:
             src (torch.Tensor): Source node IDs of edges used for initialization.
@@ -45,31 +44,28 @@ class PopTrackPredictor:
         if 0 >= k:
             raise ValueError('K must be positive')
 
-        if decay < 0 or decay > 1:
+        if decay <= 0 or decay > 1:
             raise ValueError('Decay must be in (0,1]')
 
-        if num_nodes < (len(src) + len(dst)):
+        if num_nodes <= 0:
             raise ValueError('``num_nodes`` must be set to the total number of nodes.')
 
         if k > num_nodes:
-            k = num_nodes
+            raise ValueError('``k`` must be smaller than ``num_nodes``.')
 
         self._check_input_data(src, dst, ts)
         self.popularity = torch.zeros(num_nodes)
         self.k = k
-        self.top_k: torch.Tensor = torch.zeros(self.k)
-        self.update(src, dst, ts, decay)
+        self.decay = decay
+        self.update(src, dst, ts)
 
-    def update(
-        self, src: torch.Tensor, dst: torch.Tensor, ts: torch.Tensor, decay: float = 0.9
-    ) -> None:
+    def update(self, src: torch.Tensor, dst: torch.Tensor, ts: torch.Tensor) -> None:
         """Update PopTrack cache with a batch of edges.
 
         Args:
             src (torch.Tensor): Source node IDs of the edges.
             dst (torch.Tensor): Destination node IDs of the edges.
             ts (torch.Tensor): Timestamps of the edges.
-            decay (float, optional): Decay for popularity.
 
         Raises:
             TypeError: If inputs are not ``torch.Tensor``.
@@ -79,9 +75,7 @@ class PopTrackPredictor:
         self.popularity.index_add_(
             0, dst, torch.ones_like(dst, dtype=self.popularity.dtype)
         )
-        self.popularity *= decay
-        top_k_idx, values = torch.topk(self.popularity, self.k, largest=True)
-        self.top_k = top_k_idx
+        self.popularity *= self.decay
 
     def __call__(
         self, query_src: torch.Tensor, query_dst: torch.Tensor
--- a/test/unit/test_nn/test_poptrack.py
+++ b/test/unit/test_nn/test_poptrack.py
@@ -20,7 +20,6 @@ def test_poptrack_update(decay):
         torch.tensor([1]),
         torch.tensor([1]),
         torch.tensor([7.0]),
-        decay=decay,
     )
 
     assert torch.allclose(
@@ -38,7 +37,6 @@ def test_init_valid_input():
     assert isinstance(model.popularity, torch.Tensor)
     assert model.popularity.shape == (4,)
     assert model.k == 2
-    assert len(model.top_k) == 2
 
 
 def test_bad_init_args():
@@ -48,13 +46,16 @@ def test_bad_init_args():
         )
 
     with pytest.raises(TypeError):
-        PopTrackPredictor('1', '2', '3', num_nodes=2)
+        PopTrackPredictor('1', '2', '3', num_nodes=2, k=1)
 
     src = torch.tensor([0, 1])
     dst = torch.tensor([2, 3])
     ts = torch.tensor([1, 2])
     with pytest.raises(ValueError):
         PopTrackPredictor(src, dst, ts, num_nodes=4, k=-5)
+
+    with pytest.raises(ValueError):
+        PopTrackPredictor(src, dst, ts, num_nodes=4, k=10)
 
     with pytest.raises(ValueError):
         PopTrackPredictor(src, dst, ts, num_nodes=4, decay=-0.5)
@@ -70,7 +71,7 @@ def test_bad_update_args():
     src = torch.tensor([0, 1])
     dst = torch.tensor([2, 3])
     ts = torch.tensor([1, 2])
-    model = PopTrackPredictor(src, dst, ts, num_nodes=4)
+    model = PopTrackPredictor(src, dst, ts, num_nodes=4, k=2)
 
     with pytest.raises(ValueError):
         model.update(torch.Tensor([]), torch.Tensor([]), torch.Tensor([1]))


examples/linkproppred/poptrack.py
```diff
@@ -0,0 +1,106 @@
...
+test_dg = DGraph(test_data)
+num_nodes = train_dg.num_nodes + val_dg.num_nodes + test_dg.num_nodes
+
+train_data = train_dg.materialize(materialize_features=False)
+
+hm = HookManager(keys=['val', 'test'])
+hm.register('val', TGBNegativeEdgeSamplerHook(args.dataset, split_mode='val'))
+hm.register('test', TGBNegativeEdgeSamplerHook(args.dataset, split_mode='test'))
+
+train_loader = DGDataLoader(train_dg, args.bsize)
```
ntgbaoo:
```suggestion
```

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,117 @@
...
+                or if they are empty.
+
+        Note: the predictions are not conditional on the source.
+        """
+        if 0 >= k:
+            raise ValueError(f'K must be positive')
+
+        self._check_input_data(src, dst, ts)
+
+        self.popularity = np.zeros(num_nodes)
```
ntgbaoo: Can we make it a tensor instead of a numpy? We can vectorize the update operation (as suggested below) use `dst` tensor directly without a loop over it, and retrieve int value of each.

```suggestion
        self.popularity = torch.zeros(num_nodes)
```

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,117 @@
...
+            ValueError: If input tensors do not have the same length, or are empty.
+        """
+        self._check_input_data(src, dst, ts)
+        for dst_ in dst:
+            dst_ = dst_.item()
+            self.popularity[dst_] += 1
+        self.popularity *= decay
+        top_k_idx = np.argpartition(self.popularity, -self.k)[-self.k :]
+        top_k_idx = top_k_idx[np.argsort(self.popularity[top_k_idx])[::-1]]
+        self.top_k = top_k_idx
```
ntgbaoo:
```suggestion
        top_k_idx, values = torch.topk(self.popularity,self.k,largest=True)
        self.top_k = top_k_idx
```

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,117 @@
...
+            decay (float, optional): Decay for popularity.
+
+        Raises:
+            TypeError: If inputs are not ``torch.Tensor``.
+            ValueError: If input tensors do not have the same length, or are empty.
+        """
+        self._check_input_data(src, dst, ts)
+        for dst_ in dst:
+            dst_ = dst_.item()
+            self.popularity[dst_] += 1
```
ntgbaoo: If `self.popularity` is a tensor, we can vectorize this operation as follow:
```suggestion
        self.popularity.index_add_(0, dst, torch.ones_like(dst, dtype=self.popularity.dtype))
```

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        if 0 >= k:
+            raise ValueError(f'K must be positive')
+
+        self._check_input_data(src, dst, ts)
+
+        self.popularity = np.zeros(num_nodes)
+        self.k = k
+        self.top_k: np.ndarray = np.zeros(self.k, dtype=int)
+        self.pos_prob = pos_prob
```
ntgbaoo: We currently don't use `self.pos_prob`?

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,117 @@
...
+                - An edge's probability is the popularity value of
+                    its destination node (= original implementation)
+                - Otherwise, the probability is ``0.0``.
+        """
+        pred = torch.zeros_like(query_src)
+        src_list = query_src.tolist()
+        dst_list = query_dst.tolist()
+        for i, (_, d) in enumerate(zip(src_list, dst_list)):
+            # if d in self.top_k:
+            pred[i] = self.popularity[d]
```
ntgbaoo: Do we need `sigmoid` here?
In the case of a node with an in-degree of 100, we multiply by a `decay` of 0.9. We then get a popularity score of 90? (correct me if I am wrong). Then we return 90 as a prediction for every pair having that node as `dst`?

ntgbaoo:
```suggestion
        pred = self.popularity[query_dst]
```

ekmpa: First off, thank you for the great suggestions, am re-working the code. This is something I've thought of and used in implementing Base3 which uses popularity as inspired by PopTrack, but I'm unsure about changing it directly in PopTrack, because their original implementation uses raw popularity scores. But I will test it, see if it improves results, if so we could document that as an improvement. From my understanding it's unnecessary because across batches, a frequently interacted node will accumulate a *relatively* large score, i.e ranking still works.

ekmpa: Another question is whether to reproduce the original code's parameter search on decay. It seems useful for Poptrack to adapt to different datasets, but I'm unsure whether it would make evaluation non-comparable with others who just use a default eval?

ntgbaoo: ah I see, you are right. I think it makes sense now.

ntgbaoo: I think we want to keep the examples simple and just to demonstrate how to run the methods. But I agree that this is a good point to let anyone who wants to use `Base3` know, you can leave a comment in the example file (underneath declaration of `init_decays`) or below [this note](https://github.com/tgm-team/tgm/blob/120d2641809018da254f8a8839d661af2a60ce6f/tgm/nn/modules/poptrack.py#L45C1-L46C1). In the comment, you can mention the values of `decay` that you suggest to perform parameter search on.

ekmpa: Will do! By the way, I am opening a separate PR for `Base3`, it is partially based on `PopTrack` but is a separate method

ekmpa: Confirming the above, I just tested and applying sigmoid keeps relative ranking the same => doesn't change performance at all on some datasets, and even seems to clip too aggressively in some cases and worsens performance by a few % on other datasets.

ntgbaoo: Yeah, I think that makes sense, since sigmoid doesn't really take into consideration the maximum of `self.popularity`. One way to bring prediction score to 0 and 1 without affecting the performance by normalizing with the maximum in `self.popularity`  `pred/torch.max(self.popularity)` . But I think it is not very important. It is fine the return `pred` as it is right now

ntgbaoo: Oh sorry for mixing `Base3` and `PopTrack`, for some reason, I thought this was `Base3`

ekmpa: That makes sense, yes

tgm/nn/poptrack.py
ntgbaoo, line 1: Is this another method? I think we should make a separate PR for this

benjaminnNgo:
image

btw, Format and Linting check failed against. Can you try
```
uv run ruff format .
```
before push one more time?

examples/linkproppred/poptrack.py
```diff
@@ -0,0 +1,111 @@
...
+            query_dst = torch.cat([batch.dst[idx].unsqueeze(0), neg_batch])
+
+            y_pred = model(query_src, query_dst)
+            input_dict = {
+                'y_pred_pos': y_pred[0],
+                'y_pred_neg': y_pred[1:],
+                'eval_metric': [METRIC_TGB_LINKPROPPRED],
+            }
+            perf_list.append(evaluator.eval(input_dict)[METRIC_TGB_LINKPROPPRED])
+        model.update(batch.src, batch.dst, batch.time, decay)
```
Jacob-Chmura: Nit: if decay is static throughout training perhaps it makes sense to pass it in on model init rather than on forward

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,115 @@
...
+            ValueError: If ``src``, ``dst``, and ``ts`` do not have the same length,
+                or if they are empty.
+
+        Note:
+            - The predictions are not conditional on the source.
+        """
+        if 0 >= k:
+            raise ValueError('K must be positive')
+
+        if decay < 0 or decay > 1:
```
Jacob-Chmura:
```suggestion
        if decay <= 0 or decay > 1:
```

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,115 @@
...
+            raise ValueError('Decay must be in (0,1]')
+
+        if num_nodes < (len(src) + len(dst)):
```
Jacob-Chmura: We should be careful, this need not hold, in general. `src` and `dst` are essentially our edge index, which may (are likely to) contain duplicate node ids. So comparing on `len()`  could break.

I'd personally just drop this check, instead, we can `assert num_nodes > 0`

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,115 @@
...
+            TypeError: If inputs are not ``torch.Tensor``.
+            ValueError: If input tensors do not have the same length, or are empty.
+        """
+        self._check_input_data(src, dst, ts)
+        self.popularity.index_add_(
+            0, dst, torch.ones_like(dst, dtype=self.popularity.dtype)
+        )
+        self.popularity *= decay
+        top_k_idx, values = torch.topk(self.popularity, self.k, largest=True)
+        self.top_k = top_k_idx
```
Jacob-Chmura: Nit: 
```suggestion
        self.top_k, _ = torch.topk(self.popularity, self.k, largest=True)
```

Jacob-Chmura: Actually, it seems that we never actually _read_ `self.top_k`? In which case we could just remove this altogether, and just update `self.popularity` here.

Jacob-Chmura (changes requested):
Just some minor comments, thanks for the contribution. I'm not super certain I can easily run the integration tests from a fork, but after we merge upstream I can trigger them on master. In case anything breaks it can be patched from there.

Could you also update the readme [here](https://github.com/tgm-team/tgm/blob/772457a20b74bc25174416feceab573d0e4f5026/README.md?plain=1#L42) to show `PopTrack` as an implemented method in TGM?  We forgot to do this for `ROLAND`  in https://github.com/tgm-team/tgm/pull/351, so if you can add `ROLAND` as implemented too that would be great!

examples/linkproppred/poptrack.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+seed_everything(args.seed)
+evaluator = Evaluator(name=args.dataset)
+
+train_data, val_data, test_data = DGData.from_tgb(args.dataset).split()
+train_dg = DGraph(train_data)
+val_dg = DGraph(val_data)
+test_dg = DGraph(test_data)
+num_nodes = train_dg.num_nodes + val_dg.num_nodes + test_dg.num_nodes
```
shenyangHuang: You just need `test_dg.num_nodes`. This here will create around 3 times more num_nodes than needed. 
see [here](https://github.com/tgm-team/tgm/blob/772457a20b74bc25174416feceab573d0e4f5026/examples/linkproppred/tgn.py#L391)

tgm/nn/modules/poptrack.py
shenyangHuang: yes, we never actually use `self.top_k`

tgm/nn/modules/poptrack.py
shenyangHuang: On this one, I think we should be enforcing `assert num_nodes > k`. later on we have 
```
if k > num_nodes:
            k = num_nodes
```
I don't think this is clear behavior to the user, this is better if we let the user know explicitly that the `k` should never be larger than `num_nodes` and let the user manually adjust `k <= num_nodes`.

tgm/nn/modules/poptrack.py
shenyangHuang: @ekmpa  Is there a variant of Poptrack that uses `top_k`? or maybe `top_k` can be used as a feature for some other model?

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,115 @@
...
+            k = num_nodes
+
+        self._check_input_data(src, dst, ts)
+        self.popularity = torch.zeros(num_nodes)
+        self.k = k
+        self.top_k: torch.Tensor = torch.zeros(self.k)
+        self.update(src, dst, ts, decay)
+
+    def update(
+        self, src: torch.Tensor, dst: torch.Tensor, ts: torch.Tensor, decay: float = 0.9
```
shenyangHuang: note on `decay` changes at inference time. Technically  you can adjust the `decay` factor on the fly and this might be natural at test time, considering potential distribution shift on temporal graph. However, this act more as a hyperparameter that the user can tune. For example, you can tune the decay factor on the validation set and then use the same for test set.

shenyangHuang: My personal take on this is that, this should not be part of the update function, instead, we have two options:
1. keep only decay factor at initialization. 
2. have decay at initialization but also have a dedicated setter and getter function to change the decay factor if the user wants to (this is more principled in my opinion)

tgm/nn/modules/poptrack.py
```diff
@@ -0,0 +1,115 @@
...
+        k: int = 50,
+        decay: float = 0.9,
+    ) -> None:
+        """PopTrack Predictor.
+
+        Reference: https://openreview.net/pdf?id=9kLDrE5rsW
+
+        This predictor implements the PopTrack baseline for dynamic link prediction,
+        introduced in `https://openreview.net/pdf?id=9kLDrE5rsW`.
+        It caches the k most popular nodes, and predicts the probability of a link
```
shenyangHuang: This is misleading as we never actually returns top_k and it sounds like the cache size is based on `k` and not `num_nodes`, which is false.

examples/linkproppred/poptrack.py
ekmpa: True, changed

tgm/nn/modules/poptrack.py
ekmpa: Yeah, I'm removing it, though it might be useful for negative sampling. Essentially in the paper, they only use top_k to do popularity-based negative sampling (their other contribution alongside the model).

Jacob-Chmura (approved):

Issue #355, opened 2025-12-21
Base3 Implementation


ekmpa:
**Description**
Implement and add [Base3](https://www.arxiv.org/abs/2506.12764) to the TGM example suite.

**Motivation**
Broader motivation to enable studying heuristic-based models using TGM.

PR #358, opened 2025-12-24
Base3

ekmpa:
### Summary / Description

Implementation of [Base3](https://www.arxiv.org/abs/2506.12764) & integration to the TGM example suite.

**Motivation**
Broader motivation to enable studying heuristic-based models using TGM. 

**Related Issues:** #355

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [x] Integration tests
- [ ] Performance tests

### Notes
Ordering issue on EdgeBank submodule TBC.

### DIFF
--- a/tgm/nn/modules/t_comem.py
+++ b/tgm/nn/modules/t_comem.py
@@ -29,9 +29,9 @@ class tCoMemPredictor:
             dst (torch.Tensor): Destination node IDs of edges used for initialization.
             ts (torch.Tensor): Timestamps of edges used for initialization.
             num_nodes (int): Total number of nodes in the dataset.
-            k (int, optional): threshold for popularity effect. 
-                Defaults to 50, must be positive and smaller than ``num_nodes``. 
-                In general, larger ``k`` leads to better performance but higher memory usage, 
+            k (int, optional): threshold for popularity effect.
+                Defaults to 50, must be positive and smaller than ``num_nodes``.
+                In general, larger ``k`` leads to better performance but higher memory usage,
                 though this usually stops being true after a certain point.
             window_ratio (float, optional): Ratio of the sliding window length to
                 the total observed time span (only used if ``memory_mode='fixed'``).


tgm/nn/modules/t_comem.py
Jacob-Chmura, line 1: Can we add a doc entry for this module in this file: https://github.com/tgm-team/tgm/blob/main/docs/api/nn.md?

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,220 @@
...
+
+        Args:
+            query_src (torch.Tensor): Source node IDs of the query edges.
+            query_dst (torch.Tensor): Destination node IDs of the query edges.
+
+        Returns:
+            torch.Tensor: Predictions of shape ``(len(query_src),)``, where:
+                - If ...,
+                  its probability is ``...``.
+                - Otherwise, the probability is ``0.0``.
```
Jacob-Chmura: I presume this is still left to fill out?

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        pred = torch.zeros_like(query_src)
+
+        unique_src, inv = torch.unique(query_src, return_inverse=True)
+        base_scores = torch.zeros(unique_src.size(0))
+
+        window_start = self._window_start
+        window_end = self._window_end
+        window_size = self._window_size
```
Jacob-Chmura: Nit: I'd probably just inline these to use the `self` attributes

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,220 @@
...
+        Returns:
+            torch.Tensor: Predictions of shape ``(len(query_src),)``, where:
+                - If ...,
+                  its probability is ``...``.
+                - Otherwise, the probability is ``0.0``.
+        """
+        pred = torch.zeros_like(query_src)
+
+        unique_src, inv = torch.unique(query_src, return_inverse=True)
+        base_scores = torch.zeros(unique_src.size(0))
```
Jacob-Chmura: I may be wrong but this init seems unused as we overrite later anyway

```suggestion
```

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,220 @@
...
+            self.recent_dst[s, pos] = d
+            self.recent_pos[s] = (pos + 1) % self.k
+            if self.recent_len[s] < self.k:
+                self.recent_len[s] += 1
+            self.node_to_co_occurrence[s][d] = (
+                self.node_to_co_occurrence[s].get(d, 0) + 1
+            )
+            self.node_to_co_occurrence[d][s] = (
+                self.node_to_co_occurrence[d].get(s, 0) + 1
+            )
```
Jacob-Chmura, line 125: Minor, but technically we could get away with only storing one of these directions and save 50% of our space? This would just require that we use some convention of co-occurence key (e.g. smaller node id is first node in the edge key).

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,220 @@
...
+
+        pos_idx = torch.arange(self.k, device=ts_mat.device)
+        valid_mask = pos_idx.unsqueeze(0) < len_vec.unsqueeze(1)
+
+        time_mask = (ts_mat >= window_start) & (ts_mat <= window_end)
+        mask = valid_mask & time_mask
+
+        ts_valid = torch.where(
+            mask, ts_mat, torch.tensor(-float('inf'), device=ts_mat.device)
+        )
```
Jacob-Chmura:
```suggestion
        ts_valid = torch.where(
            mask, ts_mat, torch.full_like(ts_mat, -float('inf'))
        )
```

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,220 @@
...
+        nbr_valid = torch.where(mask, nbr_mat, torch.zeros_like(nbr_mat))
+
+        decay_vals = torch.exp(-(window_end - ts_valid) / window_size)
+        pop_vals = torch.sigmoid(self.popularity[nbr_valid])
+
+        base_scores = (decay_vals * pop_vals * mask).sum(dim=1)
+        pred = base_scores[inv].clone()
+        co_vals = torch.zeros(len(query_src), device=query_src.device)
```
Jacob-Chmura:
```suggestion
        co_vals = torch.zeros_like(query_src)
```

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,220 @@
...
+
+        # co-occurrence signal
+        for i, (s, d) in enumerate(zip(query_src, query_dst)):
+            s = s.item()
+            d = d.item()
+            c = self.node_to_co_occurrence.get(s, {}).get(d, 0)
+
+            if c:
+                co_vals[i] = self.co_occurrence_weight * (c / (1 + c))
```
Jacob-Chmura: Nit, this seems equivalent:
```suggestion
            c = self.node_to_co_occurrence.get(s, {}).get(d, 0)
            co_vals[i] = self.co_occurrence_weight * (c / (1 + c))
```

tgm/nn/modules/t_comem.py
ekmpa: that makes sense, I've just pushed a new version without this change (but everything else from your review - thank you btw), but I will look into this

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,213 @@
...
+        This predictor implements the t-CoMem module for dynamic link prediction,
+        introduced in `https://www.arxiv.org/abs/2506.12764`.
+        It is a memory-based module that mixes popularity with co-occurence.
+
+        Args:
+            src (torch.Tensor): Source node IDs of edges used for initialization.
+            dst (torch.Tensor): Destination node IDs of edges used for initialization.
+            ts (torch.Tensor): Timestamps of edges used for initialization.
+            num_nodes (int): Total number of nodes in the dataset.
+            k (int, optional): threshold for popularity effect. Defaults to 50.
```
shenyangHuang: this should also be dataset specific right, or the user might want to tune this as hyperparameter, maybe mention it explicitly here.

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,213 @@
...
+from collections import defaultdict, deque
+import torch
+import numpy as np
+import math
+
+
+class tCoMemPredictor:
+    def __init__(
+        self,
+        src: torch.Tensor,
```
shenyangHuang, line 11: with the `update` function, technically you can just start empty right and then update later?

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,213 @@
...
+        Args:
+            src (torch.Tensor): Source node IDs of edges used for initialization.
+            dst (torch.Tensor): Destination node IDs of edges used for initialization.
+            ts (torch.Tensor): Timestamps of edges used for initialization.
+            num_nodes (int): Total number of nodes in the dataset.
+            k (int, optional): threshold for popularity effect. Defaults to 50.
+            window_ratio (float, optional): Ratio of the sliding window length to
+                the total observed time span (only used if ``memory_mode='fixed'``).
+                Must be in ``(0, 1]``. Defaults to ``0.15``.
+            co_occurrence_weight (float, optional): Weighting parameter for co-occurrence.
```
shenyangHuang, line 39: hyperparameter, not intuitive how one should tune this though? any guidelines?

tgm/nn/modules/t_comem.py
shenyangHuang: this is technically true for edgebank as well, it is fine to leave it as is for now

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,213 @@
...
+        Args:
+            src (torch.Tensor): Source node IDs of the edges.
+            dst (torch.Tensor): Destination node IDs of the edges.
+            ts (torch.Tensor): Timestamps of the edges.
+
+        Raises:
+            TypeError: If inputs are not ``torch.Tensor``.
+            ValueError: If input tensors do not have the same length, or are empty.
+        """
+        self._check_input_data(src, dst, ts)
```
shenyangHuang, line 107: we do the same thing in Edgebank and potentially other layers, maybe this can be a helper function in tgm to check if an input tuple is in the correct format? @Jacob-Chmura @ekmpa

shenyangHuang:
@Jacob-Chmura @ekmpa @benjaminnNgo  (As discussed a few times before) maybe in a future update or PR, we can change `edgebank`, `poptrack` and `t_comem` all into stateful hooks. Since they are all non-gradient based, they are essentially data transformations, also making an interesting research question where their output can be used as input features to a TGNN, i.e. poptrack output can be feed into TGN or simply a transformer provide positional encoding. This discussion however is outside of the scope of this PR.

tgm/nn/modules/t_comem.py
ntgbaoo: Potentially base class (#302) should have this function?

tgm/nn/modules/t_comem.py
```diff
@@ -0,0 +1,216 @@
...
+            self.recent_dst[s, pos] = d
+            self.recent_pos[s] = (pos + 1) % self.k
+            if self.recent_len[s] < self.k:
+                self.recent_len[s] += 1
+            self.node_to_co_occurrence[s][d] = (
+                self.node_to_co_occurrence[s].get(d, 0) + 1
+            )
+            self.node_to_co_occurrence[d][s] = (
+                self.node_to_co_occurrence[d].get(s, 0) + 1
+            )
```
ntgbaoo, line 125: It is not a big deal for now, we could try to vectorize this in the future. But it is not worth trying for now but we should think about it, especially on large-scale networks.

tgm/nn/modules/t_comem.py
Jacob-Chmura: My opinion is that `EdgeBank` and `Base3` can (should?) eventually be written as stateful hooks. The class hierarchy there could facilitate code sharing. But for now, I'm happy to duplicate this minor piece of code.

Issue #301, opened 2025-10-04
Node Centric Stat Hooks


Jacob-Chmura:
These should admit the same api as the seed nodes sampler. We compute these stats online for each seed node
Pass thorugh statinator to get moving avg etc

- node degree
- node activity (proprtion of time steps a node is present)
- node engagemetn: avg number of distinct nodes that establish at least one connection
- node novelty: how many nodes in this timestamp are new
- entropy: node activity entropy across batches
- node lifetime: last_seen[node] - first_seen[node]
- time sicne last seen: t - prev_seen[node]

we can also do these edge ones
- edge lifetime t - prev_seen[src, dst]
- edge novelty: how many edge sin this timestmamp ar enew

also consider edge density: (edges this batch vs. possible edges this batch based on unique nodes)

PR #361, opened 2025-12-25
 Node Centric Stat Hooks

xuanvietchu:
### Summary / Description

Added Node Centric Stat Hooks, example, and tests. Stats are computed with numpy for optimization. 
Produces:
        node_stats (Dict[int, Dict[str, float]]): Dictionary mapping node_id to statistics:
            - degree: Number of edges connected to the node in the current batch.
            - activity: Fraction of unique timesteps in which the node has appeared.
            - new_neighbors: Number of new neighbors in which the node encountered in the current batch.
            - lifetime: Time since the node was first seen.
            - time_since_last_seen: Time since the node was last seen.
            - appearances: Total number of unique timesteps the node has appeared in.
        node_macro_stats (Dict): Batch-level node statistics:
            - node_novelty: Fraction of tracked nodes in the batch that are appearing for the first time.
            - new_node_count: Number of tracked nodes in the batch that are appearing for the first time.
        edge_stats (Dict): Batch-level edge statistics:
            - edge_novelty: Fraction of new edges in the batch, that is not seen in previous batches.
            - edge_density: Edges this batch / possible edges based on unique nodes
            - new_edge_count: Number of new edges in the batch, that is not seen in previous batches.

**Related Issues:** #301 

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [x] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [ ] Integration tests
- [ ] Performance tests

### DIFF
--- a/examples/analytics/node_analytics_example.py
+++ b/examples/analytics/node_analytics_example.py
@@ -9,7 +9,7 @@ from tqdm import tqdm
 from tgm import DGraph
 from tgm.data import DGData, DGDataLoader
 from tgm.hooks import HookManager
-from tgm.hooks.node_analytics import NodeCentricAnalyticsHook
+from tgm.hooks.node_analytics import NodeAnalyticsHook
 from tgm.util.logging import enable_logging, log_latency, log_metrics_dict
 from tgm.util.seed import seed_everything
 
@@ -21,7 +21,7 @@ parser.add_argument('--seed', type=int, default=1337, help='random seed to use')
 parser.add_argument('--dataset', type=str, default='tgbn-trade', help='Dataset name')
 parser.add_argument('--bsize', type=int, default=200, help='batch size')
 parser.add_argument(
-    '--num-tracked', type=int, default=500, help='Number of nodes to track'
+    '--num-tracked', type=int, default=1, help='Number of nodes to track'
 )
 parser.add_argument(
     '--log-file-path', type=str, default=None, help='Optional path to write logs'
@@ -65,7 +65,7 @@ def run_node_analytics(
     dg: DGraph, bsize: int, tracked_nodes: torch.Tensor
 ) -> tuple[dict, list]:
     """Run node analytics hook and collect statistics."""
-    node_analytics_hook = NodeCentricAnalyticsHook(
+    node_analytics_hook = NodeAnalyticsHook(
         tracked_nodes=tracked_nodes, num_nodes=dg.num_nodes
     )
 
--- a/tgm/hooks/node_analytics.py
+++ b/tgm/hooks/node_analytics.py
@@ -11,7 +11,7 @@ from tgm.util.logging import _get_logger
 logger = _get_logger(__name__)
 
 
-class NodeCentricAnalyticsHook(StatefulHook):
+class NodeAnalyticsHook(StatefulHook):
     """Compute node-centric statistics for a specific set of tracked nodes.
 
     This hook maintains state across batches to compute temporal statistics
--- a/docs/api/hooks/hooks.md
+++ b/docs/api/hooks/hooks.md
@@ -13,4 +13,4 @@
 ::: tgm.hooks.NegativeEdgeSamplerHook
 ::: tgm.hooks.RecencyNeighborHook
 ::: tgm.hooks.BatchAnalyticsHook
-::: tgm.hooks.NodeCentricAnalyticsHook
+::: tgm.hooks.NodeAnalyticsHook
--- a/test/unit/test_hooks/test_node_analytics_hook.py
+++ b/test/unit/test_hooks/test_node_analytics_hook.py
@@ -3,7 +3,7 @@ import torch
 
 from tgm import DGBatch, DGraph
 from tgm.data import DGData
-from tgm.hooks.node_analytics import NodeCentricAnalyticsHook
+from tgm.hooks.node_analytics import NodeAnalyticsHook
 
 
 @pytest.fixture
@@ -20,25 +20,25 @@ def simple_dgraph():
 
 
 def test_node_analytics_hook_initialization():
-    """Test that NodeCentricAnalyticsHook initializes correctly."""
+    """Test that NodeAnalyticsHook initializes correctly."""
     tracked_nodes = torch.tensor([0, 1, 2])
-    hook = NodeCentricAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=5)
+    hook = NodeAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=5)
 
     assert hook.num_nodes == 5
 
 
 def test_node_analytics_hook_initialization_errors():
-    """Test that NodeCentricAnalyticsHook raises errors for invalid inputs."""
+    """Test that NodeAnalyticsHook raises errors for invalid inputs."""
     tracked_nodes = torch.tensor([0, 1, 2])
 
     with pytest.raises(ValueError, match='num_nodes must be positive'):
-        NodeCentricAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=-1)
+        NodeAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=-1)
 
 
 def test_node_analytics_hook_basic_stats(simple_dgraph):
     """Test basic statistics computation."""
     tracked_nodes = torch.tensor([0, 1, 2])
-    hook = NodeCentricAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=5)
+    hook = NodeAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=5)
 
     # Materialize first batch (2 edges at time 0)
     batch1 = simple_dgraph.slice_events(0, 2).materialize()
@@ -68,7 +68,7 @@ def test_node_analytics_hook_basic_stats(simple_dgraph):
 def test_node_analytics_hook_degree_computation(simple_dgraph):
     """Test degree computation."""
     tracked_nodes = torch.tensor([0, 1, 2])
-    hook = NodeCentricAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=5)
+    hook = NodeAnalyticsHook(tracked_nodes=tracked_nodes, num_nodes=5)
     # First batch has edges: (0->1), (0->2)
     # Node 0 has degree 2 (2 outgoing)
     # Node 1 has degree 1 (1 incoming)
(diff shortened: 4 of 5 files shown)


tgm/hooks/node_analytics.py
Jacob-Chmura, line 1: Can we update our documentation [here](https://github.com/tgm-team/tgm/blob/main/docs/api/hooks/hooks.md) to include `NodeAnalayticsHook` and `BatchAnalayticsHook`?

This will then automatically render the info here in our docs.

tgm/hooks/node_analytics.py
```diff
@@ -0,0 +1,349 @@
...
+            - new_edge_count: Number of new edges in the batch, that is not seen in previous batches.
+    """
+
+    requires: Set[str] = set()
+    produces = {'node_stats', 'node_macro_stats', 'edge_stats'}
+
+    def __init__(
+        self,
+        tracked_nodes: Tensor,
+        num_nodes: int,
```
Jacob-Chmura: Nit:
```suggestion
        num_nodes: int
```

tgm/hooks/node_analytics.py
```diff
@@ -0,0 +1,349 @@
...
+
+        # Concatenate src and dst to count all edge connections
+        edge_nodes = torch.cat([batch.src, batch.dst], dim=0)
+
+        # Vectorized degree computation
+        degrees = {}
+        nodes_np = nodes.cpu().numpy() if nodes.is_cuda else nodes.numpy()
+        edge_nodes_np = (
+            edge_nodes.cpu().numpy() if edge_nodes.is_cuda else edge_nodes.numpy()
+        )
```
Jacob-Chmura: Why is the conversion to numpy necessary? Can we just compute directly on the tensors?

tgm/hooks/node_analytics.py
```diff
@@ -0,0 +1,349 @@
...
+        """Get all neighbors for each node in this batch."""
+        # Convert to numpy for faster iteration
+        nodes_np = nodes.cpu().numpy() if nodes.is_cuda else nodes.numpy()
+        nodes_set = set(int(n) for n in nodes_np)
+        neighbors: Dict[int, Set[int]] = {int(n): set() for n in nodes_set}
+
+        if batch.src is None or batch.dst is None:
+            return neighbors
+
+        # Convert to numpy arrays for faster iteration
```
Jacob-Chmura, line 98: I'm surprised that this would be faster than iterating the tensor? How much faster is it?

tgm/hooks/node_analytics.py
```diff
@@ -0,0 +1,349 @@
...
+
+        if batch.node_ids is None or batch.node_ids.numel() == 0:
+            return node_stats
+
+        new_nodes = 0
+        for node in batch.node_ids:
+            node_id = int(node.item())
+            if node_id not in self._first_seen:
+                new_nodes += 1
+                self._seen_nodes.add(node_id)
```
Jacob-Chmura: Do we actually use `self._seen_nodes` anywhere? It appears that we can remove it altogether

tgm/hooks/node_analytics.py
```diff
@@ -0,0 +1,349 @@
...
+            all_batch_nodes.append(batch.node_ids)
+
+        # No nodes in batch, return empty stats
+        if not all_batch_nodes:
+            node_batch_stats = self._compute_node_statistics(batch)
+            edge_batch_stats = self._compute_edge_statistics(batch)
+            batch.node_stats = {}  # type: ignore[attr-defined]
+            batch.node_macro_stats = node_batch_stats  # type: ignore[attr-defined]
+            batch.edge_stats = {**edge_batch_stats}  # type: ignore[attr-defined]
+            return batch
```
Jacob-Chmura, line 235: This is currently missing coverage, and it seems like a good edge case to consider. Can we add a test on an emty batch and ensure we get reasonable defaults out of it

tgm/hooks/node_analytics.py
xuanvietchu: The reason behind most np cast is that I was profiling the code and realized that it was calling `".items()" a lot, which costs time. I just decided to cast everything to np and use np operations

I tested this by tracking 100 nodes on the example, and it's around 10% faster compared to the following:
`
degrees = {}

for node_val in nodes:
    degree = int((edge_nodes == node_val).sum().item())
    degrees[int(node_val)] = degree
`
 Also, for loops is usually faster on np arrays than tensors

tgm/hooks/node_analytics.py
xuanvietchu: Each of these .numpy() casting sections contributed to around 10% of the speed (100 nodes tracked)

tgm/hooks/node_analytics.py
xuanvietchu: This is actually an oversight on my end. That's part of my outdated code. Thanks!

tgm/hooks/node_analytics.py
Jacob-Chmura: What if we vectorize the degree counting, something like:

```python
unique_nodes, inv = torch.unique(edge_nodes, return_inverse=True)
counts = torch.bincount(inv)
deg_map = dict(zip(unique_nodes.tolist(), counts.tolist()))
return {int(n): deg_map.get(int(n), 0) for n in nodes.tolist()}
```

tgm/hooks/node_analytics.py
```diff
@@ -0,0 +1,341 @@
...
+
+class NodeAnalyticsHook(StatefulHook):
+    """Compute node-centric statistics for a specific set of tracked nodes.
+
+    This hook maintains state across batches to compute temporal statistics
+    for a specified set of nodes.
+
+    Args:
+        tracked_nodes (Tensor): 1D tensor of node IDs to track statistics for.
+        compute_moving_avg (bool): If True, compute exponential moving averages.
```
ntgbaoo: Builds doc test failed due to `compute_moving_avg` documented but not used. Do we need this attribute?

tgm/hooks/node_analytics.py
```diff
@@ -0,0 +1,341 @@
...
+            - lifetime: Time since the node was first seen.
+            - time_since_last_seen: Time since the node was last seen.
+            - appearances: Total number of unique timesteps the node has appeared in.
+        node_macro_stats (Dict): Batch-level node statistics:
+            - node_novelty: Fraction of tracked nodes in the batch that are appearing for the first time.
+            - new_node_count: Number of tracked nodes in the batch that are appearing for the first time.
+        edge_stats (Dict): Batch-level edge statistics:
+            - edge_novelty: Fraction of new edges in the batch, that is not seen in previous batches.
+            - edge_density: Edges this batch / possible edges based on unique nodes
+            - new_edge_count: Number of new edges in the batch, that is not seen in previous batches.
```
ntgbaoo, line 38: nit: The name of the hook, `NodeAnalyticsHook`, is currently a bit misleading to users. Why does node analytics return `edge_stats`. Maybe `NodeLevelAnalyticsHook` or `NodeCentricAnalyticsHook`

tgm/hooks/node_analytics.py
```diff
@@ -0,0 +1,341 @@
...
+    """
+
+    requires: Set[str] = set()
+    produces = {'node_stats', 'node_macro_stats', 'edge_stats'}
+
+    def __init__(self, tracked_nodes: Tensor, num_nodes: int) -> None:
```
ntgbaoo, line 44: we are missing documentation for `num_nodes` in the comment. That is also why the build doc check failed.

tgm/hooks/node_analytics.py
ntgbaoo: In any case, I think it is ideal to keep everything on the cuda and avoid overhead from converting from `cpu` to `cuda`, and vice versa. I think you can try out Jacob's suggestion, another way to do it for your reference:
```python
index =  torch.cat([src,dst])
degree = torch.zeros(number_of_node)
degree = degree.scatter_add_(0, index, degree.new_ones((index.size(0))))
```

I just wrote it up from the terminal. You may need to adapt to make it work in your case.

tgm/hooks/node_analytics.py
benjaminnNgo: If we want to vectorize this, we may need to convert `edge_index` into adj matrix and aggregate to get a list of neighbours for each node at the end?

ntgbaoo: But I think it makes sense why this is faster. By doing it this way: converting from cuda to cpu happens once, while in the case of looping over the tensor, the conversion happens for every iteration (since int(x..item()) will transfer from cuda to cpu by default). 

If we really want to work on tensor only, we may need to convert `edge_index` into adj matrix and aggregate to get a list of neighbours for each node at the end?

benjaminnNgo (commented):
Hey Viet!

Thank you for the PR. In general, it looks good to me. One major concern is just the doc build failed (see my suggestion). I left a minor suggestion on vectorizing the degree operation.

tgm/hooks/node_analytics.py
xuanvietchu: > ```python
> unique_nodes, inv = torch.unique(edge_nodes, return_inverse=True)
> ```

I added this to the code base, turns out ```.unique()``` and ```.bincount()``` is a bit faster (~7%) than the numpy casting and ```(edge_nodes_np == node_val).sum()``` despite of the ```toList()``` conversions

tgm/hooks/node_analytics.py
xuanvietchu: These are is my leftover code I missed during clean up, thanks!

tgm/hooks/node_analytics.py
xuanvietchu: I've updated the doc; you can try it again.

examples/analytics/node_analytics_example.py
```diff
@@ -21,7 +21,7 @@
 parser.add_argument('--dataset', type=str, default='tgbn-trade', help='Dataset name')
 parser.add_argument('--bsize', type=int, default=200, help='batch size')
 parser.add_argument(
-    '--num-tracked', type=int, default=1, help='Number of nodes to track'
+    '--num-tracked', type=int, default=500, help='Number of nodes to track'
```
Jacob-Chmura: I haven't ran with 500 nodes tracked, but for the purposes of an example, I think something smaller (e.g. 10) might be sufficient

tgm/hooks/node_analytics.py
```diff
@@ -11,15 +11,15 @@
 logger = _get_logger(__name__)
 
 
-class NodeAnalyticsHook(StatefulHook):
+class NodeCentricAnalyticsHook(StatefulHook):
```
Jacob-Chmura: I actually prefer `NodeAnalyticsHook` as it's shorter and conveys pretty much the same thing as `NodeCentricAnalyticsHook`

Issue #313, opened 2025-10-06
Add utility to add static node features directly to DGData to unify examples


Jacob-Chmura:
Consider making dgbatch frozen on ground truth (add static node feats), with ability to add new attrs.

PR #368, opened 2025-12-30
Access static_node_feats via loader reference

Jacob-Chmura:
https://github.com/tgm-team/tgm/issues/313#issuecomment-3699761577

Close https://github.com/tgm-team/tgm/issues/313

TLDR: Until the details of the offloading strategy and global/local memory loading is better understood, we should keep `DGBatch` completely local. Issue will be closed for now if we decide not to move forward with this approach ( or something similar).

What I have drafted here is to expose the dgraph handle that the loader stores. This enables us to access the static node features directly on our graph without passing them through and having the user manage their own copy.

#### Pros
- single reference / owner of large static node features memory block
- mmap / caching / offloading implications are slightly easier to reason about since they will be hidden behind the dgraph API instead of passed around globally by the user
- still no per-batch overhead
- avoids potential device mismatch and keeps `static_node_feats` read only

#### Cons
- for the user, it's not entirely obvious that the loader has or _should_ have a dgraph reference. But it seems reasonable (since the loader must iterate the entire dgraph somehow after all
- requires code changes to most examples


tgm/data/dg_data.py
```diff
@@ -475,6 +475,19 @@ def clone(self) -> DGData:
 
         return replace(self, **cloned_fields)  # type: ignore
 
+    @property
```
Jacob-Chmura, line 478: Not `cached_property` since at the `DGData` stage we still treat the underlying data as mutable (and hence the number of nodes could change).

Jacob-Chmura: TODO: test

examples/graphproppred/gcn.py
Jacob-Chmura, line 1: Most example changes are the same:

- read `static_node_feats` from `loader.dgraph.static_node_feats` instead of passing them around
- initialize `static_node_feats` if they don't exist on the underlying `full_data` (so each split get's a reference to the same tensor without cloning)
- use graph API instead of explicitly accessing `static_node_feats` where possible (e.g. `train_dg.static_node_feats_dim` instead of `static_node_feats.shape[1]`

examples/linkproppred/tgat.py
```diff
@@ -203,16 +203,17 @@ def eval(
...
+
+train_data, val_data, test_data = full_data.split()
 train_dg = DGraph(train_data, device=args.device)
 val_dg = DGraph(val_data, device=args.device)
 test_dg = DGraph(test_data, device=args.device)
 
-if train_dg.static_node_feats is not None:
-    static_node_feats = train_dg.static_node_feats
-else:
-    static_node_feats = torch.zeros((test_dg.num_nodes, 1), device=args.device)
```
Jacob-Chmura, line 214: I recognize this is changing init from zeros to randn. Should be better if not the same, but will require a quick check

examples/graphproppred/gcn.py
```diff
@@ -230,6 +229,11 @@ def eval(
     time_delta=args.raw_time_gran,
 ).discretize(args.batch_time_gran)
 
+if full_data.static_node_feats is None:
+    full_data.static_node_feats = torch.randn(
```
Jacob-Chmura, line 233: Consider adding a helper method directly on `DGData` for creating gaussian random static node features... not sure if that's helpful but this seems fine for now I suppose

ntgbaoo: I think it is fine to keep it like this and let users decide what they want.

examples/linkproppred/tgat.py
ntgbaoo: Maybe let's run two versions for 5-10 epochs and see?

examples/nodeproppred/dygformer.py
```diff
@@ -243,36 +246,29 @@ def eval(
 val_dg = DGraph(val_data, device=args.device)
 test_dg = DGraph(test_data, device=args.device)
 
+
 nbr_hook = RecencyNeighborHook(
     num_nbrs=[args.max_sequence_length - 1],  # Keep 1 slot for seed node itself
-    num_nodes=num_nodes,
+    num_nodes=test_dg.num_nodes,
```
ntgbaoo: quick question: will `test_dg.num_nodes` be the same with `full_graph.num_nodes`?

examples/nodeproppred/tgn.py
ntgbaoo, line 1: nice clean up

benjaminnNgo (commented):
I personally think these changes make sense. In this case, `static_node_feats` is global and shared across every split and batch, I think we should raise an error (prefer) or warning if users try to modify it. 
So from what I am understanding, another pros of this approach is: we can keep `static_node_feats` on disk and load parts of it to RAM if need?

examples/nodeproppred/dygformer.py
Jacob-Chmura: Good point. Not necessarily. We should update all examples ot use `full_graph.num_nodes`, since techincally training splitds could have larger node ids than test splits.

examples/linkproppred/tgat.py
```diff
@@ -203,16 +203,17 @@ def eval(
 seed_everything(args.seed)
 evaluator = Evaluator(name=args.dataset)
 
-train_data, val_data, test_data = DGData.from_tgb(args.dataset).split()
+full_data = DGData.from_tgb(args.dataset)
+if full_data.static_node_feats is None:
+    full_data.static_node_feats = torch.randn(
+        (full_data.num_nodes, 1), device=args.device
```
ntgbaoo, line 209: nit: Just realized that the feature of each node has a dimension of 1

PR #369, opened 2025-12-31
Dynamic Node Target API

### LINKED ISSUE #352: Distinguish between Node Labels and Node Features
**Description**
Currently both node labels and node features are dynamic node events, we want to distinguish them

Jacob-Chmura:
Close #352

### DIFF
--- a/tgm/core/_storage/backends/array_backend.py
+++ b/tgm/core/_storage/backends/array_backend.py
@@ -188,19 +188,18 @@ class DGStorageArrayBackend(DGStorageBase):
     def get_dynamic_node_targets(self, slice: DGSliceTracker) -> Optional[Tensor]:
         if self._data.dynamic_node_targets is None:
             return None
-        assert self._data.node_event_idx is not None  # for mypy
-        assert self._data.node_ids is not None  # for mypy
+        assert self._data.node_target_idx is not None  # for mypy
 
         lb_idx, ub_idx = self._binary_search(slice)
-        node_mask = (self._data.node_event_idx >= lb_idx) & (
-            self._data.node_event_idx < ub_idx
+        node_mask = (self._data.node_target_idx >= lb_idx) & (
+            self._data.node_target_idx < ub_idx
         )
         if node_mask.sum() == 0:
             logger.debug(f'No dynamic node targets in slice {slice}')
             return None
 
-        time = self._data.timestamps[self._data.node_event_idx[node_mask]]
-        nodes = self._data.node_ids[node_mask]
+        time = self._data.timestamps[self._data.node_target_idx[node_mask]]
+        nodes = self._data.node_target_ids[node_mask]
         indices = torch.stack([time, nodes], dim=0)
         values = self._data.dynamic_node_targets[node_mask]
 
--- a/tgm/data/split.py
+++ b/tgm/data/split.py
@@ -79,6 +79,7 @@ class SplitStrategy(ABC):
             node_timestamps=node_timestamps,
             node_ids=node_ids,
             dynamic_node_feats=dynamic_node_feats,
+            node_target_idx=node_target_idx,
             dynamic_node_targets=dynamic_node_targets,
             static_node_feats=static_node_feats,
             edge_type=edge_type,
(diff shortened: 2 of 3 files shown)


tgm/core/_storage/backends/array_backend.py
```diff
@@ -185,6 +185,37 @@ def get_dynamic_node_feats(self, slice: DGSliceTracker) -> Optional[Tensor]:
         shape = (max_time + 1, max_node_id + 1, node_feats_dim)
         return torch.sparse_coo_tensor(indices, values, shape)  # type: ignore
 
+    def get_dynamic_node_targets(self, slice: DGSliceTracker) -> Optional[Tensor]:
```
Jacob-Chmura, line 188: TODO: Lots of duplicate code

tgm/core/graph.py
```diff
@@ -89,6 +89,8 @@ def materialize(self, materialize_features: bool = True) -> DGBatch:
             batch.dynamic_node_feats = self.dynamic_node_feats._values()
         if materialize_features and self.edge_feats is not None:
             batch.edge_feats = self.edge_feats
+        if materialize_features and self.dynamic_node_targets is not None:
+            batch.dynamic_node_targets = self.dynamic_node_targets._values()
```
Jacob-Chmura, line 93: TODO: Reconsider whether we need to match sparse tensor for targets since the indices are unused and implicilty assumed to match the `dynamic_node_feats` shape

Jacob-Chmura (commented):
Need to add plenty of tests, this is a draft to get review on high level API before adding tests. Also, check which (if any) examples are actually to benefit from this update.

tgm/core/_storage/backends/array_backend.py
```diff
@@ -221,6 +252,11 @@ def get_dynamic_node_feats_dim(self) -> Optional[int]:
             return None
         return self._data.dynamic_node_feats.shape[1]
 
+    def get_dynamic_node_targets_dim(self) -> Optional[int]:
+        if self._data.dynamic_node_targets is None:
+            return None
+        return self._data.dynamic_node_targets.shape[1]
```
ntgbaoo, line 257: In case the case of dynamic node classification, the dimension of `dynamic_node_targets` would be just 1 for the label of the node, so it will crash if we try to do `.shape[1]`?

tgm/core/graph.py
ntgbaoo: I think we can't assume it matches `dynamic_node_feats`. One example case is dynamic node classification, which we don't have now in `TGB` but users will potentially come up with it.

tgm/data/split.py
ntgbaoo, line 1: Do we need to update `from_csv` and `from_pandas` as well?

benjaminnNgo (commented):
I have a concern:
- Then the dynamic node features will be now the temporal version of `static_node_features`?, then I think we should return a tensor of dynamic node features up to the current time of the slice of all the nodes appearing in that slice?

tgm/core/graph.py
Jacob-Chmura: Hmm okay. Well in that case we would need a separate timestamp tensor for node targets.

tgm/core/_storage/backends/array_backend.py
Jacob-Chmura: On construction I assert that the targets are 2d. We expect first axis to be time, second to be target dimension.

tgm/data/split.py
Jacob-Chmura: Do you mean the class method on dgdata? It was updated.

tgm/core/_storage/backends/array_backend.py
ntgbaoo: Ah I see

tgm/data/split.py
ntgbaoo: Yeah you are right, I missed them.

tgm/core/graph.py
Jacob-Chmura: Before I go ahead and implement, let's discuss (perhaps in our meeting) @shenyangHuang @benjaminnNgo 

If we don't need timestamps associated with the node targets, then this is straightforward: I can simply drop the assertion check that the target and feat shape match. On the other hand, if we need timestamps info then we have to go ahead and add another tensor to our data class.

ntgbaoo: A tensor for a timestamp makes more sense to me, since I think it should be time associated with each label

Issue #342, opened 2025-11-20
TGB graph property prediction example


benjaminnNgo:

benjaminnNgo:
If you think it makes sense, let's re-open this issue and assign it to Beta-2. I can work on it quickly as I prepared the code for the submission.

PR #374, opened 2026-01-06
Added TGB to graphproppred

benjaminnNgo:
### Summary / Description

Made graphproppred examples work on TGB datasets.

**Related Issues:** #342

#### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking Change
- [x] Refactoring
- [ ] Documentation update

#### Test Evidence

- [ ] Unit tests
- [ ] Integration tests
- [ ] Performance tests

- At the time of the submission, @shenyangHuang suggested that we should make the current graphproppred examples to work with TGB dataset. If we want to support both datasets, I think we should modify the existing script (as drafted in this PR) instead of making new example scripts to avoid additional scripts to maintain later.

### DIFF
--- a/test/integration/test_gcn.py
+++ b/test/integration/test_gcn.py
@@ -86,8 +86,6 @@ python "$ROOT_DIR/examples/graphproppred/gcn.py" \
     --dataset {dataset} \
     --device cuda \
     --epochs 5 \
-    --batch-time-gran={batch_time_gran}
-
-    """
+    --batch-time-gran={batch_time_gran}"""
     state = slurm_job_runner(cmd)
     assert state == 'COMPLETED'
--- a/test/integration/test_tgcn.py
+++ b/test/integration/test_tgcn.py
@@ -67,9 +67,7 @@ python "$ROOT_DIR/examples/graphproppred/tgcn.py" \
     --dataset {dataset} \
     --device cuda \
     --epochs 5 \
-    --batch-time-gran={batch_time_gran}
-
-    """
+    --batch-time-gran={batch_time_gran}"""
     state = slurm_job_runner(cmd)
     assert state == 'COMPLETED'
 


examples/graphproppred/persistant_forecast.py
```diff
@@ -49,6 +49,15 @@
     '--log-file-path', type=str, default=None, help='Optional path to write logs'
 )
 
+parser.add_argument(
+    '--use-tgb-dataset',
+    default=False,
+    action=argparse.BooleanOptionalAction,
+    help='if use-tgb-dataset, will use tgbl/tgbn dataset for experiments intsead of token data',
+)
```
Jacob-Chmura: I agree that we should support both here instead of having seperate scripts to maintain.

Though I think it would be fine to just have the single `--dataset` argument, that support _both_ "tgb-xxx", and custom paths. We can just have a helper like:

```python
def load_data(dataset_str: str) -> DGData:
    if pathlib.Path(dataset_str).exists():
        ... # load dataframe
    elif dataset_str.startswith("tgb"):
        ... # native tgb
    else:
        raise ValueError(...)
```

I'm okay with the slightly more complex usage that you implemented for now too.

examples/graphproppred/persistant_forecast.py
```diff
@@ -136,29 +145,35 @@ def eval(
...
+    _, val_data, test_data = full_data.split()
 
-val_dg, test_dg = DGraph(val_data), DGraph(test_data)
-val_loader = DGDataLoader(val_dg, batch_unit=args.batch_time_gran, on_empty='raise')
-test_loader = DGDataLoader(test_dg, batch_unit=args.batch_time_gran, on_empty='raise')
+    val_dg, test_dg = DGraph(val_data), DGraph(test_data)
+    val_loader = DGDataLoader(val_dg, batch_unit=args.batch_time_gran, on_empty='raise')
+    test_loader = DGDataLoader(
+        test_dg, batch_unit=args.batch_time_gran, on_empty='raise'
+    )
```
Jacob-Chmura: This should be moved out of this block so that we have valid loaders in case of custom datasets too?

examples/graphproppred/persistant_forecast.py
shenyangHuang: yeah I agree with this. actually beyond TGB, graph prop pred can run on any arbitrary TG dataset, so techically TGB-seq would work too. It just needs to find the correct IO.

@Jacob-Chmura  Do you think this `load_data` helper function should be a universal helper function for all examples? or at least some version that can load both TGB and TGB-seq for link pred?

examples/graphproppred/persistant_forecast.py
```diff
@@ -136,29 +145,35 @@ def eval(
...
-
-_, val_data, test_data = full_data.split(
-    TemporalRatioSplit(
-        train_ratio=args.train_ratio,
-        val_ratio=args.val_ratio,
-        test_ratio=args.test_ratio,
-    )
-)
+if args.use_tgb_dataset:
+    full_data = DGData.from_tgb(args.dataset).discretize(args.batch_time_gran)
```
shenyangHuang: again the discretize would apply to any custom CTDG the user loads too right? @benjaminnNgo

shenyangHuang: maybe for graph prop tasks in general, we want to specify in a tutorial or top of the file that the tasks is expected to run on snapchats / DTDGs, otherwise it is not well-defined.

shenyangHuang (changes requested):

examples/graphproppred/persistant_forecast.py
Jacob-Chmura: @shenyangHuang Generally I want the examples to be standalone scripts with all the code self contained (except for tgm-lib). This makes it easy for people to copy-paste the entire script without looking into our package. 

This means I rather duplicate code, or keep examples minimal (i.e. do not directly allow users to run across different data types). Of course, there is additional burden on our end to maintain examples, so we should only provide examples that we think are genuinely useful for showcasing our library usage.

examples/graphproppred/persistant_forecast.py
ntgbaoo: Good point. I added a brief note at the top of each example.

Jacob-Chmura (approved):

PR #378, opened 2026-01-09
DGData API Refactor

Jacob-Chmura:
This PR performs the relevant API changes on `DGData` as discussed in #377. **Note: DGraph and DGBatch APIs are left unchanged. Will open a seperate PR and stack diffs on there.**

Out of scope: Node / edge target integration

PR #379, opened 2026-01-09
DGraph API Refactor

Jacob-Chmura:
This PR stacks #378 and updates our `DGraph` (and by extension storage backend) APIs. 

### Main changes

**DGraph / DGStorage:**
- rename `static_node_feat` -> `static_node_x`, `edge_feat` -> `edge_x`, `dynamic_node_feat` -> `node_x`, etc in line with previous updates to `DGData`

**DGraph:**
- remove `nodes` property
- rename `num_edges` -> `num_edge_events` 
- add `num_node_events`
- deprecate `edges`? in favour or in addition to `edge_index` + `edge_time`
- add method to get `node_event_node_ids` and `node_event_time`

**DGStorage**
- add `get_node_events` API to storage backend (analogous to `get_edges`)

### Out of scope:
- `DGBatch` api changes
- Distinguishing features / targets in API


tgm/core/graph.py
```diff
@@ -209,33 +213,64 @@ def num_events(self) -> int:
...
-        return self._storage.get_nodes(self._slice)
-
     @_logged_cached_property
     def _edges_cpu(self) -> Tuple[Tensor, Tensor, Tensor]:
         return self._storage.get_edges(self._slice)
 
     @property
     def edges(self) -> Tuple[Tensor, Tensor, Tensor]:
         """The src, dst, time tensors over the dynamic graph."""
+        # TODO: Deprecate?
```
Jacob-Chmura: Deprecating this means updating a bunch of tests. Need to decide if it's better to keep or not.

Jacob-Chmura: I think it's okay to keep for now.

tgm/core/graph.py
```diff
@@ -209,33 +213,64 @@ def num_events(self) -> int:
...
+        return self._storage.get_node_events(self._slice)
+
+    @property
+    def node_events(self) -> Tuple[Tensor, Tensor]:
+        """The node ids and time tensors for node events over the dynamic graph."""
+        node_ids, node_time = self._node_events_cpu
+        return node_ids.to(self.device), node_time.to(self.device)
+
+    @property
+    def node_event_node_ids(self) -> Tensor:
```
Jacob-Chmura, line 249: I prefer this over `node_event_nids` even though it's more verbose.

tgm/core/graph.py
```diff
@@ -209,33 +213,64 @@ def num_events(self) -> int:
...
+        """The timestamps associated with the edge events over the dynamic graph."""
+        _, _, time = self._edges_cpu
+        return time.to(self.device)
+
+    @_logged_cached_property
+    def _node_events_cpu(self) -> Tuple[Tensor, Tensor]:
+        return self._storage.get_node_events(self._slice)
+
+    @property
+    def node_events(self) -> Tuple[Tensor, Tensor]:
```
Jacob-Chmura, line 243: I would say either we keep this (and `edges`) or remove both.

ntgbaoo: I may have missed something, but it's still not clear to me why returning `edge_idx` and the corresponding time independently instead of returning the tuple as `edges` does. 

But if we keep them both. We need to come up with very clear names. See my comments below.

tgm/core/graph.py
ntgbaoo: Yeah I like this as well. I like how your naming convention for nodes:
- `node_events` : return both node events and times
- `node_event_node_ids`: return node events only
- `node_event_time`: return time for node events only

I would prefer to keep the same convention for edge:
- `edge_events` : return both edge events and times
- `edge_event_edge_idx`: return edge events only
- `edge_event_time`: return time for edge events only

It is kinda verbose, but it is consistent with node and clearer about what to be returned

benjaminnNgo (commented):
I prefer keeping both ways of returning edge/node events for now.

test/unit/test_core/test_dgraph.py
```diff
@@ -6,26 +6,31 @@
 from tgm import DGBatch, DGraph
 from tgm.data import DGData
 
+# num_node_events
+# edge_index, edge_time
+# node_events
+# node_event_node_ids, node_time
```
ntgbaoo, line 12:
```suggestion
```

PR #391, opened 2026-01-21
Version bump

Jacob-Chmura:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "tgm-lib"
-version = "0.1.0b0"
+version = "0.1.0b1"
```
shenyangHuang, line 3: I am not sure about naming conventions but shouldn't we just do `0.2.0` ?

PR #380, opened 2026-01-09
Forced Dedup hook runs at the end

benjaminnNgo:
### Summary / Description

Temporary force dedup hook runs at the end until "seedable" and instance-level requirement is introduced to `DGHook`.

**Related Issues:** #285 
**Future Issue:** #381 (Reference for visibility)

#### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking Change
- [x] Refactoring
- [ ] Documentation update

#### Test Evidence

Describe how this PR has been tested.

- [x] Unit tests
- [ ] Integration tests
- [ ] Performance tests

#### Questions / Discussion Points

List any areas where you’d like reviewer input or have open questions.

PR #396, opened 2026-02-16
Make hooks seedable and added instance-level properties

### LINKED ISSUE #381: Instance-Level Hook Resolution, with seedable requirements

ntgbaoo:
### Summary / Description

This PR covered:
- Made `requires` and `produces` as instance-level variables for each hook
- Added `id` to each hook constructor
- Removed [hack that forces the dedup hook always run at the end ](https://github.com/tgm-team/tgm/pull/380)

**Related Issues:** #381

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [X] Refactoring
- [ ] Documentation update

#### Test Evidence

- [X] Unit tests
- [ ] Integration tests
- [ ] Performance tests

I will add this function to `BaseDGHook`:
```python
    def add_attribute_to_batch(self, batch: DGBatch ,name: str, value: Any) -> None:
        """
        Add a new attribute to providede batch.

        If `id` is specified, the new attribute name will be appended with the given `id` as a suffix.
        """
        if self.id:
            name = f'{name}_{self.id}'
        setattr(batch, name, value)
```

I will refactor our code so we use this method every time we add a new attribute to the batch

This is a draft, and unit tests will be added to improve codecov.

### DIFF
--- a/tgm/hooks/base.py
+++ b/tgm/hooks/base.py
@@ -66,8 +66,8 @@ class BaseDGHook(ABC):
     def reset_state(self) -> None:
         pass
 
-    def add_attribute_to_batch(self, batch: DGBatch, name: str, value: Any) -> None:
-        """Add a new attribute to providede batch.
+    def add_batch_attribute(self, batch: DGBatch, name: str, value: Any) -> None:
+        """Add a new attribute to provided batch.
 
         If `_id` is specified, the new attribute name will be appended with the given `id` as a suffix.
         """
@@ -90,6 +90,12 @@ class StatefulHook(BaseDGHook):
 
 @dataclass(eq=False)
 class SeedableHook(BaseDGHook):
+    """Base class for hooks that can take additional attributes.
+
+    Example hooks that are `seedable`: `DeduplicationHook` which finds set of unique nodes ID on `seed_keys` beyond `src` and `dst`.
+    `SeedableHook` are compatible with both `StatelessHook` and `StatefulHook`.
+    """
+
     seed_keys: List[str] | None = field(default_factory=list)
 
     def __post_init__(self) -> None:
--- a/tgm/hooks/base.py
+++ b/tgm/hooks/base.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
-from typing import Any, Protocol, Set, runtime_checkable
+from typing import Any, List, Protocol, Set, runtime_checkable
 
 from tgm import DGBatch, DGraph
 
@@ -11,9 +11,13 @@ from tgm import DGBatch, DGraph
 class DGHook(Protocol):
     r"""The behaviours to be executed on a DGraph before materializing."""
 
-    requires: Set[str]
-    produces: Set[str]
     has_state: bool
+
+    @property
+    def requires(self) -> Set[str]: ...
+
+    @property
+    def produces(self) -> Set[str]: ...
 
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch: ...
 
@@ -24,19 +28,35 @@ class DGHook(Protocol):
 class BaseDGHook(ABC):
     """Base class for hooks."""
 
-    requires: Set[str] = field(default_factory=set)
-    produces: Set[str] = field(default_factory=set)
-    id: str | None = None
+    _cls_requires: Set[str] = field(default_factory=set, init=False, repr=False)
+    _cls_produces: Set[str] = field(default_factory=set, init=False, repr=False)
+
+    _requires: Set[str] = field(default_factory=set)
+    _produces: Set[str] = field(default_factory=set)
+
+    _id: str | None = None
+
     has_state: bool = False
 
     def __post_init__(self) -> None:
-        if self.id:
-            self.produces = {f'{p}_{self.id}' for p in self.produces}
+        self._requires.update(type(self).__dict__.get('_cls_requires', set()))
+        self._produces.update(type(self).__dict__.get('_cls_produces', set()))
+
+    @property
+    def produces(self) -> Set[str]:
+        if self._id is None:
+            return self._produces
+        else:
+            return {f'{p}_{self._id}' for p in self._produces}
+
+    @property
+    def requires(self) -> Set[str]:
+        return self._requires
 
     def __repr__(self) -> str:
         cls_name = self.__class__.__name__
-        if self.id:
-            cls_name = f'{cls_name}_{self.id}'
+        if self._id:
+            cls_name = f'{cls_name}_{self._id}'
         return cls_name
 
     @abstractmethod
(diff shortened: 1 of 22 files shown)


tgm/hooks/dedup.py
```diff
@@ -16,27 +18,31 @@ class DeduplicationHook(StatelessHook):
     Note: Supports batches with or without negative samples and multi-hop neighbors.
     """
 
-    requires = {'edge_src', 'edge_dst'}
-    produces = {'unique_nids', 'global_to_local'}
+    def __init__(self, seed_nodes_keys: List[str] = [], id: str | None = None) -> None:
```
Jacob-Chmura: Default `seed_nodes_keys` should be `None` to prevent mutabel default argument issues.

tgm/hooks/dedup.py
```diff
@@ -16,27 +18,31 @@ class DeduplicationHook(StatelessHook):
...
+        super().__init__()
+        self.id = id
+        self.requires = {'edge_src', 'edge_dst'} | set(seed_nodes_keys)
+        self.produces = {'unique_nids', 'global_to_local'}
+        self.__post_init__()
```
Jacob-Chmura: This setup makes sense, but there are two points I think that are worth raising

- If somehow `self.__post_init__()` ends up being called twice (or more) the naming will be messed up
- Each new hooks must carefully follow the order of events we have here (super init, define custom attrs, call post init). 

The first may be an issue in case of nested object hierarchy I believe.

tgm/hooks/base.py
```diff
@@ -18,23 +20,40 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch: ...
...
-    requires: Set[str] = set()
-    produces: Set[str] = set()
+    requires: Set[str] = field(default_factory=set)
+    produces: Set[str] = field(default_factory=set)
+    id: str | None = None
     has_state: bool = False
 
+    def __post_init__(self) -> None:
+        if self.id:
+            self.produces = {f'{p}_{self.id}' for p in self.produces}
```
Jacob-Chmura: I'm not sure I have a better alternative off the top of my head, but what do you think about something like

```python
class BassDGHook(ABC):
    _cls_requires: Set[str] = field(default_factory=set, init=False, repr=False)
    _cls_produces: Set[str] = field(default_factory=set, init=False, repr=False)

    _requires: Set[str] = field(default_factory=set)
    _produces: Set[str] = field(default_factory=set)
    _id: str | None = None

def __post_init__(self):
    self._requires.update(self._cls_requires)
    self._produces.update(self._cls_produces)

@property
def produces(self) -> Set[str]:
    if self.id is None:
        return frozenset(self._produces)
    else:
        return frozenset({f'{p}_{self.id}' for p in self.produces})

@property
def requires(self) -> Set[str]:
    return frozenset(self._requires)
...

class SimpleHook(StatefulHook):
    _cls_requires = {'foo', 'bar'}
    _cls_produces = {'baz'}
    ...

class SeedableHook(StatefulHook):
    _cls_requires = {'foo', 'bar'}
    _cls_produces = {'baz'}
    
    seed_keys: List[str] = field(default_factory=list)

    def __post_init__(sefl):
        super().__post_init__() 
        self._requires.update(self.seed_keys)
```

Jacob-Chmura: I'm not 100% convinced of what I wrote above but the idea is

(a) prevent weird things happening with nested sub classes
(b) prevent user from directly modifying `requires/produces`
(c) non-seedable hooks (`SimpleHook`) should have no extra work required for hook setup

Patching `id` in the property getter has the downside that we must allocate this new set everytime `requires/produces` is called. But these are only really called outside of the hot path in the hook manager setup.

tgm/hooks/dedup.py
```diff
@@ -16,27 +18,31 @@ class DeduplicationHook(StatelessHook):
...
+        ]  # TODO: Need to retrieve seed nodes. Have NodeSeedable?
+        for node_attr in self.requires:
+            if not hasattr(batch, node_attr):
+                raise ValueError(f'Missing seed node attribut {node_attr}')
+
+            if 'nbr_nids' in node_attr:
+                for hop in range(len(batch.nbr_nids)):  # type: ignore[attr-defined]
+                    nbr_nodes = batch.nbr_nids[hop].flatten()  # type: ignore[attr-defined]
+                    nbr_mask = nbr_nodes != PADDED_NODE_ID
+                    nids.append(nbr_nodes[nbr_mask].flatten().to(batch.edge_src.device))
```
Jacob-Chmura, line 45: Hmm, this makes sense, but makes me think that our nbr sampler should output seperate produces for each hop maybe? Like `nbr_nids_hop_0`, etc? I'm not sure about that, but I don't see a workaround othe than what you implemented here.

Do we really need the device transfer here? Or do we get away without it?

tgm/hooks/dedup.py
```diff
@@ -16,27 +18,31 @@ class DeduplicationHook(StatelessHook):
...
-            batch.node_x_nids.to(batch.edge_src.device)
-        ) if batch.node_x_nids is not None else None
-
-        nids.append(
-            batch.node_y_nids.to(batch.edge_src.device)
-        ) if batch.node_y_nids is not None else None
+        nids = [
+            batch.edge_src,
+            batch.edge_dst,
+        ]  # TODO: Need to retrieve seed nodes. Have NodeSeedable?
```
Jacob-Chmura: I'm not sure what this means, can you clarify?

Jacob-Chmura (commented):
I've done a pass and left some thoughts on the design. The directions looks good but there are some details to iron out.

tgm/hooks/base.py
ntgbaoo: Yeah I think your design makes a lot more sense

tgm/hooks/dedup.py
ntgbaoo: As discussed, we will handle this in another PR

tgm/hooks/base.py
```diff
@@ -9,32 +11,87 @@
...
+
+    @abstractmethod
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
         raise NotImplementedError
 
     def reset_state(self) -> None:
         pass
 
+    def add_attribute_to_batch(self, batch: DGBatch, name: str, value: Any) -> None:
+        """Add a new attribute to providede batch.
```
shenyangHuang: typo: "providede batch" provided batch

tgm/hooks/base.py
```diff
@@ -9,32 +11,87 @@
...
+        return cls_name
+
+    @abstractmethod
     def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch:
         raise NotImplementedError
 
     def reset_state(self) -> None:
         pass
 
+    def add_attribute_to_batch(self, batch: DGBatch, name: str, value: Any) -> None:
```
shenyangHuang: minor: do we need this long of a function name? I am okay with it, but can we also call it `add_batch_attribute` or sth

shenyangHuang: we might also mention that this is the abstract method of which we can add attribute for new hook classes

tgm/hooks/base.py
```diff
@@ -9,32 +11,87 @@
...
+class StatefulHook(BaseDGHook):
     """Base class for hooks that maintain internal state."""
 
-    requires: Set[str] = set()
-    produces: Set[str] = set()
     has_state: bool = True
+
+
+@dataclass(eq=False)
+class SeedableHook(BaseDGHook):
```
shenyangHuang, line 92: no description for seedableHook, when should you use it?

shenyangHuang: also explain how does it fit with the stateful and stateless hook hierarchy we have

shenyangHuang (changes requested):
Overall looks good, some comments to address

shenyangHuang (approved):
looks good, thanks Bao

PR #400, opened 2026-04-11
Saving loading hook state

NaziaHossain066:
### Summary / Description

Adds checkpointing support for stateful hooks in TGM, enabling training to be resumed from mid-epoch interruptions without losing hook state or RNG reproducibility. Demonstrates end-to-end checkpoint save/resume in `examples/linkproppred/tgat.py`.

**Changes:**
- Added `state_dict()` / `load_state_dict()` to `DGHook`, `StatelessHook`, `StatefulHook`, `RecencyNeighborHook`, `EdgeEventsSeenNodesTrackHook`, `NodeAnalyticsHook`, and `HookManager` (with `id()`-based deduplication for shared hooks)
- Added `skip_batches` to `DGDataLoader` to skip already-processed batches on resume without running hooks on them
- Overrode `_get_iterator` in `DGDataLoader` to neutralize PyTorch's internal RNG consumption, ensuring resumed runs are bit-exact
- Updated `examples/linkproppred/tgat.py` with `save_checkpoint`, `load_checkpoint`, and `--resume` / `--checkpoint-dir` args
- Added 9 unit tests in `test/unit/test_hooks/test_state_hook.py`
**Related Issues:** # (395)

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [ ] Integration tests
- [ ] Performance tests

Should checkpointing be extended to other example files beyond `tgat.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,7 +17,7 @@ ipython_config.py
 
 # Remove previous ipynb_checkpoints
 #   git rm -r .ipynb_checkpoints/
-
+outputs/
 ### Python ###
 # Byte-compiled / optimized / DLL files
 __pycache__/
--- a/examples/linkproppred/tgat.py
+++ b/examples/linkproppred/tgat.py
@@ -1,4 +1,6 @@
 import argparse
+import os
+import random
 
 import numpy as np
 import torch
@@ -140,15 +142,19 @@ def train(
     encoder: nn.Module,
     decoder: nn.Module,
     opt: torch.optim.Optimizer,
+    epoch: int,
+    hm: object,
+    nbr_hook: object,
+    start_batch: int = -1,
 ) -> float:
     encoder.train()
     decoder.train()
     total_loss = 0
     static_node_x = loader.dgraph.static_node_x
 
-    for batch in tqdm(loader):
+    for local_idx, batch in enumerate(tqdm(loader)):
+        batch_idx = local_idx + start_batch + 1
         opt.zero_grad()
-
         z = encoder(batch, static_node_x)
         z_src, z_dst, z_neg = torch.chunk(z, 3)
 
@@ -160,6 +166,14 @@ def train(
         loss.backward()
         opt.step()
         total_loss += float(loss)
+
+        print(f'epoch {epoch} batch {batch_idx} loss {float(loss):.6f}')
+        print(
+            f'epoch {epoch} batch {batch_idx} write_pos_sum {nbr_hook._write_pos.sum().item():.0f}'
+        )
+
+        if batch_idx % 100 == 0:
+            save_checkpoint(epoch, batch_idx, encoder, decoder, opt, hm)
     return total_loss
 
 
@@ -235,7 +249,6 @@ hm = RecipeRegistry.build(
 train_key, val_key, test_key = hm.keys
 hm.register_shared(nbr_hook)
 
-train_loader = DGDataLoader(train_dg, args.bsize, hook_manager=hm)
 val_loader = DGDataLoader(val_dg, args.bsize, hook_manager=hm)
 test_loader = DGDataLoader(test_dg, args.bsize, hook_manager=hm)
 
@@ -252,14 +265,80 @@ decoder = LinkPredictor(node_dim=args.embed_dim, hidden_dim=args.embed_dim).to(
     args.device
 )
 opt = torch.optim.Adam(
-    set(encoder.parameters()) | set(decoder.parameters()), lr=float(args.lr)
+    list(encoder.parameters()) + list(decoder.parameters()), lr=float(args.lr)
 )
 
 best_val = 0.0
 
-for epoch in range(1, args.epochs + 1):
+
+def save_checkpoint(epoch, batch_idx, encoder, decoder, opt, hm):
+    os.makedirs('outputs/checkpoints', exist_ok=True)
+    path = f'outputs/checkpoints/ckpt_e{epoch}_b{batch_idx}.pt'
+    torch.save(
+        {
+            'epoch': epoch,
+            'batch_idx': batch_idx,
+            'encoder': encoder.state_dict(),
+            'decoder': decoder.state_dict(),
+            'opt': opt.state_dict(),
+            'hm': hm.state_dict(),
+            'rng_torch': torch.get_rng_state(),
+            'rng_numpy': np.random.get_state(),
+            'rng_python': random.getstate(),
+        },
+        path,
+    )
+
+
+def load_checkpoint(path, encoder, decoder, opt, hm):
+    ckpt = torch.load(path, weights_only=False)
+    encoder.load_state_dict(ckpt['encoder'])
+    decoder.load_state_dict(ckpt['decoder'])
+    try:
+        opt.load_state_dict(ckpt['opt'])
+    except (ValueError, RuntimeError) as e:
+        print(
+            f'Warning: skipping optimizer state (shape mismatch — delete outputs/checkpoints/ and rerun): {e}'
+        )
+    hm.load_state_dict(ckpt['hm'])
+    torch.set_rng_state(ckpt['rng_torch'])
+    np.random.set_state(ckpt['rng_numpy'])
+    return ckpt['epoch'], ckpt['batch_idx']
+
+
+def find_latest_checkpoint(directory):
+    if not os.path.isdir(directory):
+        return None
+    pts = [
+        os.path.join(directory, f) for f in os.listdir(directory) if f.endswith('.pt')
+    ]
+    if not pts:
+        return None
+    return max(pts, key=os.path.getmtime)
+
+
+start_epoch = 1
+start_batch = -1
+ckpt_path = find_latest_checkpoint('outputs/checkpoints')
+if ckpt_path is not None:
+    start_epoch, start_batch = load_checkpoint(ckpt_path, encoder, decoder, opt, hm)
+
+for epoch in range(start_epoch, args.epochs + 1):
+    skip = (start_batch + 1) if epoch == start_epoch else 0
+    train_loader = DGDataLoader(
+        train_dg, args.bsize, hook_manager=hm, skip_batches=skip
+    )
     with hm.activate(train_key):
-        loss = train(train_loader, encoder, decoder, opt)
+        loss = train(
+            train_loader,
+            encoder,
+            decoder,
+            opt,
+            epoch,
+            hm,
+            nbr_hook,
+            start_batch if epoch == start_epoch else -1,
+        )
 
     with hm.activate(val_key):
         val_mrr = eval(val_loader, encoder, decoder, evaluator)
--- a/examples/linkproppred/tgat.py
+++ b/examples/linkproppred/tgat.py
@@ -52,6 +52,18 @@ parser.add_argument(
 parser.add_argument(
     '--log-file-path', type=str, default=None, help='Optional path to write logs'
 )
+parser.add_argument(
+    '--checkpoint-dir',
+    type=str,
+    default='outputs/checkpoints',
+    help='Directory to save checkpoints',
+)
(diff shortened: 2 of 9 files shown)


examples/linkproppred/tgat.py
```diff
@@ -140,15 +154,19 @@ def train(
     encoder: nn.Module,
     decoder: nn.Module,
     opt: torch.optim.Optimizer,
+    epoch: int,
+    hm: object,
```
ntgbaoo, line 158:
```suggestion
    hm: HookManager,
```

examples/linkproppred/tgat.py
```diff
@@ -140,15 +154,19 @@ def train(
...
+    nbr_hook: object,
```
ntgbaoo, line 159:
```suggestion
    nbr_hook: StatefulHook,
```

examples/linkproppred/tgat.py
```diff
@@ -160,6 +178,13 @@ def train(
         loss.backward()
         opt.step()
         total_loss += float(loss)
+
+        print(
+            f'epoch {epoch} batch {batch_idx} loss {float(loss):.6f} write_pos_sum {nbr_hook._write_pos.sum().item():.0f}'
+        )
```
ntgbaoo, line 184:
```suggestion
```

examples/linkproppred/tgat.py
```diff
@@ -160,6 +178,13 @@ def train(
...
+
+        if batch_idx % 100 == 0:
```
ntgbaoo, line 186: Would be nice to have a variable called `checkpoint_interval` from the parser instead of hardcoding 100.

I think another flag is needed to indicate whether we want to checkpoint the model as well. Such as `train_checkpoint`, then the condition will have

```python
if batch_idx %  checkpoint_interval == 0 && args.train_checkpoint:
```

examples/linkproppred/tgat.py
```diff
@@ -252,14 +276,87 @@ def eval(
...
 opt = torch.optim.Adam(
-    set(encoder.parameters()) | set(decoder.parameters()), lr=float(args.lr)
+    list(encoder.parameters()) + list(decoder.parameters()), lr=float(args.lr)
 )
 
 best_val = 0.0
 
-for epoch in range(1, args.epochs + 1):
+
+def save_checkpoint(epoch, batch_idx, encoder, decoder, opt, hm):
```
ntgbaoo, line 285: Can we add a type for each input variable for this function? And would be nice to have a comment to describe the purpose of this method as well.

examples/linkproppred/tgat.py
```diff
@@ -252,14 +276,87 @@ def eval(
...
+            'rng_torch': torch.get_rng_state(),
+            'rng_numpy': np.random.get_state(),
+            'rng_python': random.getstate(),
+            'best_val': best_val,
+        },
+        path,
+    )
+
+
+def load_checkpoint(path, encoder, decoder, opt, hm):
```
ntgbaoo, line 305: same for this function. A type for each variable, a function comment is also necessary here

examples/linkproppred/tgat.py
```diff
@@ -252,14 +276,87 @@ def eval(
...
+    list(encoder.parameters()) + list(decoder.parameters()), lr=float(args.lr)
 )
 
 best_val = 0.0
 
-for epoch in range(1, args.epochs + 1):
+
+def save_checkpoint(epoch, batch_idx, encoder, decoder, opt, hm):
+    os.makedirs(args.checkpoint_dir, exist_ok=True)
+    path = os.path.join(args.checkpoint_dir, f'ckpt_e{epoch}_b{batch_idx}.pt')
```
ntgbaoo, line 287:
```suggestion
    path = os.path.join(args.checkpoint_dir, f'ckpt_epoch_{epoch}_batch_{batch_idx}.pt')
```

examples/linkproppred/tgat.py
```diff
@@ -160,6 +178,13 @@ def train(
         loss.backward()
         opt.step()
         total_loss += float(loss)
+
+        print(
+            f'epoch {epoch} batch {batch_idx} loss {float(loss):.6f} write_pos_sum {nbr_hook._write_pos.sum().item():.0f}'
+        )
+
+        if batch_idx % 100 == 0:
+            save_checkpoint(epoch, batch_idx, encoder, decoder, opt, hm)
```
ntgbaoo, line 187: I would prefer a checkpoint after each epoch rather than each batch

What do you think @shenyangHuang @Jacob-Chmura

examples/linkproppred/tgat.py
```diff
@@ -252,14 +276,87 @@ def eval(
...
+
+
+def load_checkpoint(path, encoder, decoder, opt, hm):
+    ckpt = torch.load(path, weights_only=False)
+    encoder.load_state_dict(ckpt['encoder'])
+    decoder.load_state_dict(ckpt['decoder'])
+    try:
+        opt.load_state_dict(ckpt['opt'])
+    except (ValueError, RuntimeError) as e:
+        print(
```
ntgbaoo, line 312:
```suggestion
        logger.warning(
```

examples/linkproppred/tgat.py
ntgbaoo, line 1: I think we should add the logic from `load_checkpoint`, `save_checkpoint` and `find_latest_checkpoint` to `CheckPointHandler` in `tgm/util/` if we want to have checkpointing in other examples as well, to avoid duplicated code

examples/linkproppred/tgat.py
ntgbaoo, line 159:
```suggestion

```

examples/linkproppred/tgat.py
```diff
@@ -252,14 +276,87 @@ def eval(
...
     with hm.activate(train_key):
-        loss = train(train_loader, encoder, decoder, opt)
+        loss = train(
+            train_loader,
+            encoder,
+            decoder,
+            opt,
+            epoch,
+            hm,
+            nbr_hook,
```
ntgbaoo, line 357:
```suggestion
```

tgm/data/loader.py
```diff
@@ -105,6 +105,7 @@ def __init__(
         batch_unit: str = 'r',
         on_empty: Literal['skip', 'raise', None] = 'skip',
         hook_manager: HookManager | None = None,
+        skip_batches: int = 0,
```
ntgbaoo, line 108: Under `class DGDataLoader(_SkippableDataLoaderMixin, torch.utils.data.DataLoader):`
There is a comment section for documentation. Would you mind adding a new variable `skip_batches` to `Args`?

tgm/data/loader.py
```diff
@@ -169,6 +172,25 @@ def __call__(self, slice_start: List[int]) -> DGBatch:
             batch = self._hook_manager.execute_active_hooks(dg, batch)
         return batch
 
+    def _get_iterator(self) -> Any:
+        # PyTorch's _BaseDataLoaderIter.__init__ consumes 2 global RNG samples for
+        # an internal base_seed (used only for worker processes). Save and restore
+        # around iterator creation so the global RNG state is unaffected, making
+        # checkpoint resume produce identical results to an uninterrupted run.
```
ntgbaoo, line 179:
```suggestion
        """PyTorch's _BaseDataLoaderIter.__init__ consumes 2 global RNG samples for
        an internal base_seed (used only for worker processes). Save and restore
        around iterator creation so the global RNG state is unaffected, making
        checkpoint resume produce identical results to an uninterrupted run.
        """
```

examples/linkproppred/tgat.py
```diff
@@ -50,6 +52,18 @@
 parser.add_argument(
     '--log-file-path', type=str, default=None, help='Optional path to write logs'
 )
+parser.add_argument(
+    '--checkpoint-dir',
+    type=str,
+    default='outputs/checkpoints',
```
ntgbaoo, line 58: nit:
```suggestion
    default='artifact/checkpoints',
```

.gitignore
```diff
@@ -17,7 +17,7 @@ ipython_config.py
 
 # Remove previous ipynb_checkpoints
 #   git rm -r .ipynb_checkpoints/
-
+outputs/
```
ntgbaoo, line 20: nit
```suggestion
artifact/
```

tgm/hooks/base.py
```diff
@@ -17,6 +17,10 @@ def __call__(self, dg: DGraph, batch: DGBatch) -> DGBatch: ...
 
     def reset_state(self) -> None: ...
 
+    def state_dict(self) -> dict: ...
```
ntgbaoo, line 20:
```suggestion
    @property
    def state_dict(self) -> dict: ...
```

tgm/hooks/node_analytics.py
```diff
@@ -211,6 +212,30 @@ def _compute_edge_statistics(self, batch: DGBatch) -> Dict[str, float]:
...
+        return {
+            '_first_seen': self._first_seen,
+            '_last_seen': self._last_seen,
+            '_appearances': self._appearances,
+            '_total_timesteps': self._total_timesteps,
+            '_node_timesteps': self._node_timesteps,
+            '_all_neighbors': self._all_neighbors,
+            '_engagement_sum': self._engagement_sum,
+            '_seen_edges': self._seen_edges,
+            '_tracked_mask': self._tracked_mask.cpu().clone(),
```
ntgbaoo, line 225: Do we actually need this?

benjaminnNgo (commented):
Hey @NaziaHossain066 

Thanks for the PR. Great contribution!

Overall, the PR looks pretty good. I left a couple of comments and raised some concerns for us to discuss.

I just approved CI workflows to run for this PR. It is worth checking [codecov](https://github.com/tgm-team/tgm/pull/400#issuecomment-4230281230)and adding some unit tests to cover some lines that are missing tests rn, such as:
- `state_dict` and `load_state_dict` for node analytics and node tracks

examples/linkproppred/tgat.py
NaziaHossain066: From my understanding, stateful hooks are reset at the end of each epoch via `hm.reset_state()`. Therefore, a checkpoint saved at that point would capture an empty state, effectively no different from starting the next epoch fresh.

ntgbaoo: Don't worry about this comment. I thought that I had deleted this comment (but I actually forgot lol). But yeah, you are right

examples/linkproppred/tgat.py
shenyangHuang: yes I agree

examples/linkproppred/tgat.py
shenyangHuang: should we just checkpoint only after each epoch though this also makes sense for larger datasets

examples/linkproppred/tgat.py
```diff
@@ -252,14 +276,87 @@ def eval(
...
+        print(
+            f'Warning: skipping optimizer state (shape mismatch — delete outputs/checkpoints/ and rerun): {e}'
+        )
+    hm.load_state_dict(ckpt['hm'])
+    torch.set_rng_state(ckpt['rng_torch'])
+    np.random.set_state(ckpt['rng_numpy'])
+    return ckpt['epoch'], ckpt['batch_idx'], ckpt.get('best_val', 0.0)
+
+
+def find_latest_checkpoint(directory):
```
shenyangHuang, line 321: this should be a helper function in util as Bao suggested

examples/linkproppred/tgat.py
shenyangHuang: potentially need to make this verbose by default, user should want to know when and where the checkpoint was resumed from

shenyangHuang (changes requested):

PR #402, opened 2026-04-20
BaseNN for encoder and decoder

ntgbaoo:
### Summary / Description

This PR handles:
- Fixed existing minor bugs in examples.
- Added base class for merge operations (used in `LinkPredictor`). Enabling dependency injection for `LinkPredictor`: the user can implement a custom merge operation and pass it to `LinkPredictor` instead of modifying the core implementation of `LinkPredictor`
- Added base class for graph pooling operations (used in `GraphPredictor`).
- Moved `TGAT` to `tgm.nn`
- Added base class for NN module
- Added the ability to check whether `requires` from the NN module can be satisfied by registered hooks for `HookManager.`

**Related Issues:** #302 

#### Type of Change

- [X] Bug fix
- [X] New feature
- [ ] Breaking Change
- [X] Refactoring
- [ ] Documentation update

#### Test Evidence

- [ ] Unit tests
- [ ] Integration tests
- [ ] Performance tests

Regarding the base class for the NN module, as Jacob suggested, we can make encoders and decoders as hooks. However, I believe that this will introduce some unnecessary complexity for the user:
-  `TGCN` requires additional arguments: `H`. Users need to implement an additional hook to carry over states from the previous iteration.
- Since now the hook has `ID`, thus the attributes may be produced with `_id` suffix. Users need to handle this when defining their `require`

I am more in favour of this suggestion from Jacob:
> For people bringing their own models from outside TGM, they should just implement a proxy object (which has the appropriate hook base class). Declare batch dependencies and forward the forward(batch) call to a forward() call on your own model with the appropriate arguments pulled out

Thus, in this PR, I propose a base class for "a proxy object" instead of models inside `tgm.nn`:

```python

class BaseNN(ABC, nn.Module):
    r"""Base NN for all neural network models."""

    def __init__(self) -> None:
        super().__init__()
        self._requires: Set[str] = set()

    @property
    def requires(self) -> Set[str]:
        return self._requires

    @abstractmethod
    def forward(self, batch: DGBatch, *args: Any, **kwargs: Any) -> Any: ...

```

- `_requires` and `requires` will be used by Hook Manager to check whether the produced artifacts of all registered hooks can satisfy NN module's requirement and raise detailed errors and suggestions to resolve the issue. Here is an example of error message:

```shell
tgm.exceptions.UnresolvableHookDependenciesError: Cannot resolve the following requirements {'foo', 'neighbour', 'num_edge_eventss'} from any hook registered under key 'train'.
Suggestions:
        - 'foo': Can not find any existing hooks that satisfy this requirement.
        - 'neighbour': Found keyword 'neighbour' in 'NeighborSamplerHook' documentation. If this hook produces what you are looking for, update the module requirement with the correct name and register 'NeighborSamplerHook' with key 'train'.
        - 'neighbour': Found keyword 'neighbour' in 'RecencyNeighborHook' documentation. If this hook produces what you are looking for, update the module requirement with the correct name and register 'RecencyNeighborHook' with key 'train'.
        - 'num_edge_eventss': Do you mean 'num_edge_events' or 'num_node_events'?. If so, please update the module requirement with the correct name and register 'BatchAnalyticsHook' with key 'train' to resolve this.

```

For future users who want to contribute their models to `TGM`. Technically, all they need to do is:
1. Bring their `torch.nn.Module` model implementation to `tgm.nn` without any modification
2. Implement `a proxy model` that inherits from `BaseNN`

Here is an example of a proxy model defined in the example:

```python
# example/linkproppred/tgcn.py

class RecurrentGCN(BaseNN): # Should we move these proxy models somewhere else? or keep them in examples?
    def __init__(self, node_dim: int, embed_dim: int) -> None:
        super().__init__()

        self._requires = {
            'edge_src',
            'edge_dst',
            'num_edge_eventss',
            'foo',
            'neighbour'
        }
        self.recurrent = TGCN(in_channels=node_dim, out_channels=embed_dim)
        self.linear = nn.Linear(embed_dim, embed_dim)

    def forward(
        self,
        batch: DGBatch,
        node_feat: torch.tensor,
        h: torch.Tensor | None = None,
    ) -> Tuple[torch.Tensor, torch.Tensor]:
        edge_index = torch.stack([batch.edge_src, batch.edge_dst], dim=0)
        edge_weight = batch.edge_weight if hasattr(batch, 'edge_weight') else None  # type: ignore

        h_0 = self.recurrent(node_feat, edge_index, edge_weight, h)
        z = F.relu(h_0)
        z = self.linear(z)
        return z, h_0

...

encoder = RecurrentGCN(
    node_dim=train_dg.static_node_x_dim, embed_dim=args.embed_dim
).to(args.device)
decoder = LinkPredictor(args.embed_dim).to(args.device)
opt = torch.optim.Adam(
    set(encoder.parameters()) | set(decoder.parameters()), lr=float(args.lr)
)

hm.validate_nnmodule_requirement(encoder) # This is optional, but recommended for users

...

```

For now, I only apply `BaseNN` to [`example/linkproppred/tgcn.py`](https://github.com/tgm-team/tgm/blob/302-BaseNN/examples/linkproppred/tgcn.py). I will apply `BaseNN` for all other examples and add unit tests if yall are fine with the design.


examples/linkproppred/tncn.py
```diff
@@ -122,10 +122,21 @@ def train(
         inv_neg = batch.global_to_local(batch.neg)
         inv_edge_idx_pos = torch.stack([inv_src, inv_dst], dim=0).long()
         inv_edge_idx_neg = torch.stack([inv_src, inv_neg], dim=0).long()
-        time_info = (last_update, batch.edge_time)
 
-        pos_out = decoder(z, nbr_edge_index, inv_edge_idx_pos, time_info=time_info)
-        neg_out = decoder(z, nbr_edge_index, inv_edge_idx_neg, time_info=time_info)
+        pos_out = decoder(
```
ntgbaoo, line 126: to maintain consistency between TGN and TGAT

tgm/nn/modules/attention.py
```diff
@@ -57,7 +57,7 @@ def __init__(
 
     def forward(
         self,
-        node_feat: torch.Tensor,
+        X: torch.Tensor,
```
ntgbaoo: to maintain consistency in how we define node features

Jacob-Chmura (commented):
This is a valid design, but I haven't gone through the details of the implementation.

I wanted to pitch an alternative path forward. We can discuss tradeoffs in more detail on a call.

**High-Level Objectives**
1. We want to minimize friction for users porting new models from _outside TGM._ Ideally, they shouldn't have to re-write any of the model code, and it should be easy to conform to the TGM protocol / spec
2. We ideally want to provide some hierarchy of common behaviour, so that models developed _within TGM_ can re-use / share relevant structure from existing models. 
3. We want to extend hook-resolution check to the model component, so that models requiring some batch attribute can guarantee it exists in the hook pipeline
4. We want to unify the interface of all TGM-ported models, so that training/inference loops can easily swap model definitions

**Thoughts**
(1) and (2) are kind of at odds. (1) optimizes for easily porting models developed outside of TGM, whereas (2) makes it easier to build up common infrastructure of models within TGM. 3/4 are complementrary to (2), but require slightly more work for the incoming author to implement the contract we define for e.g. encoders, decoders.

As I understand it, the design implemented here favours (2). I wanted to consider something like the below, which much more clearly favours (1). Maybe it can still provide decent outcomes for (3) and (4).

**Idea**
What if we use some sort of meta-programming / inspection to create a decorator that automatically translates a user-defined model into something that conforms to the TGM spec?

I don't want to overcomplicate things, so this is a rather radical proposal. But it would be nice if we could just do something like:
```python
from tgm.nn import tgm_spec

# In my research repo codebase
class MyModel:
    def __init__(...):
        ...
   
    def forward(foo: ..., bar: ..., baz: ...):
        ...

    ...

##############


model = MyModel(...)
tgm_model = tgm_spec(model)
...
```

The idea by that with some magic, `tgm_model` is now a transformed `nn.Module`, which has been restructued s.t.:

- The forward function is of the form `forward(batch: DGbatch, **kwargs)`. We could of course have different transforms for different components (enocder or decoder), but the eng goal is that this transformed class has a TGM based interface
- We inject some runtime check that somehow parses the set of attributes that our model needs from `DGBatch`, and ensures the hooks are setup so that the the data loader and model resolve correctly.

Is something like this even possible? Internally, it needs to do something like what you've written under the hood.

This is asking if we can offload the work of changing the interface/forward of the model away from the author, and automatically generate batch depdendencies, so as to make it easier to run models in TGM (via this proxy object / wrapper).

**Questions**
- If user has not written their model _nor_ their dataloader with TGM, it seems impossible to infer which attributes should be pulled from DGBatch, and which are _extra_. So, it seems we **must** have a base protocol take **batch** on forward. And the user is forced to restructure their code as such :/
- How much do we gain with this complexity in light of the previous point?
- Does this prevent things like jit compilation of the model?

Jacob-Chmura (commented):
Another quick point, more akin to the current design (which I am in favour of). 

As the `NNModule` protocol becomes closer to our `Hook` protocol, it makes me wonder if it would be simpler to just treat models as hooks directly. We already have the hook resolution system, and the means to create hierarchies of common behaviour (e.g. for encoders, decoders, etc).

For people bringing their own models from outside TGM, they should just implement a proxy object (which has the appropriate hook base class). Declare batch dependencies and forward the `forward(batch)` call to a `forward()` call on your own model with the appropriate arguments pulled out

Jacob-Chmura:
> It makes me wonder if it would be simpler to just treat models as hooks directly.

An immediate point of tension is that hooks execute automatically during dataloading. This is valid from a mathematical sense (models call forward on a transformed batch). But in implementation, it means that optimization like prefetch don't really work as intended (since model forward becomes part of the dataloader).

If this has no clear benefit, we could just declare the "model hook" protocol completely disjointly and have them serve as our 2  key components. In which case I'm happy to move forward with the PR design.

examples/graphproppred/gcn.py
```diff
@@ -317,7 +317,7 @@ def eval(
     log_metrics_dict(train_results, epoch=epoch)
     log_metrics_dict(val_results, epoch=epoch)
 
-    val_score = val_results['BinaryAUROC']
+    val_score = val_results['ValidationBinaryAUROC']
```
ntgbaoo: minor bug

examples/graphproppred/tgcn.py
```diff
@@ -298,7 +298,7 @@ def eval(
     log_metrics_dict(train_results, epoch=epoch)
     log_metrics_dict(val_results, epoch=epoch)
 
-    val_score = val_results['BinaryAUROC']
+    val_score = val_results['ValidationBinaryAUROC']
```
ntgbaoo: minor bug

examples/linkproppred/ctan.py
```diff
@@ -234,7 +234,7 @@ def compute_delta_t_stats(train_dg: DGraph) -> Tuple[float, float]:
 )
 train_key, val_key, test_key = hm.keys
 hm.register_shared(nbr_hook)
-hm.register_shared(DeduplicationHook())
+hm.register_shared(DeduplicationHook(seed_nodes_keys=['neg', 'nbr_nids']))
```
ntgbaoo: another minor bug

test/unit/test_nn/test_graphdecoder.py
```diff
@@ -1,8 +1,15 @@
 import pytest
 import torch
 
+from tgm.exceptions import BadAggregatorProtocolError
 from tgm.nn import GraphPredictor
-from tgm.nn.decoder.graphproppred import mean_pooling, sum_pooling
+from tgm.nn.modules import MeanEmbdPooling, SumEmbdPooling
```
shenyangHuang, line 6: I think previous wording of mean_pooling and sum_pooling is more clear.
Alternatively, we can call them:
`mean_aggr`, `sum_aggr`

shenyangHuang: EmbdPooling is kinda implicit as we will only pool over embeddings

tgm/hooks/hook_manager.py
```diff
@@ -207,6 +225,151 @@ def activate(self, key: str) -> Iterator[None]:
...
+                are found.
+
+        Notes:
+            Similarity cutoff is set to 0.6 (difflib default), which balances
+            catching common typos without introducing noisy suggestions:
+            - 0.8+ strict, only near-exact matches.
+            - 0.6  default, catches most common typos.
+            - 0.4- loose, may return irrelevant suggestions.
+        """
+        return difflib.get_close_matches(attribute, produces, n=2, cutoff=0.6)
```
shenyangHuang, line 371: difflib.get_close_matches so this is the fuzzy logic matching.

tgm/nn/modules/__init__.py
```diff
@@ -5,11 +5,25 @@
...
 from .poptrack import PopTrackPredictor
 
+# from .merge import ConcatMerge, LearnableSumMerge
+# from .embd_pooling import MeanEmbdPooling,
+from .aggregation import (
+    Aggregator,
+    ConcatMerge,
+    LearnableSumMerge,
+    MeanEmbdPooling,
+    SumEmbdPooling,
```
shenyangHuang, line 15: naming is a bit inconsistent

`cat_aggr`, `mean_aggr`, `sum_aggr`

I don't think `LearnableSumMerge` is descriptive to me. It is not clear that there can only be two elements and they will each have one nn layer associate with them. How about we just keep `sum_aggr` and have the learnable part be a projection module or sth?

shenyangHuang: `ConcatMerge` changes the output shape while all of the others doesn't change the tensor change (it is all mapping from a set of embedding to a single embedding) while `ConcatMerge` is outputing a different dim. Alternatively, we simply don't support `ConcatMerge` and user can do it themselves in pytorch without the shape problem

tgm/nn/modules/aggregation.py
```diff
@@ -0,0 +1,132 @@
...
+    def __init__(self, dim: int):
+        super().__init__()
+        self.dim = dim
+
+    @property
+    def out_channels(self) -> int:
+        r"""Dimension of the returned embedding after merging."""
+        return self.dim * 2
+
+    def __call__(self, z_src: torch.Tensor, z_dst: torch.Tensor) -> torch.Tensor:
```
shenyangHuang, line 43: it is not clear that why `Concat` should only work for 2 embeddings, this is too link prediction specific. It should work generally with a set of embeddings. In this sense, you can't know the output shape, we can just let the users call `torch.cat` themselves

shenyangHuang (changes requested):
See my comments on naming convention and how ConcatMerge and LearnableSumMerge should be modified

tgm/nn/modules/aggregation.py
ntgbaoo: For link prediction, here is what we did before for link decoder:
```python
def cat_merge(z_src: torch.Tensor, z_dst: torch.Tensor) -> torch.Tensor:
    r"""Default merging operation: Concat."""
   ...


class LearnableSumMerge(nn.Module):
    r"""Sum node-level embeddings after a linear projection."""

    def __init__(self, node_dim: int) -> None:
        ...

    def forward(self, z_src: torch.Tensor, z_dst: torch.Tensor) -> torch.Tensor:
       ....


MERGE_OP: Dict[str, Callable] = {
    'concat': cat_merge,
    'sum': LearnableSumMerge,
}


class LinkPredictor(torch.nn.Module):
    r"""Compute edge embedding given src and dst node embeddings.

    Args:
        node_dim (int): Dimension of node embedding
        out_dim (int): Dimension of output
        nlayers (int): Number of layers
        hidden_dim (int): Size of each hidden embedding
        merge_op (str): Operation to merge 2 node embeddings (concat)
    """

    def __init__(
        self,
        node_dim: int,
        out_dim: int = 1,
        nlayers: int = 2,
        hidden_dim: int = 64,
        merge_op: str = 'concat',
    ) -> None:
        super().__init__()

        if merge_op not in MERGE_OP:
            raise ValueError(
                f'{merge_op} merge operations is not support. Please choose from {list(MERGE_OP.keys())}'
            )

        if merge_op == 'concat':
            self.merge = MERGE_OP[merge_op]
            in_dim = node_dim * 2
        else:
            self.merge = MERGE_OP[merge_op](node_dim)
            in_dim = node_dim

        ...

```
when users want to use different way to merge `src` and `dst` embeddings, they need to modify the core of lib (edit `MERGE_OP` to include their custom merge), which is not ideal.

That is why I refactored this code. The idea is to have a protocol for Aggregation for merging 2 node embeddings. As long as users' custom merging strategies follow the protocol, they can inject their custom merge operations to `LinkPredictor` as follows:
```python
decoder = LinkPredictor(..., merge_op = CustomMerge())
```

> this is too link prediction specific

Yes, you are right, since we want this for the link predictor. `PyG` and `Torch` offer us an interface for aggregation to work with a set of embeddings already and we don't need to rebuild it. I think this is out of our scope, and not sure if it is necessary.

tgm/nn/modules/__init__.py
ntgbaoo: > ConcatMerge changes the output shape while all of the others doesn't change the tensor change (it is all mapping from a set of embedding to a single embedding) while ConcatMerge is outputing a different dim. Alternatively, we simply don't support ConcatMerge and user can do it themselves in pytorch without the shape problem

See my below comment, the idea to solve the problem is not able to inject a custom merge operation into `LinkPredictor` while leaving core TGM lib untouched

ntgbaoo: > naming is a bit inconsistent
cat_aggr, mean_aggr, sum_aggr

I want to use `*Merge` for operations that merge 2 node embeddings for `LinkPredictor` and `*EmbdPooling` for `GraphPredictor`.

test/unit/test_nn/test_graphdecoder.py
ntgbaoo: > I think previous wording of mean_pooling and sum_pooling is more clear.
Alternatively, we can call them:
mean_aggr, sum_aggr

I don't think `mean_aggr` and `sum_aggr` follow our naming convention for a class. [See this comment ](https://github.com/tgm-team/tgm/pull/402#discussion_r3370371017)to get an idea of why I made these into classes 

> EmbdPooling is kinda implicit as we will only pool over embeddings

I think this is exactly what we need for `GraphPredictor` for graph classification

shenyangHuang (approved):

tgm/nn/modules/aggregation.py
shenyangHuang: yes, make sense.

Issue #258, opened 2025-09-20
Vectorize Edgebank


Jacob-Chmura:

benjaminnNgo:
@Jacob-Chmura , @shenyangHuang . What do you guys think about keeping track of seen edges in 2-D tensor? It is a trade-off between space and time complexity.

PR #404, opened 2026-05-06
Vectorized edgebank

ntgbaoo:
### Summary / Description

This PR aims to vectorize `EdgeBank`

**Related Issues:** #258 

#### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence

- [ ] Unit tests
- [ ] Integration tests
- [ ] Performance tests

This PR is a draft for discussion. The approach implemented in this PR is to track an `NxN` tensor, where `N` is the number of nodes. The obvious drawback of this approach is the costly memory. The benefit of this approach is vectorizing inference and update, thus speeding up ~1.7 times


tgm/nn/modules/edgebank.py
```diff
@@ -83,27 +85,10 @@ def __init__(
...
 
-        self.memory: Dict[Tuple[int, int], int] = {}
-        # maintain bidirectional linked list with 2 pointers
-        self._head: Optional[_Event] = None
-        self._tail: Optional[_Event] = None
-
-        logger.warning(
-            'EdgeBank will be slow if events are added/updated out of order.'
-        )
+        self._memory = torch.zeros((N, N), dtype=ts.dtype, device=ts.device)
```
shenyangHuang: do we want to make this dense? how about a sparse matrix?

tgm/nn/modules/edgebank.py
```diff
@@ -46,6 +47,7 @@ def __init__(
             src (torch.Tensor): Source node IDs of edges used for initialization.
             dst (torch.Tensor): Destination node IDs of edges used for initialization.
             ts (torch.Tensor): Timestamps of edges used for initialization.
+            N (int): total number of nodes in the graph.
```
shenyangHuang: technically edgebank can be used directly for streaming setting as well, here setting a N is fine but alternatively, it can allocate larger size when a threshold is hit. (similar to arrays). So maybe we can have the N here be optional.

shenyangHuang (changes requested):
While it makes edgebank faster, it also allocates significant more space (N^2) rather than using what is needed, O(E) which is closer to O(kN). 
I would consider two improvements here:
1. use a sparse matrix, this avoids the O(N^2) cost
2. dynamically expand the size of the memory instead of request the full N at the beginning. This is more useful for later streaming settings and the cost of copying the memory can be amortized. Happy to discuss this next week.

tgm/nn/modules/edgebank.py
ntgbaoo: I thought about it. But from what I am understanding, Sparse Coordinate (COO) tensors are not mutable. Adding/removing an edge requires creating a new tensor, which is even more expensive compared to the existing implementation in `main` with a dictionary.

tgm/nn/modules/edgebank.py
ntgbaoo: `N` is only necessary if we really want to vectorize `EdgeBank`. Otherwise, we can stay with the current implementation from `main`, which supports streaming settings and memory scales linearly w.r.t the number of interactions observed.

shenyangHuang (commented):
Can we have the original dictionary backend as an option here? Alternatively, let's have the vectorized edgebank as its own method? 

`backend=dense` or `backend=dict`

ntgbaoo:
> Can we have the original dictionary backend as an option here? Alternatively, let's have the vectorized edgebank as its own method?
> 
> `backend=dense` or `backend=dict`

@shenyangHuang , added 2 backends to edgebank

tgm/nn/modules/edgebank.py
```diff
@@ -22,56 +23,218 @@ def __init__(
...
+        self.memory: Dict[Tuple[int, int], int] = {}
+        self._head: Optional[_Event] = None
+        self._tail: Optional[_Event] = None
+        self._fixed_memory = fixed_memory
+        self._pos_prob = pos_prob
+        logger.warning(
+            'EdgeBank will be slow if events are added/updated out of order.'
+        )
+
+    def _clean_up(self, window_start: float) -> None:
```
shenyangHuang: this is more similar to an update function, rather than clean up. 
In my intuition, clean up usually means reinitializing sth or something more major. This is more a routine update or --> move forward

shenyangHuang (approved):
Please address the one minor comment and we are good to go.

tgm/nn/modules/edgebank.py
ntgbaoo: Agreed. Updated

Issue #405, opened 2026-05-09
Historical sampling to `NegativeSampler`


ntgbaoo:
**Description**
Add [historical sampling](https://github.com/tgm-team/tgm/blob/main/tgm/hooks/negatives.py#L37) to `NegativeSampler` for the homogeneous setting

**Motivation**
Better negative sampling for training

**Additional context**

PR #406, opened 2026-05-11
Historical negative sampling

ntgbaoo:
### Summary / Description

This PR handles:
- Added historical negative sampling
- Reorganized hooks into sub-modules (if needed)
- Split negative samplers into files
- Added base class for `tgb` negative samplers (`tgbl`, `tkgl`, `thgl`) to avoid duplicated code

**Related Issues:** #405 

#### Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking Change
- [x] Refactoring
- [x] Documentation update

#### Test Evidence

- [X] Unit tests
- [ ] Integration tests
- [ ] Performance tests


tgm/hooks/negatives.py
```diff
@@ -37,30 +49,195 @@ def __init__(
...
+                (2, new_size - self._memory.size(1)),
+                dtype=torch.int32,
+                device=dg.device,
             )
-            self.add_batch_attribute(batch, 'neg_time', batch.edge_time.clone())
-        return batch
+            self._memory = torch.cat([self._memory, edge_buffer], dim=1)
+
+        self._memory[0, self._count : self._count + batch_size] = batch.edge_src
+        self._memory[1, self._count : self._count + batch_size] = batch.edge_dst
```
ntgbaoo: One drawback of this is that the amortized space complexity is O(number of observed edge events) rather than O(number of edges).

tgm/hooks/negatives.py
```diff
@@ -1,17 +1,19 @@
...
-from tgm.hooks import StatelessHook
+from tgm.constants import PADDED_NODE_ID
+from tgm.hooks import StatefulHook, StatelessHook
 from tgm.util.logging import _get_logger
 
 logger = _get_logger(__name__)
 
 
-class NegativeEdgeSamplerHook(StatelessHook):
+class NegativeEdgeSamplerHook(StatefulHook):
```
shenyangHuang: I think we should have two classes, one for random negative, one for historical negative, it is easier conceptually for me to understand as well. Then I can mix and match the two by having two hooks in my hm.

shenyangHuang: and also one of them is stateless, one of them is stateful, we should definitely separate them. Otherwise, by same logic, we should merge uniform sampling with recency sampling

tgm/hooks/negatives.py
```diff
@@ -37,30 +49,204 @@ def __init__(
...
+        return neg, neg_time
+
+    def _random_hist_sampling(
+        self, dg: DGraph, batch: DGBatch
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """Sample negative destination nodes from each source node's historical interactions.
+
+        For each source node in the batch, randomly selects a destination node from
+        its past interactions stored in memory. If a source node has no recorded past
+        interactions, its corresponding negative sample is set to PADDED_NODE_ID as
```
shenyangHuang: negative sample set to PADDED_NODE_ID is fine, but we need to remind users to mask those out correctly. Alternatively the hook can let you know how much is padded?

tgm/hooks/negatives.py
shenyangHuang: isn't this better? you mean we have to keep a buffer in space now?

shenyangHuang (changes requested):
Thanks Bao, two major things:

1. historical negative and random negative should be two hooks, one is stateful, one is stateless
2. I think we should now seriously consider having a folder or a base class that handles negative sampling and shouldn't throw everything in a single script anymore

tgm/hooks/negatives.py
ntgbaoo: This function will return `PADDED_NODE_ID` for nodes that don't have past interactions. However, `PADDED_NODE_ID` will be replaced with random `dst`s before returning. Here is the logic from `__call__`:

```python
elif self.strategy == 'hist_rnd':
  if self._count == 0:
      neg, neg_time = self._random_sampling(dg, batch)
      neg, neg_time = neg[: size[0]], neg_time[: size[0]]

  else: #replace PADDED_NODE_ID with random dst
      rnd_size = round(size[0] * 0.5)
      hst_size = size[0] - rnd_size
      neg_rnd, neg_time_rnd = self._random_sampling(dg, batch)
      neg_hst, neg_time_hst = self._random_hist_sampling(dg, batch)

      original_valid_mask = neg_hst != PADDED_NODE_ID
      valid_idx = torch.where(original_valid_mask)[0]
      cutoff = min(hst_size, valid_idx.size(0))

      neg = neg_rnd.clone()
      neg_time = neg_time_rnd.clone()

      chosen = valid_idx[:cutoff]
      neg[chosen] = neg_hst[chosen]
      neg_time[chosen] = neg_time_hst[chosen]
```

So `PADDED_NODE_ID` won't be propagated to downstream, and for nodes that don't have past interactions, we use random sampling

tgm/hooks/negatives.py
ntgbaoo: Ideally, we want the space complexity of this to scale linearly w.r.t the number of **unique** edges in the graphs. However, the amortized space complexity of this implementation is O(number of **observed** edge events), which means the memory will contain duplicated edges. One benefit of this implementation: it naturally enforces that edges, which appear more frequently in the past, would have higher probabilities to be sampled.

shenyangHuang (approved):
Thanks Bao, looks good on my end

Issue #413, opened 2026-05-28
Recency Sampler silently drops historical neighbors on dense graphs (degree >> buffer_size)


Postroggy:
**Describe the bug**

The `RecencyNeighborSamplerHook` uses a fixed-size circular buffer (`num_nodes × max_neighbors`). On dense graphs where a node's degree significantly exceeds `max_neighbors`, the buffer silently overwrites older interactions. This means that when querying neighbors at an earlier timestamp `t`, the sampler cannot return the correct K most-recent neighbors before `t` — because those edges have already been evicted from the buffer by newer arrivals.

This is a **semantic correctness bug**, not a performance issue. The sampler returns fewer valid neighbors than expected (or entirely wrong neighbors) for any query time that is not close to the latest global timestamp.

**Visualization of the bug**

```mermaid
graph TD
    subgraph "Node v: 1306 total interactions over time"
        A["t=1...t=500<br/>~650 edges"] --> B["t=501...t=1000<br/>~650 edges"]
    end

    subgraph "Circular Buffer (size K=32)"
        C["Holds ONLY the last 32 edges written<br/>(e.g., edges at t=990...t=1000)"]
    end

    B -->|"overwrites"| C
```

```mermaid
sequenceDiagram
    participant Train as Training Loop
    participant Buffer as Ring Buffer (K=32)
    participant Model as Model

    Note over Buffer: Buffer stores last 32 edges globally

    Train->>Buffer: advance(edges at t=1...100)
    Note over Buffer: Buffer now holds edges t=70...100

    Train->>Buffer: advance(edges at t=101...200)
    Note over Buffer: Buffer now holds edges t=170...200<br/>⚠️ edges t=1...169 are LOST forever

    Train->>Buffer: query(node_v, t=150, K=32)
    Buffer->>Model: time_mask filters t>150 → returns EMPTY or near-empty
    Note over Model: ❌ Expected: 32 neighbors before t=150<br/>Got: 0~5 neighbors (rest evicted)
```

```mermaid
graph LR
    subgraph "Sparse Graph (UCI, degree=26 < K=32) ✅"
        S1["Buffer holds ALL history"] --> S2["Query at any t → correct K neighbors"]
    end

    subgraph "Dense Graph (LastFM, degree=1306 >> K=32) ❌"
        D1["Buffer holds last 32 only"] --> D2["Query at t < t_max → most neighbors evicted"]
    end
```

**To Reproduce**

[...]

PR #416, opened 2026-06-03
Added warning to recency sampler

ntgbaoo:
### Summary / Description

Added warning and doc to recency sampler.

**Related Issues:** #413 

#### Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence

- [ ] Unit tests
- [ ] Integration tests
- [ ] Performance tests

### DIFF
--- a/tgm/hooks/neighbors/recency.py
+++ b/tgm/hooks/neighbors/recency.py
@@ -28,13 +28,11 @@ class RecencyNeighborHook(StatefulHook, SeedableHook):
 
     Note:
         - RecencyNeighborSamplerHook assumes queries occur in chronological order.
-        Query timestamp is significantly behind the buffer's current timestamp.
-        Results may be incomplete or incorrect.
+        If query timestamps lag behind the most recently pushed events, the computed neighbours may be incorrect.
         - The order of the output tensors respect the order of seed_nodes_keys.
         For instance, for seed node keys ['edge_src', 'edge_dst', 'neg'] will have the first output index (hop 0) contain the concatenation
         of batch.edge_src, batch.edge_dst, batch.neg (in that order). The next index (hop 1) will contain first-hop neighbors of batch.edge_src
         followed by first-hop neighbors of batch.edge_dst, and then those of batch.neg. This pattern repeats for deeper hops.
-
 
     Raises:
         ValueError: If the num_nbrs list is empty or has non-positive entries.
@@ -65,12 +63,6 @@ class RecencyNeighborHook(StatefulHook, SeedableHook):
             raise ValueError('num_nbrs must be non-empty')
         if not all([isinstance(x, int) and (x > 0) for x in num_nbrs]):
             raise ValueError('Each value in num_nbrs must be a positive integer')
-
-        logger.warning(
-            "RecencyNeighborSamplerHook: query timestamp is significantly behind the buffer's "
-            'current timestamp. Results may be incomplete or incorrect. '
-            'This hook assumes queries occur in chronological order.'
-        )
 
         self._num_nodes = num_nodes
         self._num_nbrs = num_nbrs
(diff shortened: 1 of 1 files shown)


tgm/hooks/neighbors/recency.py
```diff
@@ -27,11 +27,15 @@ class RecencyNeighborHook(StatefulHook, SeedableHook):
         id (str): A unique identifier for the hook. The hook’s name and all attributes it produces will be suffixed with this `id`.
 
     Note:
-        The order of the output tensors respect the order of seed_nodes_keys.
+        - RecencyNeighborSamplerHook assumes queries occur in chronological order.
+        Query timestamp is significantly behind the buffer's current timestamp.
+        Results may be incomplete or incorrect.
```
Jacob-Chmura:
```suggestion
        - RecencyNeighborSamplerHook assumes queries occur in chronological order. If query timestamps lag behind the most recently pushed events, the computed neighbours may be incorrect.
```

tgm/hooks/neighbors/recency.py
```diff
@@ -27,11 +27,15 @@ class RecencyNeighborHook(StatefulHook, SeedableHook):
...
+        - The order of the output tensors respect the order of seed_nodes_keys.
         For instance, for seed node keys ['edge_src', 'edge_dst', 'neg'] will have the first output index (hop 0) contain the concatenation
         of batch.edge_src, batch.edge_dst, batch.neg (in that order). The next index (hop 1) will contain first-hop neighbors of batch.edge_src
         followed by first-hop neighbors of batch.edge_dst, and then those of batch.neg. This pattern repeats for deeper hops.
 
+
```
Jacob-Chmura:
```suggestion
```

tgm/hooks/neighbors/recency.py
```diff
@@ -62,6 +66,12 @@ def __init__(
         if not all([isinstance(x, int) and (x > 0) for x in num_nbrs]):
             raise ValueError('Each value in num_nbrs must be a positive integer')
 
+        logger.warning(
+            "RecencyNeighborSamplerHook: query timestamp is significantly behind the buffer's "
+            'current timestamp. Results may be incomplete or incorrect. '
+            'This hook assumes queries occur in chronological order.'
+        )
+
```
Jacob-Chmura: Probably don't want this log to be triggered on each init?
```suggestion
```

tgm/hooks/neighbors/recency.py
```diff
@@ -233,6 +243,13 @@ def _get_seed_tensors(
     def _get_recency_neighbors(
         self, node_ids: torch.Tensor, query_times: torch.Tensor, k: int
     ) -> Tuple[torch.Tensor, ...]:
+        if query_times.min() < self._nbr_times.min():
+            logger.warning(
+                "RecencyNeighborSamplerHook: query timestamp is significantly behind the buffer's "
+                'current timestamp. Results may be incomplete or incorrect. '
+                'This hook assumes queries occur in chronological order.'
+            )
```
Jacob-Chmura, line 247: It would probably be good to log the time values here. We may also consider adding some extra context to e.g. ensure their dataloader has sorted timestamps, and double check when they reset hook state.

ntgbaoo: What do you think about this @Jacob-Chmura:
```python
logger.warning(
    f"RecencyNeighborSamplerHook: query timestamp - {query_times.min()}, is behind the buffer's "
    f"current timestamp - {self._nbr_times.min()}. Results may be incomplete or incorrect."
    "This hook assumes queries are processed in chronological order."
    "Please ensure that:"
    "(1) dataloader returns batches sorted by timestamp"
    "(2) batches are not being shuffled in a way that breaks temporal ordering"
    "(3) the hook state is reset appropriately between datasets/epochs/evaluation runs"
)
```

Jacob-Chmura: Lgtm, maybe put times in parens () instead of using dash "-" which might be accidentally interpreted as negative

Jacob-Chmura (approved):

tgm/hooks/neighbors/recency.py
ntgbaoo: Good point. Updated.

PR #2, opened 2025-06-11
Rename Library

Jacob-Chmura:
Close #1

PR #418, opened 2026-06-08
Discount Node Label (node_y) in batch size calculation for Dynamic Node Property Prediction Task

### LINKED ISSUE #417: Optional Discount of Node Labels from Batch Size Counting
**Description**
For node property prediction task, for node labels, there should be an optional argument to discount it when iterating over the batches. 

**Additional context**
will help debug TGN v2

shenyangHuang:
### Summary / Description

Added logic to discount node label in batch size calculation for dynamic node property prediction task. 

**Related Issues:** #417 

#### Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [ ] Refactoring
- [ ] Documentation update

#### Test Evidence

- [x] Unit tests
- [x] Integration tests
- [ ] Performance tests

- TGN (20 epochs) 0.37 Val NDCG on tgbn-trade, 0.34 Test NDCG
- DyGFormer (10 epochs) Validation ndcg=0.38
- TGAT (10 epochs) Validation ndcg=0.3746
- TPNet (100 epochs): Validation ndcg=0.3772

**Note**: 
In TGN: 
`log_metric(f'Test {METRIC_TGB_NODEPROPPRED}', test_metric, epoch=args.epochs)`


test/unit/test_data/test_dataloader.py
```diff
@@ -274,3 +274,111 @@ def test_iteration_with_empty_batch_raise():
...
+      pos 0: edge (t=0)
+      pos 1: edge (t=1)
+      pos 2: node_y (t=1)
+      pos 3: edge (t=2)
+      pos 4: node_y (t=3)
+      pos 5: edge (t=4)
+    non_label_pos = [0, 1, 3, 5]
+    """
+    edge_index = torch.IntTensor([[1, 2], [2, 3], [3, 4]])
+    edge_time = torch.LongTensor([0, 1, 2])
```
ntgbaoo: do we need these?

test/unit/test_data/test_dataloader.py
```diff
@@ -274,3 +274,111 @@ def test_iteration_with_empty_batch_raise():
...
+
+    assert len(batches) == 2
+
+    # Batch 0: slice [0, 3) → edges at pos 0,1 + node_y at pos 2 (only 2 edges)
+    b0 = batches[0]
+    assert b0.edge_src.numel() == 2
+
+    # Batch 1: slice [3, 6) → edge at pos 3,5 + node_y at pos 4 (only 2 edges)
+    b1 = batches[1]
+    assert b1.edge_src.numel() == 2
```
ntgbaoo, line 385: should we check for `node_y_nids is not None` and `node_y_nids.numel() == 1` for each batch as well?

tgm/core/_storage/base.py
```diff
@@ -56,6 +56,10 @@ def get_num_timestamps(self, slice: DGSliceTracker) -> int:
     def get_num_events(self, slice: DGSliceTracker) -> int:
         """Return the total number of events in the slice."""
 
+    @abstractmethod
+    def get_node_y_event_positions(self, slice: DGSliceTracker) -> Tensor:
```
ntgbaoo: Should we call this function `get_node_y_event_mask` or `get_node_y_event_index` to keep it consistent with our codebase?

issue.md
ntgbaoo, line 1: should we remove this?

run.sh
ntgbaoo, line 1: should we remove this?

issue.md
shenyangHuang: yes

run.sh
shenyangHuang: yes

shenyangHuang:
Context: tgm's TGN on `tgbn-trade` reported **0.37 val / 0.34 test** NDCG@10 (this PR), vs the TGB leaderboard's **0.395 ± 0.002 val / 0.374 ± 0.001 test**. A side-by-side audit of tgm against the reference implementation (`TGB/examples/nodeproppred/tgbn-trade/tgn.py` + PyG's `LastNeighborLoader`) found the following differences. Each section describes the difference, its impact, and the fix.

---

## 1. GNN `edge_index` was reversed — attention aggregation was dead 🐛

**Difference.** PyG/TGB build the neighbor edge index as `[neighbors, seed_nodes]`: with `TransformerConv`'s default `flow='source_to_target'`, messages flow from row 0 (neighbors) into row 1 (the labeled seed nodes). Our example built `[seed_nodes, neighbors]`, so the messages flowed *into the neighbors* instead. The labeled nodes never aggregated anything from their neighborhood — their embedding degenerated to a linear transform of their own memory, and the attention layer was effectively unused. It also flipped the meaning of `rel_t = last_update[edge_index[0]] - t` in the time encoding.

**Fix.** `examples/nodeproppred/tgn.py`: stack `[neighbors, seeds]` in both train and eval loops. (The linkproppred example has the same pattern and should be checked in a follow-up.)

## 2. `RecencyNeighborHook` never ingested edges from label-free batches 🐛 (core)

**Difference.** The hook's circular-buffer update was inside the `else` branch of the "no seed nodes in this batch" check. Our TGN/TGAT examples seed the hook on `node_y_nids`, which is present in only ~1% of batches — so ~99% of edges never entered the neighbor buffers. TGB inserts **every** edge into its neighbor loader (`process_edges` → `neighbor_loader.insert`). Combined with #1, the sampled neighborhoods were almost always empty.

**Fix.** `tgm/hooks/neighbors/recency.py`: the buffer update now runs for every batch that contains edges, regardless of whether the batch has seed nodes. Covered by a new unit test (`test_hook_updates_buffers_on_batches_without_seeds`). One existing test (`test_sample_with_none_seeds`) constructed the hook with `num_nodes` smaller than the graph's node count and only passed because of this bug; its setup was corrected to `dg.num_nodes`.

## 3. Time semantics: TGB predicts labels *after* ingesting same-timestamp edges ⚠️

**Difference.** In the TGB reference, labels at timestamp *t* are predicted only once every edge with `time < next_label_time` — which includes **all edges at the label's own timestamp *t*** — has been processed into memory and the neighbor loader. On `tgbn-trade` this matters a lot: the year-*t* label vector for a node is exactly that node's year-*t* edge-weight distribution, so the reference model sees the entire answer's supporting edges before predicting. tgm did the opposite on both fronts: memory was updated *after* prediction, and the recency hook filtered neighbors with a strict `time < query_time`.

**Fix (opt-in, off by default).** Two new flags on `RecencyNeighborHook` — `update_buffers_before_sailter` — plus a `--tgb-parity` flag on the example that enables both and moves the memory updatebefore prediction in label batches.                                                                                                                                                                                                
**⚠️ Data-leakage warning.** `update_buffers_before_sampling` (and `inclusive_time_filter`) are deliberately dangerous: they let a prediction at time *t* condition on edges at time *t* from the *same batch*, which is informatioleakage under a strict "predict the future" reading. Rationale for adding them anyway:
                                                                                                                                                                                                                                   - The TGB leaderboard numbers are *defined* by this protocol — every published nodeproppred baseline same-timestamp edges before predicting. Without replicating it, tgm cannot be compared to theleaderboard on equal footing; a residual gap would be a protocol difference, not a modeling difference.                                                                                                                            - One can argue it is not "cheating" under TGB's task definition: node labels at timestamp *t* descrperiod t* (e.g. a country's year-*t* trade distribution), so conditioning on period-*t* edges isanswering "summarize what happened", not "predict the future". But that is a task-semantics choice, and tgm's default should not silently make it.                                                                                 - Therefore both flags **default to False**, tgm's strict `t < label_time` causality remains the defcarries an explicit leakage warning, and the ablation below quantifies exactly how much of thereported performance comes from this protocol choice — which is itself a useful finding about the leaderboard.                                                                                                                     
## 4. Loader silently dropped the first label timestamp of val/test 🐛 (core)                                                                                                                                                      
**Difference.** With `count_node_labels=False`, batch boundaries were computed from non-label event positions only, and the first batch started at the first *edge*. Any node-label events sorted before it — which is exactly the "previous label timestamp" that `TGBSplit` deliberately pulls into the val/test splits (e.g. the 200 split) — were never yielded, so an entire label timestamp was missing from the NDCG average.

**Fix.** `tgm/data/loader.py`: the first batch now starts at event position 0 (both with and without new unit tests.

## 5. Metric averaged per batch instead of per label timestamp

**Difference.** TGB reports `mean over label timestamps` of the per-timestamp NDCG@10. tgm averaged uivalent on tgbn-trade (each label timestamp lands in one batch), but not on tgbn-genre/reddit/tokenwhere a 200-edge batch can span several label timestamps.

**Fix.** The example now computes one NDCG per unique `node_y_time` within a batch and averages over timestamps.

## 6. Non-deterministic ordering of same-timestamp events (core)

**Difference.** `DGData` sorted the global event stream with a non-stable `torch.argsort`, so a label at year *t* could land at an arbitrary position among the year-*t* edges. The correct order (edges before the labels they
explain) held only empirically.

**Fix.** `tgm/data/dg_data.py`: stable sort, guaranteeing edges → node events → node labels at equal unit test.

## 7. Protocol differences

- **Epochs**: TGB trains 50 epochs (no early stopping); tgm defaulted to 30. Now 50.
- **Model selection**: TGB evaluates test *every* epoch and reports the test score at the best-validation epoch. tgm evaluated test only when val improved. The example now follows TGB and logs `Best Validation` / `Best Test` at
the end.
- Remaining known (minor) difference: TGB uses PyG's `TimeEncoder` (`Linear(1,d)` + cos, default init) vs tgm's `Time2Vec` (fixed geometric frequency init). Kept as-is; will revisit only if a gap remains after the ablation.

## 8. Outright bugs in the examples 🐛

- `log_metric(..., epoch=epochs)` referenced an undefined variable → `NameError` the first time test ran. Fixed (`epoch=epoch`) in tgn/tgat/tgcn/gcn/gclstm/tpnet/dygformer.
- `--lr` was declared `type=str` in `tgn.py` and passed unconverted to Adam. Now `type=float`.

---

## Experiment plan

`examples/nodeproppred/run_tgn_ablation.sh` runs TGN on tgbn-trade, 50 epochs, seeds 1–5, in two con

| Config | Meaning | Expectation |
|---|---|---|
| pre-fix (PR #418) | old code | 0.37 val / 0.34 test |
| **strict** | fixes #1–#8, strict `t < label_time` causality | improvement from working attention + eval fixes |
| **parity** | strict + `--tgb-parity` (TGB time semantics) | target: 0.395 ± 0.002 val / 0.374 ± 0.

`parity − strict` isolates how much of the leaderboard number is due to TGB's same-timestamp protoco in #3). Sanity check at 1 epoch (seed 1337, GPU): strict = 0.199 val / 0.176 test, parity = 0.254val / 0.245 test.

---

## Files changed (for PR tracking)

**Core (behavior fixes, with unit tests):**
- `tgm/hooks/neighbors/recency.py` — buffer update runs on every batch with edges (fix #2); new opt-re_sampling` / `inclusive_time_filter`, both default False, docstring carries leakage warning (#3).
- `tgm/data/loader.py` — first batch starts at event 0 when `count_node_labels=False`, so leading label events are no longer dropped (#4).
- `tgm/data/dg_data.py` — stable event sort: edges before labels at equal timestamps (#6).

**Examples:**
- `examples/nodeproppred/tgn.py` — edge_index direction (#1), `--tgb-parity` flag (#3), per-label-timestamp NDCG (#5), epochs 50 + test-every-epoch + best-val model selection (#7), `--lr` as float and `epoch=epoch` (#8).
- `examples/nodeproppred/{tgat,tgcn,gcn,gclstm,tpnet,dygformer}.py` — `epoch=epochs` NameError fix (
- `examples/nodeproppred/run_tgn_ablation.sh` — 5-seed strict/parity sweep.
- `examples/nodeproppred/sbatch_tgn.sh` — SLURM submission wrapper: `sbatch examples/nodeproppred/sbseed>`.

**Tests (5 new, all passing; suite 493 passed):**
- `test/unit/test_hooks/test_recency_nbr_hook.py` — buffer updates on seed-less batches; parity-flag semantics. (`test_sample_with_none_seeds` setup corrected: it under-sized `num_nodes` and only passed because of bug #2.)
- `test/unit/test_data/test_dataloader.py` — leading node_y kept, with and without `drop_last`.
- `test/unit/test_data/test_data.py` — stable sort invariant.

## Status

- [x] All bugs fixed, unit tests green, 1-epoch smoke runs pass on GPU in both modes.
- [ ] 50-epoch seed-1 runs (strict + parity) submitted via sbatch — fill in results here.
- [ ] If parity/seed-1 lands near 0.395 val / 0.374 test: run seeds 2–5 (`run_tgn_ablation.sh`), report mean ± std vs leaderboard.
- [ ] Follow-ups: same edge_index direction check in `examples/linkproppred/tgn.py`; extend parity vt/token and TGAT/DyGFormer/TPNet; time-encoder ablation (Time2Vec vs PyG `Linear+cos`) only if a gapremains.

shenyangHuang:
## TGN on tgbn-trade — 50 epochs, seed 1, NDCG@10

| Config | Best Val | Best Test | Best epoch |
|---|---|---|---|
| pre-fix (PR #418) | 0.37 | 0.34 | — |
| **strict** (all bug fixes, `t < label_time` causality) | **0.3924** | **0.3727** | 30 |
| **parity** (`--tgb-parity`, TGB time semantics) | **0.4012** | **0.3786** | 49 |
| TGB leaderboard | 0.395 ± 0.002 | 0.374 ± 0.001 | — |

**Strict (tgm's default)**: to predict labels at timestamp t, the model can only use edges with time < t. Memory is updated after prediction, and the recency neighbor hook filters sampled neighbors with a strict < comparison. This is a clean "predict the future" protocol — no information from year t leaks into the year-t prediction.

**Parity (--tgb-parity)**: replicates what the TGB reference code actually does. TGB's loop processes all edges with time < next_label_time — which includes every edge at the label timestamp t — into TGNMemory and the neighbor loader before predicting the labels at t. The flag enables this in tgm via three changes: memory update_state runs before prediction in label batches, the recency hook ingests the batch's edges before sampling (update_buffers_before_sampling), and the neighbor time filter becomes <= (inclusive_time_filter).

Why it matters on tgbn-trade specifically: the year-t label vector of a node is literally its year-t edge-weight distribution, so parity lets the model condition on the edges that define the answer. Whether that's "leakage" depends on the task reading — TGB frames labels as summarizing the state at the end of period t, so seeing period-t edges is legitimate under that definition — but tgm keeps strict causality as the default and makes parity opt-in with a warning.

Empirically the gap turned out small: parity − strict ≈ +0.009 val / +0.006 test at seed 1, so most of the leaderboard number comes from working attention and correct evaluation, not from the same-timestamp visibility.

shenyangHuang:
to do:
1. test changes # 5, how much difference does the metric change show in performance?
2. anything with strict can probably merged. 
3. add any failing test of the bugs identified in this PR. 
4. how much efficiency change between stable sort vs. non-stable sort? + memory difference, the stable sort would use double the memory?

test/unit/test_data/test_dataloader.py
ntgbaoo: @shenyangHuang , this has not been fixed yet

examples/nodeproppred/run_tgn_ablation.sh
ntgbaoo, line 1: @shenyangHuang, please remove this

examples/nodeproppred/sbatch_tgn.sh
ntgbaoo, line 1: @shenyangHuang, please remove this

TGB_PARITY_NOTES.md
ntgbaoo, line 1: @shenyangHuang, please remove this

examples/nodeproppred/tgn.py
```diff
@@ -243,15 +281,26 @@ def eval(
...
+
+best_epoch = int(np.argmax(val_curve))
+log_metric(
+    f'Best Validation {METRIC_TGB_NODEPROPPRED}',
+    val_curve[best_epoch],
+    epoch=best_epoch + 1,
+)
+log_metric(
+    f'Best Test {METRIC_TGB_NODEPROPPRED}', test_curve[best_epoch], epoch=best_epoch + 1
+)
```
ntgbaoo, line 306: I thought this 2 versions are the same? Only report test performance on the best validation epoch? If that is the case, I would the prefer the old versions for 2 reasons:
1. Keep it consistent with the remaining examples
2. Efficiency: the new version evaluates on the test set every epoch, but the old version doesn't.

examples/nodeproppred/tgn.py
```diff
@@ -28,8 +28,14 @@
...
 parser.add_argument('--device', type=str, default='cpu', help='torch device')
-parser.add_argument('--epochs', type=int, default=30, help='number of epochs')
+parser.add_argument('--epochs', type=int, default=50, help='number of epochs')
 parser.add_argument('--lr', type=float, default=0.0001, help='learning rate')
+parser.add_argument(
+    '--tgb-parity',
+    action='store_true',
+    help='match TGB reference semantics: ingest edges at the label timestamp '
+    'into memory and neighbor buffers before predicting that label',
+)
```
ntgbaoo, line 38: If this is the proper way to train and evaluate TGN on nodeprop, I would prefer to always do it rather than having a flag to let the user decide what to do, imo.

shenyangHuang: I agree with this, will apply this change consistently

shenyangHuang:
To do:
1. ensure edge update is performed first for all nodeproppred scripts = matching `nodeproppred/tgn.py`
2. test performance of node prop pred for all methods.

examples/nodeproppred/tgn.py
shenyangHuang: agreed, will revert this

TGB_PARITY_NOTES.md
shenyangHuang: removed

examples/nodeproppred/sbatch_tgn.sh
shenyangHuang: removed

examples/nodeproppred/run_tgn_ablation.sh
shenyangHuang: removed

ntgbaoo (approved):
Thanks @shenyangHuang. lgtm!

PR #421, opened 2026-06-27
Negative sampling according to node type for THG

### LINKED ISSUE #370: Enhance on NegativeEdgeSamplerHook
**Description**
- Add [historical sampling](https://github.com/tgm-team/tgm/blob/main/tgm/hooks/negatives.py#L37) for the homogeneous setting 
- [negative sampling with respect to node types](https://github.com/shenyangHuang/TGB/blob/main/tgb/linkproppred/thg_negative_generator.py#L126) for the heterogeneous setting.

ntgbaoo:
### Summary / Description
Adds `NodeTypeNegativeSamplerHook`, a stateful hook for type-consistent negative destination sampling in temporal heterogeneous graphs. For each edge in a batch, it samples a negative destination from previously observed nodes that share the same node type as the positive destination.

**Related Issues:** #

#### Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking Change
- [x] Refactoring
- [x] Documentation update

#### Test Evidence
- [x] Unit tests 
- [ ] Integration tests
- [ ] Performance tests


tgm/hooks/negatives/sampler.py
```diff
@@ -236,3 +238,155 @@ def _update_hst_memory(self, dg: DGraph, batch: DGBatch) -> None:
         self._memory[1, self._count : self._count + batch_size] = batch.edge_dst
 
         self._count += batch_size
+
+
+@hook
+class NodeTypeNegativeSamplerHook(StatefulHook):
```
shenyangHuang: Should we consider putting this in a new file in the negatives folder? Maybe call it `hetero_sampler.py` ? In the future, other sampler can come from here too

tgm/hooks/negatives/sampler.py
```diff
@@ -236,3 +238,155 @@ def _update_hst_memory(self, dg: DGraph, batch: DGBatch) -> None:
...
+        """
+        assert self._memory is not None
+        assert dg.node_type is not None
+
+        dst_node_types = dg.node_type[batch.edge_dst]
+
+        match_node_types = dst_node_types.unsqueeze(1) == self._memory.unsqueeze(0)
+        valid_neg_mask = match_node_types.any(dim=1)
+
+        rand_weights = torch.where(
```
shenyangHuang: randomly samples from any past neighbor of the same type?

tgm/hooks/negatives/sampler.py
shenyangHuang: we could have recency sampler where you sample based on recency but with same node type.

tgm/hooks/negatives/sampler.py
```diff
@@ -236,3 +238,155 @@ def _update_hst_memory(self, dg: DGraph, batch: DGBatch) -> None:
...
+        return batch
+
+    def _node_type_sampling(
+        self, dg: DGraph, batch: DGBatch
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """Sample a random negative destination per edge, constrained to the same node type.
+
+        For each edge, looks up the node type of its destination node, then finds
+        all previously seen destination nodes (stored in memory) that share that type.
+        A random one is selected via vectorized scatter-argmax over uniform random weights,
```
shenyangHuang: we could also easily support exponential sampling by just changing the weight sampling distribution. (though might need the timestamp)

tgm/hooks/negatives/sampler.py
```diff
@@ -236,3 +238,155 @@ def _update_hst_memory(self, dg: DGraph, batch: DGBatch) -> None:
...
+            )
+            self._memory = torch.full(
+                (self._num_nodes,),
+                -1,
+                dtype=torch.int32,
+                device=dg.device,
+            )
+        assert dg.node_type is not None
+
+        self._memory[batch.edge_dst] = dg.node_type[batch.edge_dst]
```
shenyangHuang: If I have exactly 1 node of this type, and all other nodes are of another type, how expensive is it for me to sample this (1 node) type in big O?

shenyangHuang (changes requested):
Thanks for the PR Bao, I would suggest we start a new file for the heterogenous samplers in negatives folder + also see my comment and questions.

tgm/hooks/negatives/sampler.py
ntgbaoo: Good idea! Break negative sampler into:
```
heterogeneous_sampler.py
knowledge_sampler.py
homogeneous_sampler.py
```

tgm/hooks/negatives/sampler.py
ntgbaoo: Yep, I think we can even mimic recency sampling by changing the weight sampling distribution.

tgm/hooks/negatives/sampler.py
ntgbaoo: Yeah. I was debating with myself on this while working on this. Another way is to sample from the node types of all nodes, but is it data leakage: we know the node type of a node that didn't even appear yet?

tgm/hooks/negatives/sampler.py
ntgbaoo: Since we compare every `dst` node in the batch with all nodes in the graph, the theoretical complexity would be `O(batch_size x num_nodes)`. However, since we vectorize everything, the time complexity is reduced to `~O(batch_size x num_nodes/num_cuda_cores)`

tgm/hooks/negatives/sampler.py
shenyangHuang: Yes, we can only sample from node from the past. What I mean by my comment is that we can also sample with recency for example, not necessarily uniformly random, actually given same number of samples, I expect recency to perform better.

tgm/hooks/negatives/sampler.py
shenyangHuang: Yes exactly, thanks

tgm/hooks/negatives/sampler.py
shenyangHuang: Yes, mimic or directly have a true recency one as well.

tgm/hooks/negatives/sampler.py
shenyangHuang: sounds good

shenyangHuang (approved):
Looks good to me, thanks for addressing the changes. As future PR, we can have a true recency sampler for heterogenous and knowledge graph.

